PR #13, opened 2023-12-13
updated param logic on flatten routes utility

stackoverfloweth:
adjusted logic for flatten routes regex, param names must be letters only and now support optional params

### DIFF
--- a/src/utilities/flattenRoutes.ts
+++ b/src/utilities/flattenRoutes.ts
@@ -24,8 +24,8 @@ export function flattenRoutes(routes: Routes, path = ''): RouteFlat[] {
 
 export function generateRouteRegexPattern(path: string): RegExp {
   const optionalParamRegex = /(:\?[\w]+)(?=\W|$)/g
-  const paramRegex = /(:[\w]+)(?=\W|$)/g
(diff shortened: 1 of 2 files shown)


src/utilities/flattenRoutes.ts
```diff
@@ -23,7 +23,9 @@ export function flattenRoutes(routes: Routes, path = ''): RouteFlat[] {
 }
 
 export function generateRouteRegexPattern(path: string): RegExp {
-  const paramRegex = /(:[\w-]+)(?=\W|$)/g
+  const optionalParamRegex = /(:\?[\w]+)(?=\W|$)/g
+  const paramRegex = /(:[\w]+)(?=\W|$)/g
```
pleek91, line 27: nit: `requiredParamRegex`?

pleek91 (approved):
Types are failing because of an unused import. But approving cause otherwise LGTM

src/utilities/flattenRoutes.ts
stackoverfloweth: agreed

pleek91 (approved):

PR #14, opened 2023-12-13
Match alphanumeric params only

pleek91:
# Description
Loosely conforms the types for route params to match params that end with a `/ `, `-`, or `_`. 

This is accomplished by replacing `-` and `_` with `/` on each path prior to parsing out the params. Then `/` is used same as before. Each character we want to support requires a `ReplaceAll<path, '-', '/'>` type to wrap the path. This is recursive template literal type. So I think there probably is some cost to doing this. 

For that reason there are some other special characters that I'm not accounting for here that are valid url characters are `.`, `!`, `~`, `*`, `'`, `(`,  and `)`. Which are not commonly used. I vote we only add this if necessary but I don't feel strongly.

### DIFF
--- a/src/types/routeMethods.ts
+++ b/src/types/routeMethods.ts
@@ -1,6 +1,6 @@
 import { Param, ParamGetSet, ParamGetter } from '@/types/params'
 import { Route, Routes } from '@/types/routes'
-import { Identity, IsAny, IsEmptyObject, TupleCanBeAllUndefined, UnionToIntersection } from '@/types/utilities'
+import { Identity, IsAny, IsEmptyObject, ReplaceAll, TupleCanBeAllUndefined, UnionToIntersection } from '@/types/utilities'
 import { Path } from '@/utilities/path'
 
 export type RouteMethod<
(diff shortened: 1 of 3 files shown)


src/types/routeMethods.ts
```diff
@@ -58,17 +58,23 @@ export type ExtractRouteMethodParams<T> = T extends RouteMethod<infer Params>
...
   : TPath extends string
-    ? ExtractParamsFromPathString<TPath>
+    ? ExtractParamsFromPathString<UnifyParamEnds<TPath>>
     : never
 
+type ParamEnd = '/'
+
+type UnifyParamEnds<
+  TPath extends string
+> = ReplaceAll<ReplaceAll<TPath, '-', ParamEnd>, '_', ParamEnd>
```
stackoverfloweth, line 70: rather than nesting `ReplaceAll`, could we use a union?

```ts
ReplaceAll<TPath, '-' | '_', ParamEnd>
```

pleek91: Every time you have a union the type will produce a union. And we want to replace each character individually so the output type is a single template literal.

stackoverfloweth (approved):

PR #15, opened 2023-12-14
implementation of path

stackoverfloweth:


src/utilities/path.ts
```diff
@@ -12,8 +13,35 @@ export type Path<
...
+
+    if (!(paramName in params)) {
+      value[paramName] = String
+    }
+  }
+
+  return {
+    path,
+    params,
+    toString: () => path,
```
pleek91, line 45: What is the thought of this `toString` method?

src/utilities/path.ts
```diff
@@ -12,8 +13,35 @@ export type Path<
...
   params: P,
+  toString: () => string,
 }
 
-export function path<T extends string, P extends PathParams<T>>(_path: T, _params: Identity<P>): Path<T, P> {
-  throw 'not implemented'
+export function path<T extends string, P extends PathParams<T>>(path: T, params: Identity<P>): Path<T, P> {
+  const optionalParamRegex = /:\?([\w]+)(?=\W|$)/g
+  const requiredParamRegex = /:([\w]+)(?=\W|$)/g
+  const value: Record<string, any> = params
```
pleek91, line 22: I'm guessing this `value` would actually be returned as the `params` in the return of this function? So that the response of this has the complete params object?

src/utilities/flattenRoutes.ts
```diff
@@ -13,19 +14,26 @@ export function flattenRoutes(routes: Routes, path = ''): RouteFlat[] {
...
+        params: getRouteParams(route.path),
       })
     }
 
     return value
   }, [])
 }
 
-export function generateRouteRegexPattern(path: string): RegExp {
+export function getRouteParams(value: string | Path): Record<string, Param[]> {
```
pleek91, line 27: This could probably return `ExtractParamsFromPath<T>`

src/utilities/params.ts
```diff
@@ -1,13 +1,12 @@
...
+import { Param, ParamGetSet, ParamExtras, isParamGetSet, isParamGetter, InvalidRouteParamValueError, ParamReturn } from '@/types'
 
 const extras: ParamExtras = {
   invalid: (message?: string) => {
     throw new InvalidRouteParamValueError(message)
   },
 }
 
-const booleanParam: ParamGetSet = {
+const booleanParam: ParamGetSet<boolean> = {
```
pleek91, line 9: Adding types to these causes the errors in the utilities. I also coulda sworn I got eslint warnings about always true conditions. 

I'd actually lean toward not adding the types here so they are `ParamGetSet<any>` that way typescript and eslint force us to cover all the possibilities. Otherwise we might miss a typescript error that is possible at runtime.

src/utilities/params.test.ts
```diff
@@ -61,13 +61,13 @@ describe('setParamValue', () => {
   test('returns for correct value for Boolean', () => {
     expect(setParamValue(true, Boolean)).toBe('true')
     expect(setParamValue(false, Boolean)).toBe('false')
-    expect(() => setParamValue('', Boolean)).toThrowError()
+    expect(() => setParamValue('' as unknown as boolean, Boolean)).toThrowError()
```
pleek91, line 64: I would maybe use a `@ts-expect-error` rather than `as` here that way if it every doesn't error ts will tell us rather than always just casting to the correct value.

src/types/routeFlat.ts
```diff
@@ -1,5 +1,9 @@
+import { Param } from '@/types/params'
+import { Path } from '@/utilities'
+
 export type RouteFlat = {
   name: string,
-  path: string,
+  path: string | Path,
```
pleek91, line 6: Curious your thoughts of making this a union. At this point I assumed we'd always be dealing with a string since the params are a separate property.

pleek91 (commented):
Left some comments for us to talk through tomorrow sync

src/utilities/flattenRoutes.ts
stackoverfloweth: moved to `types/routeFlat.ts`

src/types/routeFlat.ts
```diff
@@ -1,5 +1,9 @@
...
   regex: RegExp,
+  params: Record<string, Param[]>,
```
stackoverfloweth, line 8: should use `ExtractParamsFromPath<T>`, but with new type for `P` since all params will be populated now (defaults to String)

src/utilities/flattenRoutes.ts
```diff
@@ -1,31 +1,35 @@
 import { RouteFlat, Routes, isParentRoute, isNamedRoute } from '@/types'
+import { Path, combineParams, path as createPath } from '@/utilities'
 
-export function flattenRoutes(routes: Routes, path = ''): RouteFlat[] {
+export function flattenRoutes(routes: Routes, parentPath = '', parentParams = {}): RouteFlat[] {
```
pleek91, line 4: Yeah this signature looks right to me because it matches pretty closely how the types are generated.

src/utilities/flattenRoutes.ts
```diff
@@ -1,31 +1,35 @@
...
   return routes.reduce<RouteFlat[]>((value, route) => {
-    const fullPath = path + route.path
+    const { params }: Path = typeof route.path === 'string' ? createPath(route.path, {}) : route.path
```
pleek91, line 6: fwiw I liked the utility you had for this before. Also, why alias the `path` utility?

src/utilities/paramValidation.ts
```diff
@@ -0,0 +1,10 @@
+import { ExtractParamsFromPath, Route, RouteFlat, Routes } from '@/types'
+
+export function routeParamsAreValid(path: string, route: RouteFlat): boolean {
```
pleek91, line 3: This makes the ts error go away until we implement this. 
```suggestion
export function routeParamsAreValid(_path: string, _route: RouteFlat): boolean {
```

src/utilities/params.ts
```diff
@@ -45,14 +44,14 @@ const numberParam: ParamGetSet = {
...
-      invalid()
+      throw invalid()
     }
 
     return value.toString()
   },
 }
 
-export function getParamValue(value: string, param: Param): unknown {
+export function getParamValue<T extends Param>(value: string, param: T): unknown {
```
pleek91, line 54: Why not keep the overload with the strict type? Did the strict type cause issues where this was being used?

src/utilities/path.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      childId: [optional(String)],
+    }))
+  })
+
+  test('given path not as string, returns each param with corresponding param', () => {
+    const userDefinedParams = {
+      parentId: Boolean,
+    }
+
+    const response = path('/parent/:parentId/child/:childId', userDefinedParams)
```
pleek91, line 41: nit: this is more readable IMO since you're not using `userDefinedParams` in your expect. 
```suggestion
    const response = path('/parent/:parentId/child/:childId', {
      parentId: Boolean,
    })
```

src/utilities/path.spec.ts
pleek91, line 1: We'll want to add a test for a param that appears multiple times.

src/utilities/path.ts
```diff
@@ -1,19 +1,37 @@
...
 }
 
 export type Path<
   T extends string = any,
-  P extends PathParams<T> = any
+  P extends Record<string, Param[]> = any
 > = {
   path: T,
-  params: P,
+  params: Required<P>,
```
pleek91, line 13: I think we could do better here. I'm guessing this is also breaking the route method types which currently depend on Path['Params'] being a non generic value.

src/utilities/path.ts
```diff
@@ -1,19 +1,37 @@
...
+  const matches = Array.from(path.matchAll(paramPattern))
+
+  const paramAssignments = matches.map(([match, paramName]) => {
+    const isOptional = match.startsWith(':?')
+    const param = getParam(params, paramName)
+
+    return {
+      [paramName]: [isOptional ? optional(param) : param],
+    }
+  })
```
pleek91, line 31: Nice, this is so much better.

src/utilities/flattenRoutes.ts
stackoverfloweth: what utility? `path` at some point was reused, but now this is fixed

src/utilities/params.ts
stackoverfloweth: by overload do you mean the generic? I assumed you would wanna drop it since it's not actually used for anything anymore with the return type being `unknown`

src/utilities/flattenRoutes.ts
pleek91: You had a function called getParams or something that had this ternary and just returned the params.

src/utilities/params.ts
pleek91: My thought was the overload is the correct type and the implementation is unknown just because typescript won't narrow the generic inside the implementation. But when you call this function with a value and a param it correctly tells you what type it will return from the overload.

pleek91: In my commit this function was overloaded with a single overload that had the return type to extract the param return value. That becomes the exported signature. And the implementation uses unknown

stackoverfloweth: nice! I somehow missed that

src/types/optional.ts
```diff
@@ -0,0 +1,21 @@
+import { ExtractParamType, Param, ParamGetSet } from '@/types'
+import { getParamValue, setParamValue } from '@/utilities'
+
+export function optional<TParam extends Param, TValue extends ExtractParamType<TParam>>(param: TParam): ParamGetSet<TValue | undefined> {
```
pleek91, line 4:
```suggestion
export function optional<TParam extends Param>(param: TParam): ParamGetSet<ExtractParamType<TParam> | undefined> {
```

src/types/path.ts
```diff
@@ -0,0 +1,11 @@
+import { Param } from '@/types/params'
+
+export type PathParams = Record<string, Param[]>
+
+export type Path<
+  T extends string = any,
+  P extends PathParams = any
+> = {
+  path: T,
+  params: Required<P>,
```
pleek91, line 10: This type of the return will need to reflect that it actually returns the MergedParams because we'll need that type to fix the route method types.

PR #16, opened 2023-12-17
Path types

pleek91:
# Description
Path originally just returned the same types as its arguments. Which was never really correct but allowed for all the other types to be written. Implementing the type path should actually return by having all route method types use Path and having Path be responsible for extracting param types.

### DIFF
--- a/src/types/routeMethods.ts
+++ b/src/types/routeMethods.ts
@@ -58,20 +58,20 @@ export type ExtractRouteMethodParams<T> = T extends RouteMethod<infer Params>
 export type ExtractParamsFromPath<
   TPath extends Route['path']
 > = TPath extends Path
-  ? ExtractParamsFromPathString<UnifyParamEnds<TPath['path']>, TPath['params']>
+  ? TPath['params']
   : TPath extends string
-    ? ExtractParamsFromPathString<UnifyParamEnds<TPath>>
+    ? Path<TPath, {}>['params']
     : never
 
 type ParamEnd = '/'
 
-type UnifyParamEnds<
+export type UnifyParamEnds<
   TPath extends string
 > = ReplaceAll<ReplaceAll<TPath, '-', ParamEnd>, '_', ParamEnd>
 
 export type ExtractParamsFromPathString<
   TPath extends string,
-  TParams extends Record<string, Param> = Record<never, never>
+  TParams extends Record<string, Param | undefined> = Record<never, never>
 > = TPath extends `${infer Path}${ParamEnd}`
   ? ExtractParamsFromPathString<Path, TParams>
   : TPath extends `${string}:${infer Param}${ParamEnd}${infer Rest}`
@@ -115,7 +115,7 @@ type ExtractParamName<
 
 type ExtractPathParamType<
   TParam extends string,
-  TParams extends Record<string, Param>
+  TParams extends Record<string, Param | undefined>
 > = TParam extends `?${infer OptionalParam}`
   ? OptionalParam extends keyof TParams
     ? ExtractParamType<TParams[OptionalParam]> | undefined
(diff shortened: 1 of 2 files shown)


src/types/routeMethods.ts
```diff
@@ -58,20 +58,20 @@ export type ExtractRouteMethodParams<T> = T extends RouteMethod<infer Params>
 export type ExtractParamsFromPath<
   TPath extends Route['path']
 > = TPath extends Path
-  ? ExtractParamsFromPathString<UnifyParamEnds<TPath['path']>, TPath['params']>
+  ? TPath['params']
   : TPath extends string
-    ? ExtractParamsFromPathString<UnifyParamEnds<TPath>>
+    ? Path<TPath, {}>['params']
     : never
```
pleek91, line 64: Here is where you can see rather than extracting param types it just accepts what Path returns for the params type.

src/types/routeMethods.ts
```diff
@@ -58,20 +58,20 @@ export type ExtractRouteMethodParams<T> = T extends RouteMethod<infer Params>
...
 
 type ParamEnd = '/'
 
-type UnifyParamEnds<
+export type UnifyParamEnds<
```
pleek91, line 68: Exporting because this is used by Path. Wanna organize these a little better the the diff is cleaner if I don't move things around.

src/types/routeMethods.ts
```diff
@@ -58,20 +58,20 @@ export type ExtractRouteMethodParams<T> = T extends RouteMethod<infer Params>
...
   TPath extends string
 > = ReplaceAll<ReplaceAll<TPath, '-', ParamEnd>, '_', ParamEnd>
 
 export type ExtractParamsFromPathString<
   TPath extends string,
-  TParams extends Record<string, Param> = Record<never, never>
+  TParams extends Record<string, Param | undefined> = Record<never, never>
```
pleek91, line 74: These `| undefined` additions just account for the user defined param types being an optional record which adds `| undefined` to every property.

src/utilities/path.ts
```diff
@@ -11,7 +11,7 @@ export type Path<
   P extends PathParams<T> = any
 > = {
   path: T,
-  params: P,
+  params: Identity<ExtractParamsFromPathString<UnifyParamEnds<T>, P>>,
```
pleek91, line 14: Here's the implementation of path params. Essentially just moving the `ExtractParamsFromPathString` type here rather than using it in the route method types.

src/utilities/path.ts
```diff
@@ -11,7 +11,7 @@ export type Path<
   P extends PathParams<T> = any
 > = {
   path: T,
-  params: P,
+  params: Identity<ExtractParamsFromPathString<T, P>>,
```
pleek91, line 14: Here's the implementation of path params. Essentially just moving the ExtractParamsFromPathString type here rather than using it in the route method types.

src/types/routeMethods.ts
```diff
@@ -71,12 +71,12 @@ type UnifyParamEnds<
...
-  TParams extends Record<string, Param> = Record<never, never>
-> = TPath extends `${infer Path}${ParamEnd}`
+  TParams extends Record<string, Param | undefined> = Record<never, never>
+> = UnifyParamEnds<TPath> extends `${infer Path}${ParamEnd}`
   ? ExtractParamsFromPathString<Path, TParams>
-  : TPath extends `${string}:${infer Param}${ParamEnd}${infer Rest}`
+  : UnifyParamEnds<TPath> extends `${string}:${infer Param}${ParamEnd}${infer Rest}`
     ? MergeParams<{ [P in ExtractParamName<Param>]: ExtractPathParamType<Param, TParams> }, ExtractParamsFromPathString<Rest, TParams>>
-    : TPath extends `${string}:${infer Param}`
+    : UnifyParamEnds<TPath> extends `${string}:${infer Param}`
```
pleek91, line 79: I moved wrapping the path in `UnifyParamEnds` into this type so that we don't have to remember wrap any path string we pas into this type in it.

stackoverfloweth (approved):

PR #18, opened 2023-12-18
implemented path function

stackoverfloweth:


src/types/optional.ts
pleek91, line 1: Thoughts on just putting this directly in `utilities/path` where its used? I don't see us using this anywhere else and we don't need to export it as a public utility of the library. Otherwise `utilities/optional` would be more accurate yeah?

src/utilities/mergeParams.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      {
+        foo: [String],
+      },
+    ]
+
+    const response = mergeParams(...input)
+
+    expect(response).toHaveProperty('foo')
+    expect(response.foo).toMatchObject([String])
+  })
```
pleek91, line 38: This seems incorrect to me. Shouldn't this type be `[String, String]`? The tuple is a direct representation of how many params by that name there are in the order the appear.

src/utilities/mergeParams.ts
```diff
@@ -0,0 +1,19 @@
+import { Param } from '@/types'
+
+export function mergeParams(...params: Record<string, Param[]>[]): Record<string, Param[]> {
+  return params.reduce<Record<string, Param[]>>((combined, collection) => {
+    Object.entries(collection).forEach(([key, params]) => {
+      if (key in combined) {
+        params.forEach(param => {
+          if (!combined[key].includes(param)) {
```
pleek91, line 8: Referring to the test I commented on I think this condition needs to be removed.

src/utilities/path.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(response.params).toMatchObject({
+      parentId: [String],
+      childId: [String],
+    })
+  })
+
+  test('given path with optional params, returns each param name as type String with optional', () => {
+    const response = path('/parent/:?parentId/child/:?childId', {})
+
+    expect(JSON.stringify(response.params)).toMatch(JSON.stringify({
```
pleek91, line 24: curious why you had to stringify here but not in the other tests? Just because of the optional function?

src/utilities/path.ts
```diff
@@ -14,6 +15,25 @@ export type Path<
...
+    const param = getParam(params, paramName)
+
+    return {
+      [paramName]: [isOptional ? optional(param) : param],
+    }
+  })
+
+  return {
+    path,
+    params: mergeParams(...paramAssignments) as Identity<ExtractParamsFromPathString<T, P>>,
```
pleek91, line 37: Would this work as well? Just cause its shorter and will automatically update if we change the Path type. Don't feel strongly about this just had the thought when I saw it. 
```suggestion
    params: mergeParams(...paramAssignments) as Path<T, P>["params"],
```

src/utilities/mergeParams.spec.ts
stackoverfloweth: agreed. I think this was originally written to solve a different problem and never got revisited. Good catch

src/utilities/path.spec.ts
stackoverfloweth: if I don't stringify, the `optional(String)` just ends up as `[Function Boolean]`. This is the only way I could find to verify that the param will be "Optional" as we've defined it with the business logic found in optional function

src/utilities/mergeParams.ts
```diff
@@ -0,0 +1,15 @@
+import { Param } from '@/types'
+
+export function mergeParams(...params: Record<string, Param[]>[]): Record<string, Param[]> {
+  return params.reduce<Record<string, Param[]>>((combined, collection) => {
+    Object.entries(collection).forEach(([key, params]) => {
+      if (key in combined) {
+        params.forEach(param => combined[key].push(param))
+      } else {
+        combined[key] = params
```
pleek91, line 9: We might want to spread into a new array here. Otherwise we'll be update this first tuple in place with any additional values that get merged into it.

stackoverfloweth: so so? 

```ts
combined[key] = [...params]
```

pleek91 (approved):

PR #20, opened 2023-12-18
Make sure to always return a tuple for params in ExtractParamsFromPathString

pleek91:
# Description
Fixes single param not being a tuple


stackoverfloweth (approved):
glad this was an easy one

PR #21, opened 2023-12-18
Validate params

stackoverfloweth:


src/utilities/flattenRoutes.ts
```diff
@@ -1,31 +1,39 @@
...
+export function toPath(value: string | Path): Path {
+  return typeof value === 'string' ? path(value, {}) : value
+}
+
+export function generateRouteRegexPattern(value: string): RegExp {
   const optionalParamRegex = /(:\?[\w]+)(?=\W|$)/g
   const requiredParamRegex = /(:[\w]+)(?=\W|$)/g
 
-  const routeRegex = path.replace(optionalParamRegex, '([^/]*)').replace(requiredParamRegex, '([^/]+)')
+  const routeRegex = value.replace(optionalParamRegex, '(.*)').replace(requiredParamRegex, '(.+)')
```
pleek91, line 37: In the path utility there's a single regex that matches both optional and required params. Can we reuse that here? Quite possible they are incompatible.

src/utilities/paramValidation.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  test('given route with OPTIONAL string param WITHOUT value present, returns true', () => {
+    const route: Route = {
+      name: 'simple-params',
+      path: '/simple/:?simple',
+      component,
+    }
+    const [flat] = flattenRoutes([route])
+
+    const response = routeParamsAreValid('/simple', flat)
```
pleek91, line 55: I think this test is incorrect. It shouldn't match `/simple` because it's missing the `/` at the end right?

src/utilities/paramValidation.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      component,
+    }
+    const [flat] = flattenRoutes([route])
+
+    const response = routeParamsAreValid('/simple/123', flat)
+
+    expect(response).toBe(true)
+  })
+
+  test('given route with non-string param with value that does NOT satisfy, returns false', () => {
```
pleek91, line 75: Love seeing this working!

src/utilities/paramsFinder.ts
```diff
@@ -0,0 +1,9 @@
+export function findParamValues(value: string, path: string, paramName: string): string[] {
```
pleek91, line 1: Is value the url? If so maybe that's a better name for this argument. Took me a minute to figure this out from the tests.

src/utilities/paramsFinder.spec.ts
pleek91, line 1: Not seeing any tests that handle optional params.

src/utilities/paramValidation.ts
```diff
@@ -0,0 +1,24 @@
...
+import { getParamValue, findParamValues } from '@/utilities'
+
+export function routeParamsAreValid(value: string, route: RouteFlat): boolean {
+  const entries = Object.entries<Param[]>(route.params)
+
+  if (entries.length === 0) {
+    return true
+  }
+
+  for (const [key, paramsTuple] of entries) {
```
pleek91, line 11: Couple nits. Variable name and also don't need the condition to dump early because an array of 0 will just return true anyway. 

```suggestion
  const params = Object.entries<Param[]>(route.params)

  for (const [key, paramsTuple] of params) {
```

pleek91 (commented):
This is great. Some questions but otherwise this looks good to go

pleek91:
Although I just realized the tests are failing lol. Lemme know if this needs another review once tests are passing.

src/utilities/flattenRoutes.ts
stackoverfloweth: that's a good catch, but there is a subtle difference in the replacement values.

Notice how the optional gets replaced with `(.*)` (any char 0 or more times)
and the required gets replaced with `(.+)` (any char 1 or more times)

src/utilities/paramValidation.spec.ts
stackoverfloweth: 100%. This test was failing too and I was just about to dig into why. Thanks @pleek91 👏

src/utilities/flattenRoutes.ts
pleek91: ah right 👍

src/utilities/paramsFinder.ts
stackoverfloweth: yeah, I considered this as well. I was just nervous to introduce a new term into our lexicon since `url` is not currently used anywhere

pleek91: We'll end up using url at some point. And I think we'll be passing it in directly to this utility. So I think that makes sense.

src/utilities/paramValidation.ts
stackoverfloweth: yep, love this change

src/types/params.ts
```diff
@@ -2,15 +2,15 @@ export type ParamExtras = {
   invalid: (message?: string) => never,
 }
 
-export type ParamGetter<T = any> = (value: string, extras: ParamExtras) => T
-export type ParamSetter<T = any> = (value: T, extras: ParamExtras) => string
+export type ParamGetter<T = any> = (value: string | undefined, extras: ParamExtras) => T
+export type ParamSetter<T = any> = (value: T, extras: ParamExtras) => string | undefined
```
pleek91, line 6: Seeing this now I'm second guessing adding the `undefined` to ParamGetter. A param getter should never be called with an undefined value right?

I also think the same for ParamSetter. It should always return a string. 

Optional values shouldn't ever pass through userland get/set methods. 

These are purely to support the `optional` utility yeah? Wonder if there's a clean way to support that without adding `undefined` to the exported types.

src/utilities/paramValidation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  for (const [key, paramsTuple] of params) {
+    const stringValues = findParamValues(url, route.path, key)
+
+    try {
+      paramsTuple.forEach((param, index) => {
+        const stringValue = stringValues[index]
+
+        return getParamValue(stringValue, param)
+      })
```
pleek91, line 15: nit: doesn't seem like the const assignment is doing anything for us here. 
```suggestion
      paramsTuple.forEach((param, index) => getParamValue(stringValues[index], param))
```

src/utilities/paramValidation.ts
```diff
@@ -0,0 +1,22 @@
+import { Param, RouteFlat } from '@/types'
+import { getParamValue, findParamValues } from '@/utilities'
+
+export function routeParamsAreValid(url: string, route: RouteFlat): boolean {
+  const params = Object.entries<Param[]>(route.params)
+
+  for (const [key, paramsTuple] of params) {
+    const stringValues = findParamValues(url, route.path, key)
```
pleek91, line 8: Are these always strings or could undefined exist in this array?

src/utilities/params.ts
```diff
@@ -81,7 +106,7 @@ export function getParamValue<T extends Param>(value: string, param: T): unknown
   return value
 }
 
-export function setParamValue(value: unknown, param: Param): string {
+export function setParamValue(value: unknown, param: Param): string | undefined {
```
pleek91, line 109: Seeing the `undefined` show up here makes more sense to me. Is the thought that when we're setting all `undefined` values will be coerced to empty strings?

src/utilities/path.ts
```diff
@@ -41,15 +42,15 @@ export function path<T extends string, P extends PathParams<T>>(path: T, params:
...
         return undefined
       }
 
       return getParamValue(value, param)
     },
     set: (value) => {
-      if (value === undefined) {
-        return ''
+      if (!stringHasValue(value)) {
+        return undefined
```
pleek91, line 53: Should we just coerce to an empty string here?

src/utilities/paramValidation.ts
stackoverfloweth: you're right that in the previous configuration these values were `(string | undefined)[]`, but after making these code review suggestions it's actually all strings, `undefined` values are coming back as an empty string, which is a requirement after changing `getParamValue` back to only accepting a string value

pleek91: That makes sense to me 👍

src/types/params.ts
stackoverfloweth: no worries. I think it will work out just as well to use empty strings as being synonymous with `undefined`

src/utilities/params.ts
stackoverfloweth: I reverted this change along side `getParamValue`, I think they should mirror each other

src/utilities/params.ts
```diff
@@ -30,6 +47,10 @@ const booleanParam: ParamGetSet<unknown> = {
 
 const numberParam: ParamGetSet<unknown> = {
   get: (value, { invalid }) => {
+    if (value === '') {
+      throw invalid()
+    }
```
pleek91, line 52: Didn't expect to see this here. An empty string should only match an optional param right?

src/utilities/paramsFinder.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  test('given paramName that matches multiple param on route and value is present, returns each value in array', () => {
+    const response = findParamValues('/simple/ABC/DEF/gap/GHI', '/simple/:simple/:simple/gap/:simple', 'simple')
+
+    expect(response).toMatchObject(['ABC', 'DEF', 'GHI'])
+  })
+
+  test('given paramName that matches multiple param on route value is present with some optional, returns each value in array including empty string', () => {
+    const response = findParamValues('/simple/ABC//gap/GHI', '/simple/:simple/:?simple/gap/:simple', 'simple')
+
+    expect(response).toMatchObject(['ABC', '', 'GHI'])
```
pleek91, line 38: I still expected to get undefined for a missing optional parameter. Since that's what the optional utility returns right? The typescript types are gonna say this is string | undefined

pleek91 (commented):
Overall looks really good. Like the direction this is going. Two questions but otherwise this is ready to merge I think.

src/utilities/params.ts
stackoverfloweth: the process for validating params is to call the getter. So if the param finder gets an empty string for a value, and the route says a param is NOT optional and uses `Number`, it's going to call this function with an empty string, get an exception thrown, and know that param validation failed

src/utilities/paramsFinder.spec.ts
stackoverfloweth: paramFinder runs before paramValidation. This service just plucks param values (as strings) out of a given URL. The next step would be to take those values and run them through the param getter (which could be using the optional utility). Only at that point would the value actually be transformed into `undefined`

pleek91: ah gotcha, that makes sense

src/utilities/params.ts
pleek91: hmm that makes sense. But I also think the user shouldn't be responsible for checking if the value is an empty string. If a param is required the regex to match it should be `+` (at least one) right? Vs an optional param the regex would be `*` (0 or more). That way if a required param is missing from the url it doesn't even match. By the time we get to user defined `Param`s I would expect there to be a non empty string to pass into the get.

pleek91: If the regex does need to match `*` for any param maybe we should create a `required` wrapper utility that is the opposite of the `optional` utility` and wrap all `Param`s in either `optional` or `required`. That way `Param` only ever deals with actual values.

stackoverfloweth: actually you're correct that the regex test does use `+`, so technically an empty string should never makes it's way to a required param. I added this to test logic from paramValidation, but the control actually happens before paramValidation so I'll update this and move the test to before paramValidation

src/utilities/params.ts
```diff
@@ -7,6 +7,19 @@ const extras: ParamExtras = {
...
+const stringParam: ParamGetSet<unknown> = {
+  get: (value) => {
+    return value
+  },
+  set: (value, { invalid }) => {
+    if (typeof value !== 'string') {
+      throw invalid()
+    }
+
+    return value.toString()
```
pleek91, line 19: we probably don't need the toString() since we've already checked if its not a string right?

src/utilities/params.spec.ts
```diff
@@ -13,7 +13,6 @@ describe('getParamValue', () => {
   test('returns for correct value for Number', () => {
     expect(getParamValue('1', Number)).toBe(1)
     expect(getParamValue('1.5', Number)).toBe(1.5)
-    expect(() => getParamValue('', Number)).toThrowError()
```
pleek91, line 16: You removed this because an empty string will never be passed to the getter yeah? Could we add a failing example like `foo` (anything that returns `NaN`.

pleek91 (approved):
Last two comments are non blocking. This looks great!

src/utilities/paramsFinder.spec.ts
pleek91: @stackoverfloweth do we need a optional param test here?

stackoverfloweth: I did add a test for optional params (line 35) "given paramName that matches multiple param on route value is present with some optional, returns each value in array including empty string"

pleek91: Oh that's my bad. I saw this comment was unresolved and clicked on the filename from the comment and it took me to the commit I commented on instead of the latest. TY

src/utilities/params.spec.ts
stackoverfloweth: correct. I think through the course of this PR we agreed that empty strings will never match a param unless it's optional, and in that case it would be replaced with `undefined`. But yeah, it's probably worth having a case where these tests throw an error 👍

stackoverfloweth: I actually updated all the tests that used an empty string to use `foo` instead, you cool with that?

pleek91: Yeah that makes sense to me 👍

pleek91 (approved):

PR #22, opened 2023-12-19
Middleware types

pleek91:
# Description
Some basic types for what functionality middleware might have. Pretty simple but I don't think it needs much. See comments for more notes.


src/types/middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+import { MaybePromise } from '@/types/utilities'
+
+type Route = {
+  name: string,
+  params: Record<string, unknown>,
+  query: unknown,
+  hash: string,
+}
+
+interface RouteState {
```
pleek91, line 11: This is intentionally empty. The idea is that users can declare their own RouteState at implementation via declaration merging. This will likely need a namespace or module to work properly.

src/types/middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+
+}
+
+interface RouteRejection {
+  NotFound: unknown,
+}
```
pleek91, line 17: This is an interface so users can add their own rejections types at implementation. Each rejection type will be able to render a component specified when creating the router.

src/types/middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+type RouteReject = (type?: keyof RouteRejection) => void
+
+type MiddleWareExtras = {
+  from: Route | null,
+  state: RouteState,
+  reject: RouteReject,
+  scope: EffectScope,
+}
+
+export type RouteMiddleware = (route: Route, extras: MiddleWareExtras) => MaybePromise<void>
```
pleek91, line 28: This is pretty simple stuff. The `route` is the route you're going to. `scope` is maybe not necessary but seemed like a nice to have. That will be the route's effect scope that is disposed when the route is exited.

src/types/middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+
+}
+
+interface RouteRejection {
+  NotFound: unknown,
+}
+
+type RouteReject = (type?: keyof RouteRejection) => void
+
+type MiddleWareExtras = {
```
pleek91, line 21: Should we provide the router here even though it won't have types?

pleek91: From my testing typescript at least seems happy if I just import a router into my middleware file and use it directly

pleek91: I believe when a project that implements this is built, both the routes, the router, and the middleware definitions end up in the same bundle. And since router is referenced/invoked in a function. There is no circular dependency/import. So as long as typescript is happy with this I think we're safe to say that importing your router and using it in middleware is recommended. But I could be easily convinced that providing the router as an "extra" would be beneficial. Maybe wait until someone asks for it

src/types/middleware.ts
```diff
@@ -0,0 +1,28 @@
+import { EffectScope } from 'vue'
+import { MaybePromise } from '@/types/utilities'
+
+type Route = {
+  name: string,
+  params: Record<string, unknown>,
+  query: unknown,
+  hash: string,
+}
```
pleek91, line 9: We need to define this better. We'll want some type for the route that has been "resolved". I think the RouteFlat type might be what is used here but I'm not sure.

stackoverfloweth: I designed some components at my previous company that had types that built off a base and was really nice to work with, maybe we can do something similar here?

in this case I think the base is something like

```ts
type Route = {
  name: string,
  ...
}
```

and then we build on that with something like

```ts
type Resolved<T extends Route> = T & {
  params: Record<string, Param[]>
  ...
}
```

so components/services that expect a resolved route would be written as

```ts
function middleware(route: Resolved<Route>): void
```

I could see this being too "inheritance" for your liking, just throwing it out there as a possible syntax

src/types/middleware.ts
stackoverfloweth: does this need to be exported to work the way we want it to?

src/types/middleware.ts
stackoverfloweth, line 21: nit, I like middleware as 1 word

```suggestion
type MiddlewareExtras = {
```

stackoverfloweth: I like the idea of including in extras. Even if we suggest and users get a better experience from importing, it's nice to provide an alternative syntax for users who are more comfortable not importing. That also would allow developers to write middleware that's agnostic of where or how to import a local router

src/types/middleware.ts
pleek91: Yeah probably exported as part of a ts module or namespace. Don't 100% have my head wrapped around that. Just waited to nail down the internal types.

src/types/middleware.ts
pleek91: No I like that, I wanna see what that looks like.

src/types/middleware.ts
pleek91: Agreed. Will update.

src/types/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type RouteReject = (type?: keyof RouteRejection) => void
+
+type MiddlewareExtras = {
+  from: Route | null,
+  state: RouteState,
+  reject: RouteReject,
+  scope: EffectScope,
+  router: Omit<Router<[]>, 'routes'> & { routes: Record<string, any> },
```
pleek91, line 27: We'll likely want a better type for this at some point. But for now this works. Will revisit once we spec out what `Router` is more.

src/types/middleware.ts
pleek91: I think we just need to export it and then users can add their custom types like this
```typescript
declare module '@kitbag/router {
  export interface RouteState {
    user: User,
    account: Account,
  }

  export interface RouteRejection {
    Auth: unknown
  }
}
```

So I've updated these to be exported.

stackoverfloweth (approved):

PR #23, opened 2023-12-21
Router types

pleek91:
# Description
Fills out the `Router` type. Decided to leave the `routeMatch` method as part of the router for now to avoid making this change bigger.

### DIFF
--- a/src/utilities/createRouter.ts
+++ b/src/utilities/createRouter.ts
@@ -12,7 +12,7 @@ export type Router<T extends Routes> = {
 }
 
 function createRouteMethods<T extends Routes>(_routes: T): RouteMethods<T> {
-  throw 'not implemented'
+  return {} as any
 }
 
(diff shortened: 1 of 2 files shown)


src/utilities/createRouter.ts
```diff
@@ -2,18 +2,55 @@ import { RouteFlat, RouteMethods, Routes } from '@/types'
...
+    push,
+    replace,
+    forward,
+    back,
+    go,
     routeMatch,
-  } as Router<T>
+  }
+
+  return router
```
pleek91, line 55: Worth pointing out. I ran into that "Type instantiation is excessively deep and possibly infinite" error here again if I did either
```javascript
const router: Router<T> = { ... }
```
or
```javascript
return { ... }
```
But just letting it infer the type for the router const and then returning it is fine for some reason.

stackoverfloweth (approved):

PR #24, opened 2023-12-21
Add types for what a RouteMethod returns

pleek91:
# Description
Simple type for what RouteMethod will return


stackoverfloweth (approved):
so exciting to see actual end-user router things!

PR #25, opened 2023-12-21
Add RouteMeta and simplify Route types

pleek91:
# Description
Set out to add `meta` to the `Route` type. Figured it was time to create a `BaseRoute` type that had all the common pieces. When I did that I started getting circular reference errors which I realized was because of the generics for the path that don't actually need to be generic. Removing the generic simplified the types and make them more accurate by removing the default `any`.


src/utilities/flattenRoutes.spec.ts
pleek91, line 1: Since `Route` no longer has a generic that defaults to `any` this test had a type error because `.path` could potentially be a `Path` based on the types. Using satisfies lets typescript know its actually just a string.

src/types/routeMethods.ts
pleek91, line 1: I'm not 100% sure why the `extends string | Path` is needed since `TRoute extends Route` and `Route` must have a `path` of `string | Path`. I think this was again something that was being a little generous because `Route` defaulted the path to `any`.

stackoverfloweth (approved):

PR #26, opened 2023-12-21
RouteFlat -> Resolved<Route>

stackoverfloweth:
- rename flat to resolved
 - using `Resolved<T>` approach to nesting types
 - include the rest of route in what was previously RouteFlat


src/types/resolved.ts
```diff
@@ -1,6 +1,7 @@
 import { Param } from '@/types/params'
+import { Route } from '@/types/routes'
 
-export type RouteFlat = {
+export type Resolved<T extends Route> = T & {
```
pleek91, line 4: We could probably use `T` to tell us what `path` and `params` actually are. I don't think there's much of a benefit to `path` but maybe to `params`? No change requested. Just an idea.

pleek91: On a separate note. Doesn't really seem like we're using `T` for anything. Should this just be `Route & { ... }`?

pleek91:
I really like "Resolved Route" as a term over "RouteFlat". Seems more intuitive.

src/types/resolved.ts
stackoverfloweth: moved T into `matched` property, I still prefer having the explicitness of how this is written over `Resolved` just being a standalone type

pleek91 (approved):

PR #27, opened 2023-12-22
created setParamValue function in paramsFinder service

stackoverfloweth:
works similar to what was named `findParamValues` but in reverse.
takes `path: string`, and an optional ParamReplace (solution to what would be 4 params and making all optional)
```ts
{
  name: string,
  params: Param[],
  values: unknown[],
}
```

finds that param in path and replaces with param setter value

I'm tempted to rename `paramsFinder` now that it has both `getParamValues` and `setParamValues`, any suggestions?

### DIFF
--- a/src/types/params.ts
+++ b/src/types/params.ts
@@ -12,8 +12,16 @@ export type ParamGetSet<T = any> = {
 
 export type Param = ParamGetter | ParamGetSet | RegExp | BooleanConstructor | NumberConstructor | StringConstructor
 
+function isNotConstructor(value: Param): boolean {
+  return value !== String && value !== Boolean && value !== Number
+}
+
 export function isParamGetter(value: Param): value is ParamGetter {
-  return typeof value === 'function' && value !== Boolean && value !== Number
+  return typeof value === 'function' && isNotConstructor(value)
+}
+
+export function isParamSetter(value: Param): value is ParamSetter {
+  return typeof value === 'function' && isNotConstructor(value)
 }
 
 export function isParamGetSet(value: Param): value is ParamGetSet {
--- a/src/types/params.ts
+++ b/src/types/params.ts
@@ -20,10 +20,6 @@ export function isParamGetter(value: Param): value is ParamGetter {
   return typeof value === 'function' && isNotConstructor(value)
 }
 
-export function isParamSetter(value: Param): value is ParamSetter {
-  return typeof value === 'function' && isNotConstructor(value)
-}
-
 export function isParamGetSet(value: Param): value is ParamGetSet {
   return typeof value === 'object'
     && 'get' in value
--- a/src/utilities/params.ts
+++ b/src/utilities/params.ts
@@ -1,4 +1,4 @@
-import { Param, ParamGetSet, ParamExtras, isParamGetSet, isParamGetter, ExtractParamType, isParamSetter } from '@/types'
+import { Param, ParamGetSet, ParamExtras, isParamGetSet, isParamGetter, ExtractParamType } from '@/types'
 import { InvalidRouteParamValueError } from '@/types/invalidRouteParamValueError'
 
 const extras: ParamExtras = {
@@ -102,20 +102,8 @@ export function setParamValue(value: unknown, param: Param): string {
     return numberParam.set(value, extras)
   }
 
-  if (isParamSetter(param)) {
-    return param(value, extras)
-  }
-
   if (isParamGetSet(param)) {
     return param.set(value, extras)
-  }
-
-  if (param instanceof RegExp) {
-    if (typeof value === 'string' && param.test(value)) {
-      return value
-    }
-
-    throw new InvalidRouteParamValueError()
   }
 
   try {
--- a/src/utilities/params.ts
+++ b/src/utilities/params.ts
@@ -1,4 +1,4 @@
-import { Param, ParamGetSet, ParamExtras, isParamGetSet, isParamGetter, ExtractParamType } from '@/types'
+import { Param, ParamGetSet, ParamExtras, isParamGetSet, isParamGetter, ExtractParamType, isParamSetter } from '@/types'
 import { InvalidRouteParamValueError } from '@/types/invalidRouteParamValueError'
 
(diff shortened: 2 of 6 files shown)


src/utilities/params.ts
```diff
@@ -102,12 +102,8 @@ export function setParamValue(value: unknown, param: Param): string {
...
   }
 
-  if (isParamGetter(param)) {
-    const stringValue = (value as any).toString()
-
-    param(stringValue, extras)
-
-    return stringValue
+  if (isParamSetter(param)) {
+    return param(value, extras)
```
pleek91, line 106: This change doesn't seem right to me. A param cannot be type `ParamSetter`. Only `ParamGetter`. So this check will return true (because both types have the same signature) but then most likely error because its being type guarded as a `ParamSetter` when its actually a `ParamGetter` and a `ParamGetter` expects a string as its value.

src/types/params.ts
```diff
@@ -12,8 +12,16 @@ export type ParamGetSet<T = any> = {
...
+  return value !== String && value !== Boolean && value !== Number
+}
+
 export function isParamGetter(value: Param): value is ParamGetter {
-  return typeof value === 'function' && value !== Boolean && value !== Number
+  return typeof value === 'function' && isNotConstructor(value)
+}
+
+export function isParamSetter(value: Param): value is ParamSetter {
+  return typeof value === 'function' && isNotConstructor(value)
```
pleek91, line 24: What's your thought here?

pleek91 (commented):
Two questions. Overall this looks nice. I really like how we have `getParamValue` and `getParamValues`. Though maybe `getParamValuesFromPath` crossed my mind. Idk that I have a preference. 

paramFinder.ts is fine I think? You could also combine these into `param.ts` 🤷

src/utilities/params.ts
stackoverfloweth: I see that `Param` cannot be `ParamSetter`, that makes sense. However getting a `ParamGetter` doesn't make sense for SETTING a param value, the contract is backwards. 

I felt super confident in that change when how we had to contort the value to call a getter.  But yeah, I'm not sure how ParamSetter would ever be used by a developer except within a ParamGetSet. 

If a developer provides ONLY a ParamGetter, I don't think we can assume it gives us anything useful for converting back into URL strings. By providing only a getter my expectation would be that all we do on the way back is call String constructor?

src/types/params.ts
stackoverfloweth: there might not be much utility in defining a ParamSetter type at all, so this function likely goes away? Let's see where we land in the other thread

src/utilities/params.ts
pleek91: If only a ParamGetter is provided by the user that means that whatever value they provide for the param when setting can be directly converted to a string. It goes straight into the url. 

But just because it can be directly converted to a string does not mean that string value will be valid when passed into the ParamGetter the user provided. 

So the question is, if a value is given is invalid based on the ParamGetter. When do we error? We either

1. Convert the provided value into a string, create a new url, and then try to match the url with a route. A valid string will match the requested route, an invalid string will not. So we get a NotFound rejection

2. Convert the provided value into a string. Check if it passes the getter and if it does go ahead and create the new url. If it does not error right away. URL does not get updated and no rejection happens. It's just a JavaScript runtime error. 

The second is what I had. Not because I think that's necessarily better (I think creating the url and rejecting is a good experience) but because it's consistent with other get/set patterns. A Boolean for example will error in the setter. And user land GetSets can also error in the setter. 

It also felt weird that if you had a ParamGetter than the unit tests wouldn't error the same way for the get vs the set. 

Totally open to a different implementation for handling sets of ParamGetter.

src/types/params.ts
pleek91: It only exists for ParamGetSet. 

But I'm mostly wondering about the isNotConstructor piece. Those technically are valid. From a types perspective. You just thinking they're different from a runtime perspective because we check for and override them?

src/utilities/params.ts
stackoverfloweth: Apologies, maybe I'm being dense here I am behind on sleep lol

You're losing me at "...if a value is given is invalid based on the ParamGetter...".  We already know the route, and the "values" given here refers to the values provided by the developer at runtime calling a routeMethod that ultimately need to be converted to strings so we can form a complete URL. So "values" will be whatever `T` is (boolean, number, etc) NOT the string value. 

If my route is 
```ts
{
  name: 'foo', 
  path: path('/foo/:bar', { bar: Boolean })
  ...
}
```

We're working through what happens when the developer later calls `router.routes.foo(true)`

"values" is `[true]` and needs to be converted to a string to form the URL "foo/true"

"ParamGetters" are responsible for converting a string value into `T`, which I'm not seeing how is supposed to help us here.

It seems like maybe you're confusing assembling a URL given a route and values (with setParamValue) with matching a URL given a url string (with getParamValue)?

src/types/params.ts
stackoverfloweth: ah, I see. I only pulled that out to not repeat myself in `IsParamSetter`, the logic inside was already defined in `IsParamGetter`. Though I did add "value !== String" since we now have a string param we use if we are given `String` for a param.

src/utilities/paramsFinder.ts
```diff
@@ -1,11 +1,44 @@
...
+}
+
+export function setParamValuesOnUrl(path: string, paramReplace?: ParamReplace): string {
+  if (!paramReplace) {
+    return path
+  }
+
+  const { name } = { ...paramReplace }
+  const params = paramReplace.params ?? []
+  const values = paramReplace.values ?? []
```
pleek91, line 24: I think this is the same and more concise yeah? If not just ignore. 
```suggestion
  const { name, params = [], values = [] } = { ...paramReplace }
```

pleek91 (approved):

src/utilities/paramsFinder.ts
stackoverfloweth: much better, thank you 🙏

PR #30, opened 2023-12-26
added parent route names to Resolved route

stackoverfloweth:
After this change `Resolved<Route>` will have `parentNames: string[]`. This is needed for the implementation of createRouteMethods, which takes a flat array of `Resolved<Route>` and needs to recreate the hierarchy of the original nested routes.


src/utilities/routeResolver.ts
```diff
@@ -1,17 +1,19 @@
-import { Resolved, Routes, isParentRoute, isNamedRoute, Route } from '@/types'
+import { Resolved, Routes, isParentRoute, isNamedRoute, Route, Param } from '@/types'
 import { mergeParams, path as createPath } from '@/utilities'
 
-export function resolveRoutes(routes: Routes, parentPath = '', parentParams = {}): Resolved<Route>[] {
+type RouteParentContext = {
+  parentPath: string,
+  parentParams: Record<string, Param[]>,
+  parentNames: string[],
+}
```
stackoverfloweth, line 8: this is how I chose to get around the max args count of 3

src/utilities/routeResolver.spec.ts
pleek91, line 1: This is unrelated yeah? Just combining two tests into one?

src/utilities/routeResolver.ts
pleek91: I like this pattern anyway where we have one required argument and then a second optional argument that has additional contextual stuff in it 👍

src/types/resolved.ts
```diff
@@ -6,5 +6,6 @@ export type Resolved<T extends Route> = {
   name: string,
   path: string,
   regex: RegExp,
+  parentNames: T['name'][],
```
pleek91, line 9: this feels a bit awkward to me. But I see why you went this direction. But I don't think this is something we'll want as part of the libraries public types and we've talked about `Resolved<Route>` being the type that is used for middleware and such. This seems like an implementation detail and not something we'd actually give to the user. Any thoughts on that?

src/utilities/routeResolver.spec.ts
stackoverfloweth: correct, big fan of `test.each`

src/types/resolved.ts
stackoverfloweth: agreed. Let's include this in our conversation about types this afternoon.

pleek91 (approved):

PR #33, opened 2023-12-30
Fix types for RouteMethodResponse

pleek91:
# Description
Fixes the types for `RouteMethodResponse` so that both the sync and async values are the same. 

An example of implementing this type is [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBASg9gV2BAshYALOATGEDOYcAdvtALxQDeAUFPVAgE4A2AXFPsEwJbEDmdBpgjEOABSZwAtjzIAZHgGsIAHgDys4KvhJU6LLgJFSEADRQA5COKWAfHYDa1jKMsBdGgF8aNAGYIxADGwDwkUFJ6ABQAlBy6yGiYOHiEJGTUQvRB6cARxukQHJo82gn6yUZpphYubnZQlLQMLYysHJZ+cHCWWVA+fTmkeTYAhgBGLEWwiIkGKQWmjZmtDMzsVl09Zn0tNhxQUUwEMY0NxBAA7lCSMnJqJWWzFYapJmS1NvZ2UUcEcCwAG4QU7kBpkYAAFR40ggs1+oIax3wAOBf2qZBiFgAjAAGfExGIAOhs6KxfQGLWOwGYxCgY0mEG8viGXGobRY-WWkWQsRorIBECJLDg-CilnwIGClgs6xivgA3FFRpLgodESsGKy8lQOV5lqNLqNShFnrEFYN0oLhaLxSqpUEZRz5V4YuagA)

### DIFF
--- a/src/types/routeMethods.test-d.ts
+++ b/src/types/routeMethods.test-d.ts
@@ -426,3 +426,18 @@ test('param names must be alphanumeric', () => {
     param4: string,
   }>()
 })
+
+test('route method returns correct typ when called', async () => {
+  const routes = [
+    {
+      name: 'foo',
+      path: '/foo',
+      component,
+    },
+  ] as const satisfies Routes
+
+  const router = createRouter(routes)
+
+  expectTypeOf(router.routes.foo()).toMatchTypeOf<{ url: string }>()
+  expectTypeOf(await router.routes.foo()).toEqualTypeOf<{ url: string }>()
+})
--- a/src/types/routeMethods.test-d.ts
+++ b/src/types/routeMethods.test-d.ts
@@ -427,7 +427,7 @@ test('param names must be alphanumeric', () => {
   }>()
 })
 
-test('route method returns correct typ when called', async () => {
+test('route method returns correct type when called', async () => {
   const routes = [
     {
       name: 'foo',
--- a/src/types/routeMethods.ts
+++ b/src/types/routeMethods.ts
@@ -14,9 +14,10 @@ export type RouteMethodOptions = {
   skipRouting: boolean,
 }
 
-export type RouteMethodResponse = PromiseLike<{
+export type RouteMethodResponse = {
   url: string,
-}>
+  then: PromiseLike<Omit<RouteMethodResponse, 'then'>>['then'],
+}
 
 export type RouteMethods<
   TRoutes extends Routes,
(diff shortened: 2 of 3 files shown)


src/types/routeMethods.ts
```diff
@@ -14,9 +14,10 @@ export type RouteMethodOptions = {
   skipRouting: boolean,
 }
 
-export type RouteMethodResponse = PromiseLike<{
+export type RouteMethodResponse = {
   url: string,
-}>
+  then: PromiseLike<Omit<RouteMethodResponse, 'then'>>['then'],
+}
```
pleek91, line 20: There are a few ways to write this but this is what I settled on. The other options mean we would need to define a type for what a route method returns minus the `then`. Something like this
```typescript
type RouteMethodResponse = { url: string }
type RouteMethodPromiseLike = RouteMethodResponse & PromiseLike<RouteMethodResponse>
```
And I didn't like having two types because one is always poorly named.

src/types/routeMethods.test-d.ts
```diff
@@ -425,4 +425,19 @@ test('param names must be alphanumeric', () => {
...
+    {
+      name: 'foo',
+      path: '/foo',
+      component,
+    },
+  ] as const satisfies Routes
+
+  const router = createRouter(routes)
+
+  expectTypeOf(router.routes.foo()).toMatchTypeOf<{ url: string }>()
```
pleek91, line 441: First time I'm using `toMatchTypeOf` which is a little looser than `toEqualTypeOf` because I could figure out how to match the `then`. So matching all but the `then` and then testing the thenable return type on the next line

src/types/routeMethods.test-d.ts
```diff
@@ -425,4 +425,19 @@ test('param names must be alphanumeric', () => {
     param3: string,
     param4: string,
   }>()
+})
+
+test('route method returns correct typ when called', async () => {
```
stackoverfloweth, line 430:
```suggestion
test('route method returns correct type when called', async () => {
```

src/types/routeMethods.ts
stackoverfloweth: would this be the same?

```ts
type ThenableResponse<T> = T & {
  then: PromiseLike<T>,
}

export type RouteMethodResponse = ThenableResponse<{
  url: string,
}>
```

I don't feel strongly, but I do think something like above is simpler

pleek91: Close. It would be T & PromiseLike<T>. but then the return type for a route method would be Thenable rather than RouteMethodResponse. I don't feel strongly but named responses are nice over general types

pleek91: After thinking about it more I think having a `Thenable` type would be nice. We'll probably use it for other types as well.

pleek91: Updated

stackoverfloweth (approved):

PR #35, opened 2023-12-30
Create route methods

stackoverfloweth:


src/utilities/createRouteMethods.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    route.matches.forEach(match => {
+      if (!match.name) {
+        return
+      }
+
+      if (match === route.matched && isPublicRoute(route.matched)) {
+        level[route.name] = createCallableNode(route)
+        return
+      }
```
pleek91, line 16: Originally this had `Object.assign` with what might exist already at `level[route.name]` but that should never be the case. If we get here we're at a leaf. And the tests all pass so unless we're missing a test I think this is all we need.

pleek91: Just realized I was only looking at the type tests... The unit tests indeed are broken. Looking into that.

pleek91: The `Object.assign` was necessary. Put that back 👍

pleek91 (approved):

PR #36, opened 2023-12-30
Fix route methods showing up in types with `never`

pleek91:
# Description
Previously if a leaf was not public it would still show up in the `RouteMethods` type but with the type of `never`. This tweaks the `Identity` utility type to remove properties if their type is `never` and wraps the `RouteMethods` type definition in `Identity` to remove these properties.


stackoverfloweth (approved):
this makes sense, how much do you care that the implementation currently does not match this? Should we create an issue in github for things like that as potential things to address before release?

pleek91:
> this makes sense, how much do you care that the implementation currently does not match this? Should we create an issue in github for things like that as potential things to address before release?

@stackoverfloweth probably worth tracking. Will create an issue

PR #37, opened 2023-12-31
Add `route` to `Router` type and abstract router logic into separate utilities

pleek91:


src/utilities/createRouter.ts
```diff
@@ -1,46 +1,54 @@
...
+  const push: RouterPush = async (url, options) => {
+    const match = routeMatch(resolvedWithRegex, url)
 
-    return route
-  }
+    if (!match) {
+      const method = options?.replace ? window.location.replace : window.location.assign
+      method(url)
+      return
+    }
```
pleek91, line 37: crude implementation but should work just fine (need to test this)

src/utilities/createRouter.ts
```diff
@@ -1,60 +1,68 @@
...
-  const back: Router<T>['forward'] = () => {
-    throw 'not implemented'
+  const back: RouterNavigation = (number = -1) => {
+    return go(number)
   }
 
-  const go: Router<T>['go'] = (_number) => {
+  const go: RouterNavigation = (_number) => {
     throw 'not implemented'
   }
```
pleek91, line 56: These are actually final implementations I believe 🎉

src/utilities/resolveRoutesRegex.ts
pleek91, line 1: Moved this into its own utility with a dedicated type

src/utilities/routeMatch.ts
pleek91, line 1: Moved this into its own utility and just passing in the routes. This makes this much simpler to test and also moves logic out of the `createRouter` utility.

src/utilities/createRouter.ts
```diff
@@ -1,60 +1,68 @@
+import { readonly } from 'vue'
 import { Resolved, Route, RouteMethods, Routes } from '@/types'
-import { createRouteMethods, generateRouteRegexPattern, resolveRoutes, routeParamsAreValid } from '@/utilities'
+import { createRouteMethods, resolveRoutes } from '@/utilities'
+import { resolveRoutesRegex } from '@/utilities/resolveRoutesRegex'
+import { routeMatch } from '@/utilities/routeMatch'
```
stackoverfloweth, line 5: nit: we already have an import from `@/utilities`, this could just be combined?

src/utilities/createRouter.ts
```diff
@@ -1,60 +1,68 @@
...
   TRoutes extends Routes
 > = {
   routes: RouteMethods<TRoutes>,
-  push: (url: string) => void,
-  replace: (url: string) => void,
-  back: () => void,
-  forward: () => void,
-  go: (number: number) => void,
-  routeMatch: (path: string) => Resolved<TRoutes[number]> | undefined,
+  route: Readonly<Resolved<Route>>,
```
stackoverfloweth, line 15: is this the "current route"? I expected that to be possibly undefined?

src/utilities/createRouter.ts
stackoverfloweth: love to see it, even if it is crude. How do you feel about moving these out so this `createRouter` doesn't get too large?

stackoverfloweth (approved):
things are getting fun!

src/utilities/createRouter.ts
pleek91: Yeah I think pretty much all logic will get moved out of createRouter. But not sure how these methods will work with routers internals yet.

src/utilities/createRouter.ts
pleek91: It could, don't really feel strongly about that personally

src/utilities/createRouter.ts
pleek91: Yes it's the current route. Which should never be undefined? When the app loads there should be a route right? 

Unless there's a rejection 🤔 is that what you're thinking?

stackoverfloweth: I guess my thinking was more in error states, but I don't know that it's worth having to check for undefined everywhere in the event that the user might be in one of those "computed routes" where something like not-found, or needs-auth is controlling the route. I didn't know if those were still going to be legit "routes"

pleek91: It's a great question that I hadn't thought of. I don't know the answer. Will have to think about that.

PR #38, opened 2023-12-31
Abstract updating the window location into a separate utility and add tests

pleek91:


src/utilities/updateWindowLocation.spec.ts
```diff
@@ -0,0 +1,20 @@
+// @vitest-environment happy-dom
```
pleek91, line 1: worth calling out I'm targeting happy-dom as the environment for this test file specifically which allows testing window.location. The other tests are still running in the node environment which I think we want to keep because of ssr support which we will need to support.

stackoverfloweth (approved):
this looks fine aside from accessing `window` directly, in my history branch I created `utilities/window.ts`, which gives a safe way of accessing window. Not sure if you'd wanna bring that into this PR?

```ts
// https://stackoverflow.com/a/62557418/3511012
function globalExists(varName: string): boolean {
  const globalEval = eval
  try {
    globalEval(varName)
    return true
  } catch {
    return false
  }
}

export function getWindow(): Window {
  if (!globalExists('window')) {
    throw 'mock for window not implemented'
  }

  return window
}
```

pleek91:
@stackoverfloweth I was thinking about that but decided to put that off and see how we're going to handle routing without the window (routing in node). I'm not sure this method will even get called or if we'll call it still but it'll do nothing.

PR #40, opened 2023-12-31
Fix test errors with creating route methods

pleek91:
# Description
- Fixes test errors from typescript error introduced by https://github.com/kitbagjs/router/pull/36 
- Unifies test files to use "spec" keyword. 
- Updates createRouteMethod and tests to reflect types by removing empty objects


src/.eslintrc
pleek91, line 1: I think using ts comments is going to be necessary in type files especially to test negatives. So I don't think we need to link for them in spec files.

src/utilities/createRouteMethods.spec.ts
pleek91, line 1: Typescript sees that `isPublic` could be `false` so the type so it complains. Probably some type guarding we could do but didn't seem worth introducing for these tests.

src/utilities/createRouteMethods.ts
pleek91, line 1: This makes sure we don't add empty objects for leaves that are not callable.

stackoverfloweth (approved):

PR #41, opened 2024-01-03
Better browser support for updating urls

pleek91:
# Description
Replaces the `updateWindowLocation` utility with an `updateBrowserUrl` utility which automatically uses `history` or `location` apis based on origin. 

Introduces an `isBrowser` const which is used to test if the browser can even be updated. 

Add's environments to the vitest config so any file with `.browser.spec.ts` automatically runs in a `happy-dom` environment. `.spec.ts` files continue to run in `node`.


src/utilities/isBrowser.ts
pleek91, line 1: Suggesting this as an alternative to the `globalExists` pattern. Thinking avoiding `eval` is probably a good idea for a public repository.

src/utilities/updateBrowserUrl.browser.spec.ts
```diff
@@ -0,0 +1,35 @@
+
+import { expect, test, vi } from 'vitest'
+import { updateBrowserUrl } from '@/utilities/updateBrowserUrl'
+
+test('updates the window location', () => {
+  vi.spyOn(window.location, 'assign').mockImplementation(() => {})
```
pleek91, line 6: these `mockImplementation` calls make sure the actual location isn't updated. I tried a few different attempts to keep the location changes isolated to each test but this was the only thing I tried that worked.

src/utilities/isBrowser.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/prefer-optional-chain
+export const isBrowser: boolean = typeof window !== 'undefined' && typeof window.document !== 'undefined'
```
stackoverfloweth, line 2: is this better than the `globalExists` utility used in prefect-design? I feel like that was even safer, as this presumes accessing `window` won't crash

vite.config.js
```diff
@@ -3,6 +3,12 @@ import { defineConfig } from 'vitest/config'
 import dts from 'vite-plugin-dts'
 
 export default defineConfig({
+  test: {
+    environmentMatchGlobs: [
+      ['**\/*.browser.spec.ts', 'happy-dom'],
+      ['**\/*.spec.ts', 'node'],
+    ]
```
stackoverfloweth, line 10: this is clever, nicely done

src/utilities/updateBrowserUrl.browser.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+test('updates the history', () => {
+  vi.spyOn(history, 'pushState').mockImplementation(() => {})
+
+  const url = '/foo'
+
+  updateBrowserUrl(url)
+
+  expect(history.pushState).toHaveBeenCalledWith({}, '', url)
+})
```
stackoverfloweth, line 32: I feel like there is some overlap here with the service I'm working on. It might make sense in the future to update this service/test to call our local history service and have tests over there that ensure that it properly sends updated state over to browser history as well

src/utilities/updateBrowserUrl.ts
```diff
@@ -0,0 +1,39 @@
+import { isBrowser } from '@/utilities/isBrowser'
+
+type UpdateBrowserUrlOptions = {
+  replace?: boolean,
+}
```
stackoverfloweth, line 5: I have a virtually identical type in my history service. Is it worth trying to combine these?

src/utilities/updateBrowserUrl.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return window.location.assign(url)
+}
+
+function isSameOrigin(url: string): boolean {
+  const { origin } = new URL(url, window.location.origin)
+
+  return origin === window.location.origin
+}
```
stackoverfloweth, line 39: this feels like one of the more reusable functions we've stuck in utilities. Is it worth pulling this out?

src/utilities/updateBrowserUrl.ts
```diff
@@ -0,0 +1,39 @@
...
+  return updateWindow(url, options)
+}
+
+function updateHistory(url: string, options: UpdateBrowserUrlOptions): void {
+  if (options.replace) {
+    return history.replaceState({}, '', url)
+  }
+
+  history.pushState({}, '', url)
+}
```
stackoverfloweth, line 25: again just calling out that this responsibility probably should/will get moved out in favor of something like `routerHistory.update(url, options)`

src/utilities/updateBrowserUrl.ts
```diff
@@ -0,0 +1,39 @@
+import { isBrowser } from '@/utilities/isBrowser'
+
+type UpdateBrowserUrlOptions = {
+  replace?: boolean,
+}
+
+export function updateBrowserUrl(url: string, options: UpdateBrowserUrlOptions = {}): void {
```
stackoverfloweth, line 7: on a popstate event where the browser is telling us the user navigated outside of our router's control, would you imagine we call this function?

src/utilities/updateBrowserUrl.ts
pleek91: If we need it somewhere else sure

src/utilities/updateBrowserUrl.ts
pleek91: Let's sync up on this. I didn't imagine history being in charge of updating the url. But just tracking history itself.

src/utilities/updateBrowserUrl.ts
pleek91: if a popstate event is emitted the browser already has an updated url. So we wouldn't need to call this.

src/utilities/isBrowser.ts
pleek91: Idk, as long as we have tests that don't error in a browser environment or in a node environment I think this is sufficient. I think it's preferred to not use `eval` unless we absolutely have to.

PR #42, opened 2024-01-04
RouterNavigation implementation for web & memory

stackoverfloweth:
this PR also updates `isBrowser` from being a const to being a function. This makes it simpler to mock the implementation without having to setup different test files that will rely on actual implementation of isBrowser


package.json
```diff
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@prefecthq/eslint-config": "^1.0.31",
+    "@testing-library/dom": "^9.3.3",
```
stackoverfloweth, line 32: I added this at one point to simulate popstate events. That's no longer needed in this PR, but could be coming? I'll let you decide if this should be removed

pleek91: If it's not used in this PR let's remove it for now.

src/utilities/routerNavigation.browser.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      history.update(url)
+
+      expect(utilities.updateBrowserUrl).toHaveBeenCalledWith(url)
+    })
+  })
+
+  describe('when isBrowser is false', () => {
+    beforeEach(() => {
+      vi.spyOn(isBrowser, 'isBrowser').mockImplementation(() => false)
+    })
```
pleek91, line 55: So rather than mocking this why not run the tests in node? Two test files. One node and one browser. Test what needs to be tested in each. 

That way it's a true test running code in the correct environment and no mocking is needed.

src/utilities/routerNavigation.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+function createMemoryNavigation(): RouterNavigation {
+  return {
+    go: () => {},
+    back: () => {},
+    forward: () => {},
+    update: updateBrowserUrl,
+  }
```
pleek91, line 30: Can these just throw not implemented for now. IMO we do want the go, forward, and back methods to error. If you're running in node and you call one of these something's wrong and not going to work as expected.

src/utilities/routerNavigation.ts
```diff
@@ -0,0 +1,31 @@
+import { isBrowser } from '@/utilities/isBrowser'
+import { updateBrowserUrl } from '@/utilities/updateBrowserUrl'
+
+type RouterNavigation = {
+  forward: () => void,
+  back: () => void,
+  go: (delta: number) => void,
+  update: (url: string) => void,
```
pleek91, line 8: This will need to accept options with `replace?: boolean` right?

PR #43, opened 2024-01-06
Update navigation and implement onLocationUpdate logic

pleek91:
# Description
Makes `updateBrowserUrl` a promise. Never resolves when the update is not same origin. 

Updates `createRouterNavigation` to better support both node and browser environments

Updates `createRouter` to implement navigation and simplify implementation of router navigation methods.


src/utilities/updateBrowserUrl.spec.ts
pleek91, line 1: Deleted this test because we decided `updateBrowserUrl` shouldn't even get called if `isBrowser()` is `false`.

stackoverfloweth (approved):

PR #44, opened 2024-01-07
Implement router.route

pleek91:
# Description
sets the initial route and updates the route when navigation happens.


src/utilities/createRouter.ts
```diff
@@ -1,41 +1,70 @@
...
+export function createRouter<T extends Routes>(routes: T, options?: RouterOptions): Router<T> {
   const resolved = resolveRoutes(routes)
   const resolvedWithRegex = resolveRoutesRegex(resolved)
   const navigation = createRouterNavigation({
     onLocationUpdate,
   })
 
-  // todo: implement this
-  const route: Router<T>['route'] = readonly({} as any)
+  const route: Resolved<Route> = reactive(getInitialRoute())
```
stackoverfloweth, line 33: I think this makes sense, just wanted to call out that we could not support having an `undefined` router.route. So if something errors or we're handling not-found, we need to maybe assign this to some internal route that we defined as a fallback

src/utilities/createRouter.ts
```diff
@@ -1,41 +1,70 @@
...
-  async function onLocationUpdate(url: string): Promise<void> {
-    const match = routeMatch(resolvedWithRegex, url)
+  function getInitialUrl(): string {
+    if (options?.initialUrl) {
+      return options.initialUrl
+    }
+
+    if (isBrowser()) {
+      return window.location.toString()
+    }
```
stackoverfloweth, line 42: I have a gut feeling that when I see `isBrowser` I want to pull it out into it's own service. I can see that this is maybe a bit extreme here but putting the comment in anyways to get your feedback. The fewer and smaller the tests that require a node and browser version the better IMO. I can supply a PR that makes this change for you to consider

stackoverfloweth: https://github.com/kitbagjs/router/pull/45

src/utilities/createRouter.ts
pleek91: Correct. When a not found or a rejection happens we'll still need to create a resolved route. I've got some ideas for that and I might work on that soon.

stackoverfloweth (approved):

PR #45, opened 2024-01-07
code review suggestions

stackoverfloweth:


src/utilities/currentLocation.ts
pleek91, line 1: What's this?

src/utilities/getInitialRoute.browser.spec.ts
pleek91, line 1: Unless we're importing getInitialUrl and testing that specifically this is still testing createRouter. So I wouldn't rename these test files.

src/utilities/getInitialRoute.browser.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { describe, expect, test } from 'vitest'
+import { createRouter } from '@/utilities/createRouter'
+import { component } from '@/utilities/testHelpers'
+
+describe('getInitialRoute', () => {
```
pleek91, line 5: Do you think every test should be wrapped in a describe? Vitest automatically groups tests by their file so IMO wrapping all the tests in a describe with the name of the utility being tested (which is the same as the name of the file) is not needed and we should reserve describe for organizing larger test files where we want to group tests of similar nature. If every test is in a single describe we don't need it IMO.

src/utilities/getInitialRoute.ts
pleek91, line 1: Abstracting this means less code in createRouter so I'm fine with this. But I don't think an isBrowser existing in a function means it much be abstracted.

src/utilities/getInitialUrl.browser.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const response = getInitialUrl(initialRoute)
+
+
+    expect(response).toBe(initialRoute)
+  })
+
+  test('defaults to window.location', () => {
+    const initialRoute = random.number().toString()
+    vi.stubGlobal('location', initialRoute)
```
pleek91, line 17: Nice, TIL. This will be helpful.

pleek91 (approved):

PR #47, opened 2024-01-08
add support for vue components as part of library export

stackoverfloweth:
this PR adds router-view, but only as stub, the real value of this PR is all the supporting infrastructure needed to export vue components with the library

this PR also removed the script and github workflow to run `tsc --no-emit` per the recommendation by vitest at the bottom of [this page](https://vitest.dev/guide/testing-types#testing-types)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,14 +7,6 @@ jobs:
     name: Type Validation
     runs-on: ubuntu-latest
 
-    steps:
-      - name: Setup
-        id: setup
-        uses: kitbagjs/actions-setup-project@main
-
-      - name: Validate types
-        run: npm run types
-
   unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,6 @@ name: Tests
 on: [pull_request]
 
 jobs:
-  typescript:
-    name: Type Validation
-    runs-on: ubuntu-latest
-
   unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
--- a/src/utilities/createRouter.ts
+++ b/src/utilities/createRouter.ts
@@ -1,4 +1,5 @@
-import { DeepReadonly, reactive, readonly } from 'vue'
+import { DeepReadonly, reactive, readonly, App } from 'vue'
+import { RouterView } from '@/components'
 import { Resolved, Route, RouteMethods, Routes } from '@/types'
 import { createRouteMethods, createRouterNavigation, resolveRoutes, routeMatch, getInitialUrl, resolveRoutesRegex } from '@/utilities'
 
@@ -19,6 +20,7 @@ export type Router<
   back: () => void,
   forward: () => void,
   go: (delta: number) => void,
+  install: (app: App) => void,
 }
 
 export function createRouter<T extends Routes>(routes: T, options: RouterOptions = {}): Router<T> {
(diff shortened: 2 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -3,18 +3,6 @@ name: Tests
 on: [pull_request]
 
 jobs:
-  typescript:
```
pleek91, line 6: Just gonna let vitest verify types yeah?

src/utilities/createRouter.ts
pleek91, line 1: Nice

tsconfig.json
```diff
@@ -20,5 +20,6 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true
   },
-  "include": ["src"]
+  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.tsx", "src/**/*.vue"],
```
pleek91, line 23: We don't need .tax do we?

tsconfig.json
```diff
@@ -20,5 +20,6 @@
...
+  "exclude": ["node_modules", "dist", "public"]
```
pleek91, line 24: We don't have a public folder right?

vite.config.js
```diff
@@ -1,9 +1,13 @@
 import { resolve } from 'path'
 import { defineConfig } from 'vitest/config'
+import vue from '@vitejs/plugin-vue'
 import dts from 'vite-plugin-dts'
 
 export default defineConfig({
   test: {
+    typecheck: {
+      checker: 'vue-tsc',
+    },
```
pleek91, line 10: Interesting, haven't seen this before.

.github/workflows/test.yml
stackoverfloweth: yeah, I put it in the desc for the PR, this is what vitest recommends

> this PR also removed the script and github workflow to run tsc --no-emit per the recommendation by vitest at the bottom of [this page](https://vitest.dev/guide/testing-types#testing-types)

tsconfig.json
stackoverfloweth: you're right, this is just copy pasta from new vite project

tsconfig.json
stackoverfloweth: you're right, this is just copy pasta from new vite project

vite.config.js
stackoverfloweth: typechecking was failing on import of .vue file, this solved it

.github/workflows/test.yml
pleek91: Cool. What about this? Do we even need that package? 
> Vitest uses tsc --noEmit or vue-tsc --noEmit, depending on your configuration, so you can remove these scripts from your pipeline.

stackoverfloweth: I don't follow, what package are you referring to?

pleek91: Sorry, vue-tsc. Seems like vitest already had that as a dependency? Idk just wondering

stackoverfloweth: vitest is setup to use it, but I don't know that it will install for you. We will need it for our build script too

pleek91: Remind me why we need it for the build?

stackoverfloweth: The build before this PR called `tsc && vite build`. Calling tsc will error on .vue import.

pleek91: Is that for type checking or for building types? Because we also have vite-dts which bundles the types for dist.

stackoverfloweth: Type checking for sure, not sure about building.

pleek91: Wonder if we even need it. We already test the types are valid with vitest. Maybe we don't need that as part of the build step. 

Approved either way.

stackoverfloweth: I don't think vitest can properly test types if vue files are in the mix.

PR #48, opened 2024-01-09
Implement RouterView

pleek91:
# Description
Adds a basic implementation for the `RouterView` component. Changed the existing boilerplate component from a sfc to a basic typescript component with a render function. 

- Removed `vue-tsc` as a dependency (everything seems to work fine without it)
- Added `@vue/test-utils`

Notes:
I don't like how in `RouterView` the types for the route component need some coercion. Seems like its getting an expanded type for some reason and typescript cannot sort it out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   },
   "homepage": "https://github.com/kitbagjs/router#readme",
   "scripts": {
-    "build": "vue-tsc && vite build",
+    "build": "vite build",
     "test": "vitest",
     "test:types": "vitest typecheck",
     "lint": "eslint ./src",
@@ -29,14 +29,14 @@
   "devDependencies": {
     "@prefecthq/eslint-config": "^1.0.31",
     "@vitejs/plugin-vue": "^5.0.2",
+    "@vue/test-utils": "^2.4.3",
     "eslint": "^8.54.0",
     "happy-dom": "^12.10.3",
     "typescript": "^5.0.2",
     "vite": "^5.0.11",
     "vite-plugin-dts": "^3.6.3",
     "vitest": "^0.34.6",
-    "vue": "^3.4.5",
-    "vue-tsc": "^1.8.27"
+    "vue": "^3.4.5"
   },
   "peerDependencies": {
     "vue": "^3.4.5"
--- a/src/components/helloWorld.ts
+++ b/src/components/helloWorld.ts
@@ -1,0 +1,7 @@
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+  name: 'HelloWorld',
+  template: 'hello world',
+  expose: [],
+})
--- a/src/components/routerView.browser.spec.ts
+++ b/src/components/routerView.browser.spec.ts
@@ -1,0 +1,107 @@
+import { mount } from '@vue/test-utils'
+import { expect, test } from 'vitest'
+import { Route } from '@/types'
+import { createRouter } from '@/utilities'
+
+test('renders component for initial route', () => {
+  const route = {
+    name: 'parent',
+    path: '/',
+    component: { template: 'hello world' },
+  } as const satisfies Route
+
+  const router = createRouter([route], {
+    initialUrl: route.path,
+  })
+
+  const root = {
+    template: '<RouterView/>',
+  }
+
+  const app = mount(root, {
+    global: {
+      plugins: [router],
+    },
+  })
+
+  expect(app.html()).toBe(route.component.template)
+})
+
+test('renders components for initial route', () => {
+  const childRoute = {
+    name: 'child',
+    path: 'child',
+    component: { template: 'Child' },
+  } as const satisfies Route
+
+  const parentRoute = {
+    name: 'parent',
+    path: '',
+    component: { template: '<RouterView />' },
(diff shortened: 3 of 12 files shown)


src/components/routerView.browser.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  expect(app.html()).toBe(childB.component.template)
+
+  await router.push(childC.path)
+
+  expect(app.html()).toBe(childC.component.template)
+
+  await router.push(childA.path)
+
+  expect(app.html()).toBe(childA.component.template)
```
pleek91, line 105: This test seems to always fail whenever I `router.push` back to the initial url. I can change up which route is used for the initial url and then this will error in a different place.

src/components/routerView.ts
pleek91, line 1: I'm actually quite surprised this works as well as it does.

src/components/routerView.ts
```diff
@@ -0,0 +1,30 @@
...
+    const routerView = resolveComponent('RouterView')
+
+    const component = computed(() => {
+      const route = router.route.matches[depth]
+
+      return route.component ?? routerView
+    })
+
+    // @ts-expect-error
+    return () => h(component.value)
```
pleek91, line 28: Not sure whats up with this. Will have to dig into the types here. Added this just to test other theories.

src/utilities/createRouter.ts
```diff
@@ -11,7 +11,7 @@ type RouterPush = (url: string, options?: { replace: boolean }) => Promise<void>
 type RouterReplace = (url: string) => Promise<void>
 
 export type Router<
-  TRoutes extends Routes
+  TRoutes extends Routes = []
```
pleek91, line 14: Added this default so that the injection key on line 26 wouldn't need the empty `[]`. All the type tests passed so think this is fine.

package.json
```diff
@@ -7,7 +7,7 @@
   },
   "homepage": "https://github.com/kitbagjs/router#readme",
   "scripts": {
-    "build": "vue-tsc && vite build",
+    "build": "vite build",
```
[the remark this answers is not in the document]
pleek91: As long as we have type testing that fails if any types fail idk that we need to stop a build from building if types fail. Its often useful for debugging to be able to build even with a type error. Do you feel strongly we should prevent builds with type errors? 

I think of this similarly to unit tests. Both types and code tests can be failing but the build can still run. And I don't think we'd stop a build if a unit test was failing. But we wouldn't publish a new version. 

I don't feel strongly here. Just trying to keep the repo as minimal as possible and not have any dependencies we don't absolutely need.

package.json
```diff
@@ -29,14 +29,14 @@
...
+    "@vue/test-utils": "^2.4.3",
     "eslint": "^8.54.0",
     "happy-dom": "^12.10.3",
     "typescript": "^5.0.2",
     "vite": "^5.0.11",
     "vite-plugin-dts": "^3.6.3",
     "vitest": "^0.34.6",
-    "vue": "^3.4.5",
-    "vue-tsc": "^1.8.27"
+    "vue": "^3.4.5"
```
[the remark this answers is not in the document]
pleek91: We won't need it unless we run it in ci. vite with the vue plugin can build and output types just fine

src/components/helloWorld.ts
```diff
@@ -0,0 +1,7 @@
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+  name: 'HelloWorld',
+  template: 'hello world',
+  expose: [],
+})
```
stackoverfloweth, line 7: are you planning on shipping this?

src/components/routerView.ts
```diff
@@ -9,22 +10,28 @@ export default defineComponent({
...
     const component = computed(() => {
-      const route = router.route.matches[depth]
+      const routeComponent = router.route.matches[depth].component
+
+      if (routeComponent) {
+        if (typeof routeComponent === 'function') {
+          return defineAsyncComponent(routeComponent as AsyncComponentLoader)
+        }
+
+        return routeComponent as RouteComponent
```
stackoverfloweth, line 29: in my testing this line doesn't need `as RouteComponent`

src/components/helloWorld.ts
pleek91: I think we'll need something like this for our unit tests. We should probably make a folder or something for files that are just for tests. Thoughts?

src/components/routerView.ts
pleek91: Nice, idk what I changed to make that work now but happy to remove an `as`.

src/components/helloWorld.ts
stackoverfloweth: yeah, I guess it's no more offensive than `@/utilities/testHelpers`

stackoverfloweth (approved):

src/components/helloWorld.ts
pleek91: right, I just couldn't put this there. Well I guess I could have as the default export but that feels weird. I vote we keep moving as as we have more stuff like this maybe we find a spot for this stuff. We already wanna organize the repo more at some point.

PR #50, opened 2024-01-10
created overload of router.push

stackoverfloweth:
this overload takes `{ name: T['name'] }` and params and attempts to assemble URL. In my testing though, I was disappointed by the lack of intellisense


src/utilities/createRouter.ts
```diff
@@ -1,21 +1,26 @@
...
+import { createRouteMethods, createRouterNavigation, resolveRoutes, routeMatch, getInitialUrl, resolveRoutesRegex, assembleUrl } from '@/utilities'
 
 type RouterOptions = {
   initialUrl?: string,
 }
 
-type RouterPush = (url: string, options?: { replace: boolean }) => Promise<void>
+type RouterPushOptions = { replace?: boolean }
+type RouterPush<T extends Routes> = {
+  (route: { name: T[number]['name'] }, params?: Record<string, unknown>, options?: RouterPushOptions): Promise<void>,
```
pleek91, line 12: I think we should be able to type the params yeah?

src/utilities/createRouter.ts
```diff
@@ -59,8 +64,22 @@ export function createRouter<T extends Routes>(routes: T, options: RouterOptions
...
+      if (!route) {
+        throw `No route found with name "${urlOrRoute.name}"`
+      }
+
+      const url = assembleUrl(route, optionsOrParams)
+
+      return push(url, possiblyOptions)
+    }
+
+    await navigation.update(urlOrRoute, optionsOrParams)
```
pleek91, line 82: I wonder if we can abstract the implementation of this into another file. createRouter is so small and easy to read. Be nice to keep logic out of it. Just a thought. Not an action item.

src/utilities/createRouter.spec.ts
pleek91, line 1: Should we have a separate set of tests for router.push?

src/utilities/createRouter.ts
```diff
@@ -45,8 +44,20 @@ export function createRouter<T extends Routes>(routes: T, options: RouterOptions
...
   }
 
-  const push: RouterPush = async (url, options) => {
-    await navigation.update(url, options)
+  const push: RouterPush<T> = (urlOrRoute, optionsOrParams: Record<string, any> = {}, possiblyOptions: Record<string, any> = {}) => {
+    if (typeof urlOrRoute === 'string') {
+      return navigation.update(urlOrRoute, optionsOrParams)
+    }
+
+    const route = resolved.find(({ name }) => name === urlOrRoute.name)
```
pleek91, line 52: This won't work for a nested route. Because `Flattened<Routes>` will produce things `parent.child.grandchild` which won't match a resolve route.

We could probably use `name.split('.').at(-1)` to match the current behavior.

Though matching a resolved route by its name means routes couldn't have the same names like `parentA.child` and `parentB.child`. So far we've said route names do not need to be unique. Idk if that's worth worrying about here.

src/types/router.ts
```diff
@@ -13,7 +14,10 @@ export type RouterPushOptions = {
   replace?: boolean,
 }
 
-export type RouterPush = (url: string, options?: RouterPushOptions) => Promise<void>
+export type RouterPush<TRoutes extends Routes = Routes> = {
+  (url: string, options?: RouterPushOptions): Promise<void>,
+  <TRoute extends keyof Flattened<TRoutes>>(route: { name: TRoute }, params?: Flattened<TRoutes>[TRoute], options?: RouterPushOptions): Promise<void>,
```
pleek91, line 19: I'm thinking through the devx/syntax of these three different ways of navigating to a route. 
- Route method
- Route Flat (`parent.child` & params)
- Url
I think we might want the route name to include route params and options like
```
{
  route: 'parent.child',
  params: {},
  query: {},
}
```
That way the overload becomes
```
// this type needs generics for Flattened
type RouterPushRoute = {
  route: T,
  params: TRoutes<T>,
  query: Record<string, unknown>,
  replace?: boolean,
}

// this need generics as well
type RouterPush = {
  (url: string, options?: RouterPushOptions} => Promise<void>,
  (route: RouterPushRoute) => Promise<void>
  (routeMethod => () => ReturnType<RouteMethod>) => Promise<void>
}
```
That way the arguments are unique types and we can type guard easily. 

And then we can match the same syntax with `RouterLink` where anything `router.push` accepts `RouterLink.to` can accept as well. This might not make sense when reading. Happy to chat through this.

pleek91 (commented):
Seeing that 'possibly infinite' type error in `createRouter` again 😬

src/types/router.ts
stackoverfloweth: I don't feel strongly so I will make this change 👍

src/utilities/createRouter.ts
stackoverfloweth: great call, I've got a service that takes Resolved<Route>, combines it's `.matches` to create a string that we can compare the full history. That should retain support for leaf names that aren't unique

pleek91: Great idea

src/utilities/createRouter.ts
```diff
@@ -45,18 +43,38 @@ export function createRouter<T extends Routes>(routes: T, options: RouterOptions
...
+  async function replace(url: string, options: RouterReplaceOptions = {}): Promise<void> {
+    await navigation.update(url, { ...options, replace: true })
   }
 
   const router = {
-    routes: createRouteMethods<T>(resolved, push),
+    routes: createRouteMethods<T>(resolved, pushUrl),
     route: readonly(route),
-    push,
+    push: push as RouterPush<T>,
```
pleek91, line 77: This is pretty much the same as what we did for `createRouteMethods` where we just tell typescript what it is. We could probably abstract the implementation out into a `createRouterPush` utility that hides the type casting. 

Though with some fresh eyes I might be able to get rid of this.

src/types/router.ts
```diff
@@ -13,7 +14,20 @@ export type RouterPushOptions = {
...
 }
 
-export type RouterPush = (url: string, options?: RouterPushOptions) => Promise<void>
+type RouterPushRoute<
+  TRoutes extends Routes,
+  TRoute extends PropertyKey
+> = RouterPushOptions & {
+  name: TRoute,
+  params?: TRoute extends keyof Flattened<TRoutes> ? Flattened<TRoutes>[TRoute] : Record<never, never>,
+}
```
pleek91, line 23: I realized basically typescript didn't like whenever we passed `Flattend<TRoutes>` into another type as the generic. So "flattening" the types fixed the infinite depth error that was in this file.

pleek91 (approved):

PR #51, opened 2024-01-12
Implement route methods and organize router and route method types

pleek91:
# Description
- Moved `Router` and associated types into `@/types/router`
- Moved `RouteMethod` and associated types into `@/types/routeMethod`
- Updated `RouteMethod` type to match what we discussed at standup
- Updated `createRouteMethods` to implement the route methods to match the expected type
- Wrote unit tests

### DIFF
--- a/src/compositions/useRoute.ts
+++ b/src/compositions/useRoute.ts
@@ -1,4 +1,5 @@
-import { ExtractRouteMethodParams, RouteMethod } from '@/types/routeMethods'
+import { RouteMethod } from '@/types/routeMethod'
+import { ExtractRouteMethodParams } from '@/types/routeMethods'
 import { Identity } from '@/types/utilities'
 
 type Route<T extends RouteMethod> = {
--- a/src/types/routeMethod.ts
+++ b/src/types/routeMethod.ts
@@ -1,0 +1,18 @@
+import { IsEmptyObject } from '@/types/utilities'
+
+export type RouteMethod<
+  TParams extends Record<string, unknown> = any
+> = IsEmptyObject<TParams> extends false
+  ? (params: TParams, options?: RouteMethodOptions) => RouteMethodResponse
+  : (options?: RouteMethodOptions) => RouteMethodResponse
+
+export type RouteMethodOptions = {
+  replace: boolean,
+  skipRouting: boolean,
+}
+
+export type RouteMethodResponse = {
+  url: string,
+  // push: RouteMethodPush,
+  // replace: RouteMethodReplace,
+}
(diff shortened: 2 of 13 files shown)


src/compositions/useRoute.ts
pleek91, line 1: I'm tempted to delete the compositions. Annoying to fix their imports and types when they don't even work yet.

src/types/routeMethod.ts
pleek91, line 1: The `RouteMethod` type was moved here and rewritten. Mot of this file actually is new.

src/types/routeMethods.spec-d.ts
```diff
@@ -83,7 +82,7 @@ test('matches individual optional params', () => {
 
   expectTypeOf(router.routes.foo).parameter(0).toEqualTypeOf<{
     foo?: string,
-  }>()
+  } | undefined>()
```
pleek91, line 85: Fixing the RouteMethod type caught that these tests were passing but incorrect. All params are optional so the params argument itself is optional as well.

src/utilities/createRouteMethods.spec.ts
```diff
@@ -17,7 +17,7 @@ test.each([
   ] as const satisfies Routes
   const resolved = resolveRoutes(routes)
 
-  const response = createRouteMethods<typeof routes>(resolved)
+  const response = createRouteMethods<typeof routes>(resolved, vi.fn())
```
pleek91, line 20: `createRouteMethods` now requires `RouterPush` is passed as the second argument. So all these tests have to mock it. Felt gross updating these tests with an empty mock. But when I wrote the new tests at the bottom of this file it was really nice to be able to mock it.

src/utilities/createRouteMethods.ts
```diff
@@ -30,16 +34,56 @@ export function createRouteMethods<T extends Routes>(routes: Resolved<Route>[]):
...
-function createCallableNode(route: Resolved<Route>): RouteMethod {
-  const node: RouteMethod = (values) => {
-    const url = assembleUrl(route, values)
-    const { then } = new Promise<Then<RouteMethodResponse>>(resolve => resolve({ url }))
+type CreateRouteMethodArgs = {
+  route: Resolved<Route>,
+  routerPush: RouterPush,
+}
+
+function createRouteMethod({ route, routerPush }: CreateRouteMethodArgs): RouteMethod {
```
pleek91, line 42: I thought the implementation here turned out nice and clean. Good sign that we're making a good choice.

stackoverfloweth (approved):

PR #52, opened 2024-01-13
Implement RouterLink component

pleek91:
# Description
Most basic implementation. Several things this doesn't account for. But seems to work as expected for basic rendering and clicking.

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -25,7 +25,8 @@
     }
   },
   "rules": {
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
+    "vue/new-line-between-multi-line-property": "off"
   },
   "overrides": [
     {
--- a/src/components/routerLink.browser.spec.ts
+++ b/src/components/routerLink.browser.spec.ts
@@ -1,5 +1,5 @@
-import { flushPromises, mount } from '@vue/test-utils'
-import { expect, test } from 'vitest'
+import { mount } from '@vue/test-utils'
+import { expect, test, vi } from 'vitest'
 import { h } from 'vue'
 import routerLink from '@/components/routerLink'
 import { Route } from '@/types'
@@ -36,24 +36,27 @@ test('renders an anchor tag with the correct href and slot content', () => {
   expect(wrapper.html()).toBe(`<a href="${href}">${content}</a>`)
 })
 
-test('updates route component when clicked', async () => {
-  const routeAContent = 'routeA'
+test.each([
+  true,
+  false,
+])('calls router.push with url and replace %s', (replace) => {
   const routeA = {
     name: 'routeA',
     path: '/routeA',
-    component: { render: () => h(routerLink, { to: router.routes.routeB }, routeAContent) },
+    component: { render: () => h(routerLink, { to: router.routes.routeB, replace }) },
   } as const satisfies Route
 
-  const routeBContent = 'routeB'
   const routeB = {
     name: 'routeB',
     path: '/routeB',
-    component: { template: routeBContent },
+    component,
   } as const satisfies Route
 
   const router = createRouter([routeA, routeB], {
     initialUrl: routeA.path,
   })
+
+  const spy = vi.spyOn(router, 'push')
 
   const root = {
     template: '<RouterView />',
@@ -65,11 +68,7 @@ test('updates route component when clicked', async () => {
     },
   })
 
-  expect(app.text()).toBe(routeAContent)
-
   app.find('a').trigger('click')
 
-  await flushPromises()
-
-  expect(app.text()).toBe(routeBContent)
+  expect(spy).toHaveBeenLastCalledWith(routeB.path, { replace })
 })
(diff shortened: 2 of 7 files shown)


src/components/routerLink.browser.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    name: 'routeB',
+    path: '/routeB',
+    component,
+  } as const satisfies Route
+
+  const router = createRouter([routeA, routeB], {
+    initialUrl: routeA.path,
+  })
+
+  const spy = vi.spyOn(router, 'push')
```
pleek91, line 59: Being able to test like this makes me feel the router design is going in the right direction.

src/components/routerView.ts
pleek91, line 1: unrelated, just updating to match the link.

src/.eslintrc
```diff
@@ -25,7 +25,8 @@
     }
   },
   "rules": {
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
```
stackoverfloweth, line 28: yeah I don't like this rule. kitbag needs it's own eslint-config

src/components/routerLink.browser.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { mount } from '@vue/test-utils'
+import { expect, test, vi } from 'vitest'
+import { h } from 'vue'
+import routerLink from '@/components/routerLink'
+import { Route } from '@/types'
+import { component, createRouter } from '@/utilities'
```
stackoverfloweth, line 6: no need to do anything now, just thinking. I don't think testHelpers was supposed to be included in barrel, I likely added it with my barrel vscode plugin. Might have to update this import if/when it gets removed again

src/components/routerLink.browser.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  const href = path.replace(':param', param)
+
+  const route = {
+    name: 'parent',
+    path,
+    component,
+  } as const satisfies Route
+
+  const router = createRouter([route], {
+    initialUrl: route.path,
```
stackoverfloweth, line 21: if this spec is running in happydom do we still need to clarify an `initialURL`?

stackoverfloweth: I think long term that was the vision, maybe we just didn't implement the "go get it from the browser" part yet

src/components/routerLink.browser.spec.ts
stackoverfloweth: 💯

src/compositions/useRouter.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function useRouter(): Router {
+  const router = inject(routerInjectionKey)
+
+  if (!router) {
+    throw new Error('Router is not installed')
+  }
+
+  // any prevents typescript from trying to verify the types are compatible
+  return router as any
```
stackoverfloweth, line 13: do we actually need this? In my testing locally it works fine without `as any`

src/components/routerLink.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useRouter } from '@/compositions/useRouter'
+import { RouteMethod } from '@/types/routeMethod'
+
+export default defineComponent({
+  name: 'RouterLink',
+  expose: [],
+  props: {
+    to: {
+      type: Function as PropType<() => ReturnType<RouteMethod>>,
+      required: true,
```
stackoverfloweth, line 11: I'm assuming one of the things this doesn't account for yet are the other types we will accept for `to`, like a string?

src/components/routerLink.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+    replace: {
+      type: Boolean,
+      required: false,
```
stackoverfloweth, line 15:
```suggestion
```

nit, this is inferred by vue by default

stackoverfloweth (approved):

src/.eslintrc
pleek91: This ^

src/components/routerLink.browser.spec.ts
pleek91: We should really move test stuff into its own folder

src/components/routerLink.browser.spec.ts
pleek91: Correct. But for a unit test it's also nice to explicitly set it to the route defined rather than just magically matching the window.location

src/compositions/useRouter.ts
pleek91: I was getting infinitely deep errors. Possible the cause of that was fixed. Will check

src/components/routerLink.ts
pleek91: Correct, we'll have to make to a union.

stackoverfloweth (approved):

src/compositions/useRouter.ts
pleek91: Yep no longer needed. Good catch 👍

PR #53, opened 2024-01-13
Add type for flattening routes into an object of route & params pairs

pleek91:
# Description
Introduces a new type called `Flattened` that accepts `Route | Routes`. Produces a flag record keyed by the dot notation name of the route with values of each individual route's params

WIP - Need to add tests

### DIFF
--- a/src/types/flattened.ts
+++ b/src/types/flattened.ts
@@ -1,0 +1,48 @@
+import { ExtractParamsFromPath, MarkOptionalParams, MergeParams } from '@/types/routeMethods'
+import { Public, Route, Routes } from '@/types/routes'
+import { Identity, UnionToIntersection } from '@/types/utilities'
+
+export type Flattened<
+  TRoute extends Route | Routes,
+  TPrefix extends string = '',
+  TParams extends Record<string, unknown> = Record<never, never>
+> = Identity<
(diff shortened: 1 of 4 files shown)


stackoverfloweth (approved):
I'm annoyed with how simple this appears after you did all the work to figure it out 🤣  Nice work

src/types/flattened.ts
```diff
@@ -0,0 +1,48 @@
...
+  TRoute extends Route,
+  TPrefix extends string,
+  TParams extends Record<string, unknown> = Record<never, never>
+> = TRoute extends { path: infer Path, children: infer Children extends Routes }
+  ? TRoute extends Public<TRoute> & { name: infer Name extends string }
+    ? Flattened<Children, Prefix<Name, TPrefix>, MergeParams<TParams, ExtractParamsFromPath<Path>>>
+    : Flattened<Children, Prefix<'', TPrefix>, MergeParams<TParams, ExtractParamsFromPath<Path>>>
+  : Record<never, never>
+
+type Prefix<
```
pleek91, line 39: This accepts a name and a prefix and joins them with dot notation. But it checks if either is an empty string. So don't end up with things like `parent.` or `.child` or `parent..grandchild`

PR #56, opened 2024-01-15
RFC: Add type support for route query

pleek91:
# Description
Adds initial type support for query params on `Route`. This works very similarly to out path params work. `Route.query` is type `string | Query`. 

## Examples
This example has three params. `name`, `gender`, and `age`. `name` and `gender` are required, `age` is optional. All are strings.
```typescript
const route = {
  name: 'route',
  path: '/route',
  query: 'name=:name&gender=:gender&age=:?age'
} as const satisfies Route
```
Same example but using the `query` utility to make the `age` param a `number`
```typescript
const route = {
  name: 'route',
  path: '/route',
  query: query('name=:name&gender=:gender&age=:?age', {
    age: Number
  })
} as const satisfies Route
```
## Route Matching
This is not implemented. Didn't want to get to far without discussing. But the idea is that any route that has a query would require those params exist unless they are optional in order to match the route. If a parent route specifies a `query` it is added to the `query` on any child routes. Unlike the path, the order of params in the query does not matter. Also additional params can exist. 

## Route methods & `Router.push`
I'm not 100% sure how to account for required query params when routing. I briefly thought about combining all params together when asking for values. But that get's weird with multiple params with the same name because the ordering of query params. I think a better experience would be making both `params` and `queryParams` required but keep their types separate. 

So when passing a `string` to `Router.push` there's no issue. 
When using the object route syntax there's no issue (`{ route: 'parent.child', queryParams: { age: 1 } }`)
But it doesn't really fit with the route methods devx.

### DIFF
--- a/src/types/params.ts
+++ b/src/types/params.ts
@@ -27,3 +27,30 @@ export function isParamGetSet(value: Param): value is ParamGetSet {
     && 'set' in value
     && typeof value.set === 'function'
 }
+
+export type ExtractParamName<
+  TParam extends string
+> = TParam extends `?${infer Param}`
+  ? Param extends ''
+    ? never
+    : Param
+  : TParam extends ''
+    ? never
+    : TParam
+
+export type ExtractPathParamType<
+  TParam extends string,
+  TParams extends Record<string, Param | undefined>
+> = TParam extends `?${infer OptionalParam}`
+  ? OptionalParam extends keyof TParams
+    ? ExtractParamType<TParams[OptionalParam]> | undefined
+    : string | undefined
+  : TParam extends keyof TParams
+    ? ExtractParamType<TParams[TParam]>
+    : string
+
+export type ExtractParamType<TParam extends Param | undefined> = TParam extends ParamGetSet<infer Type>
+  ? Type
+  : TParam extends ParamGetter
+    ? ReturnType<TParam>
+    : string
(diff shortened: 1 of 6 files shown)


src/types/params.ts
pleek91, line 1: These types are moved from `routeMethods.ts` Ended up reusing quite a bit of the types from route methods and paths for query params so organized things and exported stuff.

src/types/routeMethods.ts
pleek91, line 1: Same functionality. Just moved some types out of this file into other files.

src/utilities/params.ts
```diff
@@ -1,5 +1,46 @@
...
+    },
+    set: (value) => {
+      if (!stringHasValue(value)) {
+        return ''
+      }
+
+      return setParamValue(value, param)
+    },
+  }
+}
```
pleek91, line 27: These are unchanged just moved from `utilities/path.ts` to here.

src/utilities/params.ts
```diff
@@ -1,5 +1,46 @@
...
+
+export function getParamsForString<TInput extends string, TParams extends Record<string, Param | undefined>>(string: TInput, params: TParams): Record<string, unknown[]> {
```
pleek91, line 29: This is an abstraction of the guts of the `path` utility because this piece is exactly the same for the `query` utility.

src/utilities/path.ts
pleek91, line 1: Just moving things around

stackoverfloweth (approved):

PR #57, opened 2024-01-17
Query updates to route resolver

stackoverfloweth:
- added `asArray` utility
- updated `mergeParams` utility to expect possibility of `Record<string, Param | Param[]>`, previously assumed params would be array, now normalizes to array
- added concatenating "query" on routes inside of route resolver
- including query params along side path params concatenated "params" inside of route resolver


src/utilities/resolveRoutes.ts
```diff
@@ -32,11 +34,16 @@ export function resolveRoutes(routes: Routes, parentContext: ParentContext = {})
...
+        query: fullQuery.map(({ query }) => query.toString()).join('&'),
+        params: mergeParams(reduceParams(fullPath), reduceParams(fullQuery)),
       })
     }
 
     return value
   }, [])
+}
+
+function reduceParams(entries: { params: Record<PropertyKey, unknown> }[]): Record<PropertyKey, Param[]> {
```
pleek91, line 47: Does this not work?
```suggestion
function reduceParams(entries: Path[] | Query[]): Record<string, Param[]> {
```

Curious about why the generic record type for the params and also why `PropertyKey` rather than `string` for the return type.

src/utilities/resolveRoutes.spec.ts
```diff
@@ -90,6 +90,53 @@ describe('resolveRoutes', () => {
...
+      query: 'start=?:startDate',
+      children: [childRoute],
+    }
+
+    const routes = resolveRoutes([parentRoute])
+    const resolvedChild = routes.find(route => route.name === childRoute.name)
+
+    expect(resolvedChild?.params).toMatchObject({ workspaceId: [String], startDate: [String], handle: [String] })
+    // check
+    expect(Object.keys(resolvedChild?.params ?? {})).toMatchObject(['workspaceId', 'startDate', 'handle'])
```
pleek91, line 137: Won't the first expect verify the keys are correct?

pleek91 (approved):
Looking good! Couple questions but otherwise ship it

src/utilities/mergeParams.ts
```diff
@@ -1,12 +1,15 @@
...
   return paramRecords.reduce<Record<string, Param[]>>((combined, collection) => {
     Object.entries(collection).forEach(([key, params]) => {
+      const paramsArray = asArray(params)
+
       if (key in combined) {
-        params.forEach(param => combined[key].push(param))
+        paramsArray.forEach(param => combined[key].push(param))
       } else {
-        combined[key] = [...params]
+        combined[key] = [...paramsArray]
```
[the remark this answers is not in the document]
stackoverfloweth: I like most of this, TS doesn't like `combined[key] ??= []`, it's that annoying error we get when doing a reduce

> Unnecessary conditional, expected left-hand side of `??` operator to be possibly null or undefined

because we're saying the "combined" is `Record<string, Param[]>`, so it say's it won't be undefined. I think I'm going to do the following

```ts
if (!(key in combined)) {
  combined[key] = []
}

combined[key].push(...asArray(params))
```

src/utilities/resolveRoutes.ts
stackoverfloweth: yep, both seem fine 👍

src/utilities/resolveRoutes.spec.ts
stackoverfloweth: yes, but not in the correct order. that comment was supposed to say "check order"

pleek91: Why do we care about the order of the keys in the object?

stackoverfloweth: Because the order of params here is the order they will be evaluated when trying to match a url. Parent path params, then child path params, then parent query params, then child query params

pleek91: That only matters within same named params though right?

pleek91: The object keys are not going to be in that order. They're alphabetical I believe.

stackoverfloweth: Ah you're right. I was confusing the two. I'll remove the line that checks order.

PR #58, opened 2024-01-18
Update route method and flattened types to support query params

pleek91:
# Description
Updates both the `Flattened` and the `RouteMethods` types (and their individual pieces) to support query params.


src/types/flattened.ts
pleek91, line 1: The changes here are mostly renaming `TParams` to `TPathParams` and then adding an additional generic for `TQueryParams`. The only other change is the introduction of the `RoutePathParams` and `RouteQueryParams` types which simplify combining a parent params with the current route's params. Removes the need to infer and combine things in multiple places and can rather just call that those types to get the params for a route.

src/types/routeMethods.ts
pleek91, line 1: The changes here are almost identical to the changes in `flattened`. Added query params. And simplified some places where `Path` was being inferred by using the `RoutePathParams` and `RouteQueryParams` types.

PR #59, opened 2024-01-20
More complete types for `RouterPush` and abstact implementation into `createRouterPush`

pleek91:
# Description
Updates the types for `RouterPush` to be a little more accurate. Was hoping to remove the need for an `as any` but still had that same issue in `createRouterPush`. But at least now its consistent with `createRouteMethods`.

### DIFF
--- a/src/types/routeMethods.ts
+++ b/src/types/routeMethods.ts
@@ -48,15 +48,15 @@ export type ExtractRouteMethodParams<T extends RouteMethod> =
 export type RoutePathParams<
   TRoute extends Route,
   TPathParams extends Record<string, unknown>
-> = TRoute extends { path: infer Path }
-  ? MergeParams<TPathParams, ExtractParamsFromPath<Path>>
+> = TRoute extends { path: infer TPath extends string | Path }
+  ? MergeParams<TPathParams, ExtractParamsFromPath<TPath>>
   : MergeParams<TPathParams, {}>
 
 export type RouteQueryParams<
   TRoute extends Route,
   TQueryParams extends Record<string, unknown>
-> = TRoute extends { query: infer Query }
-  ? MergeParams<TQueryParams, ExtractParamsFromQuery<Query>>
+> = TRoute extends { query: infer TQuery extends string | Query }
+  ? MergeParams<TQueryParams, ExtractParamsFromQuery<TQuery>>
   : MergeParams<TQueryParams, {}>
 
 type ExtractParamsFromPath<
--- a/src/types/router.ts
+++ b/src/types/router.ts
@@ -3,6 +3,7 @@ import { Flattened } from '@/types/flattened'
 import { Resolved } from '@/types/resolved'
 import { RouteMethods } from '@/types/routeMethods'
 import { Route, Routes } from '@/types/routes'
+import { AllPropertiesAreOptional, Identity } from '@/types/utilities'
 
 export type RouterOptions = {
   initialUrl?: string,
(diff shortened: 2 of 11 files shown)


src/types/routeMethods.ts
pleek91, line 1: Unrelated but when I was testing some things to help with the recursive types I added this and felt it was worth keeping.

src/types/router.ts
```diff
@@ -14,19 +15,26 @@ export type RouterPushOptions = {
...
 }
 
-type RouterPushRoute<
-  TRoutes extends Routes,
-  TRoute extends PropertyKey
-> = RouterPushOptions & {
-  name: TRoute,
-  params?: TRoute extends keyof Flattened<TRoutes> ? Flattened<TRoutes>[TRoute] : Record<never, never>,
-}
+type RoutePushConfigParams<
```
pleek91, line 18: This type just makes sure the `params` property is not required if there are no required params.

src/types/routes.ts
pleek91, line 1: Unrelated but noticed this didn't get updated when I added query to the route types.

src/utilities/createRouteMethods.spec.ts
pleek91, line 1: Updated `createRouteMethods` to accept an object rather than multiple arguments after I wrote `createRouterPush` that way and thought it felt good. So this is the same functionally.

src/utilities/createRouter.ts
```diff
@@ -43,38 +44,16 @@ export function createRouter<T extends Routes>(routes: T, options: RouterOptions
     Object.assign(route, newRoute)
   }
 
-  function pushUrl(url: string, options: RouterPushOptions = {}): Promise<void> {
-    return navigation.update(url, options)
+  function replace(url: string, options: RouterReplaceOptions = {}): Promise<void> {
+    return push(url, { ...options, replace: true })
```
pleek91, line 48: Updated `replace` to just call `router.push`. That way theres only one implementation of what "replace" means.

src/utilities/createRouterPush.ts
```diff
@@ -0,0 +1,39 @@
+import { Resolved, Route, RouterPush, RouterPushOptions, Routes } from '@/types'
+import { flattenParentMatches } from '@/utilities/flattenParentMatches'
+import { RouterNavigation } from '@/utilities/routerNavigation'
+import { component } from '@/utilities/testHelpers'
+import { assembleUrl } from '@/utilities/urlAssembly'
+
+type DummyRoutes = [{ name: string, path: '/', component: typeof component }]
```
pleek91, line 7: Curious what you think of this. I'm using this below in the push implementation on line 16 so that typescript knows the correct types for the arguments and return types. I cannot use just `RouterPush<[]>` because an empty array produces no type for the overload and then it says `string` is the only type for `urlOrRouteConfig`.

stackoverfloweth: we can't use `Route`? Is this intended to be a temporary fix?

pleek91: Cannot use Route or Routes because they both have generics that cause the infinite depth errors. But something fixed that is more narrow works just fine. 

Not intended to be temporary. Its gross. And open to other suggestions. But this is the most type safe I could make this.

stackoverfloweth: that makes sense. The things I really don't like about it is the name and the fact that it uses component from our utilities/testHelpers. 

What if we changed the type to

```ts
type AnyRoute = [{ name: string, path: string, component: RouteComponent }]
```

pleek91: Yep that totally work. Good suggestion

stackoverfloweth (approved):

PR #60, opened 2024-01-20
add depth: number property to resolved routes

stackoverfloweth:
to be used as part of "scoring" possible route matches


pleek91 (approved):
Nice, super simple

PR #61, opened 2024-01-20
refactored route matching to include new logic for matching query

stackoverfloweth:
The objective here is to make sure that routes that have query defined properly narrow the results on route match logic.

The rule is a bit different from path matching, which expects a full Regex match including ^ (start) and $ (end). Query can include extra unexpected bits, and the order does not matter. 

Some refactoring happened in this PR to accomplish the goals above. 
- I considered changing `ResolveWithRegex` to go from `regex: Regexp` to `regex: {param: Regexp, query: Regexp[]}` but decided instead to ditch the concept of `ResolveWithRegex` type altogether. This also meant doing away with idea of sticking that regexp on the route once and using that everytime we match. Now it generates that Regex every time, which I suspect has a imperceivable impact on performance.
- The route match logic used to "does that route regex pass?" and "does param validation pass?", I needed a 3rd condition, so I decided to introduce a type `RouteMatchRule` so all of these rules could share the same syntax `(route: Resolved<Route>, url: string) => boolean` and matching a route just means calling each "rule" and making sure they all pass.
- I introduced `urlSplitter`, which takes the "url" and returns it's constituent parts (using path and query, for now). This meant updating some unrelated tests which would establish routes/initial URL without a leading slash. A url value without a slash is interpreted as having "domain" only.


src/utilities/paramValidation.spec.ts
```diff
@@ -11,7 +11,7 @@ test('given route WITHOUT params, always return true', () => {
   }
   const [resolved] = resolveRoutes([route])
 
-  const response = routeParamsAreValid('/no-params', resolved)
+  const response = routeParamsAreValid(resolved, '/no-params')
```
stackoverfloweth, line 14: just swapping order of params to match `RouteMatchRule`

src/utilities/routeRegex.ts
```diff
@@ -1,7 +1,22 @@
-export function generateRouteRegexPattern(value: string): RegExp {
+import { Resolved, Route } from '@/types'
+
+export function generateRoutePathRegexPattern(route: Resolved<Route>): RegExp {
```
stackoverfloweth, line 3: the path part is just renamed, this file is now also responsible for generating the `RegExp[]` for a routes query

src/utilities/urlSplitter.ts
```diff
@@ -0,0 +1,30 @@
...
+  protocol: string | undefined,
+  domain: string | undefined,
+  port: string | undefined,
+  path: string | undefined,
+  query: string | undefined,
+  hash: string | undefined,
+}
+
+export function splitUrl(url: string): SplitURL {
+  const urlPattern = /^(?:(?<protocol>\w+):\/\/)?(?:([^/@:]+)(?::([^/@:]+))?@)?(?<domain>[^/?#:]+)?(?:(?<port>:\d+))?(?<path>\/[^?#]*)?(?<query>\?[^#]*)?(?<hash>#.*)?$/
```
stackoverfloweth, line 11: this was written by GPT, could probably be cleaned up

src/utilities/urlSplitter.ts
```diff
@@ -0,0 +1,30 @@
...
+    protocol: match?.groups?.protocol,
+    domain: match?.groups?.domain,
+    port: clean(match?.groups?.port, ':'),
+    path: match?.groups?.path,
+    query: clean(match?.groups?.query, '?'),
+    hash: clean(match?.groups?.hash, '#'),
+  }
+}
+
+function clean(value: string | undefined, toRemove: string): string | undefined {
```
stackoverfloweth, line 24: I'm on the fence about this. Would you expect query to start with "?", has to start with "#", and port to start with ":"?

src/utilities/urlSplitter.ts
pleek91: Couldn't we just use the URL class for this?

src/utilities/urlSplitter.ts
pleek91: I don't think it matters much as long as we're consistent

src/utilities/routeMatchRegexRules.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    const response = routeQueryMatches(resolved, url)
+
+    expect(response).toBe(false)
+  })
+
+  test.each([
+    ['?without=params&static=true'],
+    ['www.kitbag.io?without=params&static=true'],
+    ['www.kitbag.io/some/path?without=params&static=true'],
```
pleek91, line 88: I wouldn't expect this case to return true. Because the route has a path of `''` but and this path wouldn't match right?

pleek91 (commented):
This looks great. Couple questions about test cases and url splitting. But overall really nice PR 👍

src/utilities/urlSplitter.ts
stackoverfloweth: It's very possible, but my first attempt resulted in constant errors being thrown for "invalid urls". I'll do a little research and report back

pleek91: Most likely do to a lack of complete urls. That class needs a fully qualified url. But you can also provide it a base to use.

src/utilities/routeMatchRegexRules.spec.ts
stackoverfloweth: you're right that if we were testing `routeMatch` this case would fail, but in this case we're only testing `routeQueryMatches`

src/utilities/createMaybeRelativeUrl.ts
pleek91, line 1: Hmm so not having a base in node is why all the tests were changed to run in a browser environment?

stackoverfloweth: correct. This service needs window.location if the path is relative. We tend to write tests with relative urls and we also tend not to mock dependencies so a handful of test files now need `window.location`

pleek91: Hmm yeah makes sense but I don't love not having any node env tests. Maybe we can discuss testing patterns and find a better solution. For now though this makes sense to me 👍

pleek91 (approved):

PR #62, opened 2024-01-20
using route scoring on all route matches, returning the highest score

stackoverfloweth:
add concept of route score: more expected query params fulfilled the higher the value. If equal then the route with higher depth will have the higher value

updated route match logic to get score for each potential match, return first (sort desc, so first with highest score)


src/utilities/routeMatchScore.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const [aRoute, bRoute] = resolveRoutes(routes)
+
+    const sortByRouteScore = getRouteScoreSortMethod('/?color=red&id=1&extra=ok')
+    const response = [aRoute, bRoute].sort(sortByRouteScore)
+
+    expect(response).toMatchObject([bRoute, aRoute])
+  })
+
+  test('given routes with different query scores and asc sort, returns them sorted by query score ascending', () => {
```
pleek91, line 91: Curious your thoughts on the sorts. What do we need asc for?

pleek91 (commented):
Couple questions but overall this makes sense to me, nicely done 👍

src/utilities/routeMatchScore.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+    } as const satisfies Route
+    const [resolved] = resolveRoutes([route])
+    const actualQuery = new URLSearchParams('/?can=have&some=queries')
+
+    const response = countExpectedQueryKeys(resolved, actualQuery)
+
+    expect(response).toBe(0)
+  })
+
+  test('given route with query where url has keys but different values, still counts as present', () => {
```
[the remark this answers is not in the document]
stackoverfloweth: you're correct. And I recognized that such behavior was not obvious why we're only matching keys and note values, so to protect ourselves from future us coming back in here and adding unnecessary complexity I wrote a test.

src/utilities/routeMatchScore.spec.ts
stackoverfloweth: over-engineering for no good reason

pleek91: So should we remove the sorts or do you think they'll have value in the future

stackoverfloweth: removed 👍

pleek91 (approved):

PR #63, opened 2024-01-22
Type optimizations

pleek91:
# Description
Type updates to improve performance of the RouteMethods type by allowing nested types to be deferred.


src/types/flattened.ts
pleek91, line 1: Mostly unrelated but was looking at improving this type as well and decided to keep some of the changes I had made. Removing the Identity type. Breaking `Flattened` into two types (to be consistent with the `RouteMethods` type. And specifying `unknown[]` where we know that it the expected type.

src/types/routeMethods.spec-d.ts
```diff
@@ -376,7 +376,7 @@ test('all routes with a name can be called unless disabled', () => {
   expectTypeOf(router.routes.parent.child).toBeFunction()
   expectTypeOf(router.routes.parent2).not.toBeFunction()
   expectTypeOf(router.routes.parent2.child2).toBeFunction()
-  expectTypeOf(router.routes.parent2).not.toHaveProperty('child3')
+  expectTypeOf(router.routes.parent2.child3).toMatchTypeOf<never>()
```
pleek91, line 379: Removing the `Identity` type from `RouteMethods` gives significant performance improvements. The downside is now non public routes show up as `never`. I think we can probably fix this but for now not a high priority. Functionally and devx wise still pretty good as is.

src/types/routeMethods.spec-d.ts
```diff
@@ -407,25 +407,6 @@ test('public parent routes have correct type for parameters', () => {
   }>()
 })
 
-test('param names must be alphanumeric', () => {
```
pleek91, line 410: Another thing I noticed is the `UnifyParamEnds` type reduced the number of parameters that could be parsed. I don't think the use case here is strong enough to accept the performance cost and limiting the number of params. So changing my mind on this "feature" and voting to remove and make `/` the only end character.

src/types/routeMethods.ts
```diff
@@ -6,14 +6,14 @@ import { Query } from '@/utilities/query'
 
 export type RouteMethods<
   TRoutes extends Routes = Routes,
-  TPathParams extends Record<string, unknown> = Record<never, never>,
-  TQueryParams extends Record<string, unknown> = Record<never, never>
-> = Identity<UnionToIntersection<RouteMethodsTuple<TRoutes, TPathParams, TQueryParams>[number]>>
+  TPathParams extends Record<string, unknown[]> = Record<never, never>,
+  TQueryParams extends Record<string, unknown[]> = Record<never, never>
+> = UnionToIntersection<RouteMethodsTuple<TRoutes, TPathParams, TQueryParams>[number]>
```
pleek91, line 11: Removing this `Identity` means
- Non public routes show up in the object as `never`. 
- Typescript can defer evaluation of nested types until they are access. Which gives us a huge performance boost.

src/types/routeMethods.ts
```diff
@@ -22,21 +22,21 @@ type RouteMethodsTuple<
...
+  TPathParams extends Record<string, unknown[]>,
+  TQueryParams extends Record<string, unknown[]>
+> = TRoute extends { children: infer Children extends Routes }
+  ? TRoute extends Public<TRoute>
+    ? RouteMethods<Children, RoutePathParams<TRoute, TPathParams>, RouteQueryParams<TRoute, TQueryParams>> & CreateRouteMethod<TRoute, TPathParams, TQueryParams>
+    : RouteMethods<Children, RoutePathParams<TRoute, TPathParams>, RouteQueryParams<TRoute, TQueryParams>>
   : TRoute extends Public<TRoute>
-    ? CreateRouteMethod<MergeParams<RoutePathParams<TRoute, TPathParams>, RouteQueryParams<TRoute, TQueryParams>>>
+    ? CreateRouteMethod<TRoute, TPathParams, TQueryParams>
     : never
```
pleek91, line 34: Simplified this type a bit by removing a ternary and also letting `CreateRouteMethod` do the merging of params.

src/utilities/createRouteMethods.spec.ts
```diff
@@ -22,10 +22,8 @@ test.each([
   const response = createRouteMethods<typeof routes>({ resolved, push })
 
   if (isPublic !== false) {
-    // @ts-expect-error
```
pleek91, line 25: These no longer error because the never types. Which is kinda nice.

src/types/routeMethods.spec-d.ts
stackoverfloweth: Is this the same type that we use to parse params in the query though? If so we will need "&" as a possible end character as well.

src/types/flattened.ts
```diff
@@ -1,42 +1,43 @@
...
 
 type RouteChildrenFlat<
   TRoute extends Route,
   TPrefix extends string,
-  TPathParams extends Record<string, unknown> = Record<never, never>,
-  TQueryParams extends Record<string, unknown> = Record<never, never>
+  TPathParams extends Record<string, unknown[]> = Record<never, never>,
+  TQueryParams extends Record<string, unknown[]> = Record<never, never>
 > = TRoute extends { children: infer Children extends Routes }
   ? TRoute extends Public<TRoute> & { name: infer Name extends string }
```
stackoverfloweth, line 38: I see that this hasn't changed but I'm curious, why does the route need to be named AND public? I would expect the prefix to be updated only if named. I didn't expected the route being public to matter here?

stackoverfloweth (approved):
This is great, it seems to simplify the types at the same time it improves performance 👏 

Speaking of simplicity, I'd like to test how well I understand the flow through these types

-> Flattened: only purpose is to call RouteFlat and RouteChildrenFlat
  -> RouteFlat: establishes Record type for current level
  -> RouteChildrenFlat: if route is public & named update prefix, either way recurse any children
    -> FlattenedRoutes: maps routes, calls back up to Flattened

it seems `FlattenedRoutes` is the only exported type that is actually used? Should `Flattened` still be exported?

src/types/flattened.ts
pleek91: Um good question. I guess if it has a name that should be part of the dot notation even if its not public yeah? That's probably a bug.

pleek91:
> This is great, it seems to simplify the types at the same time it improves performance 👏
> 
> Speaking of simplicity, I'd like to test how well I understand the flow through these types
> 
> -> Flattened: only purpose is to call RouteFlat and RouteChildrenFlat -> RouteFlat: establishes Record type for current level -> RouteChildrenFlat: if route is public & named update prefix, either way recurse any children -> FlattenedRoutes: maps routes, calls back up to Flattened
> 
> it seems `FlattenedRoutes` is the only exported type that is actually used? Should `Flattened` still be exported?

Yep that tracks. We probably don't need to export `Flattened`. But we also probably need to pivot on our syntax if I cannot figure out a way to defer some of the type definition for this.

PR #65, opened 2024-01-23
Url assembly with query

stackoverfloweth:
this ones pretty straightforward, finally


src/utilities/urlAssembly.spec.ts
```diff
@@ -1,83 +1,170 @@
...
-  ['/simple/:simple'],
-  [path('/simple/:simple', { simple: String })],
-])('given route with required string param provided, returns route Path with string with values interpolated', (path) => {
-  const route = {
-    name: 'simple',
-    path,
-    component,
-  } satisfies Route
-  const [resolved] = resolveRoutes([route])
+describe('query params', () => {
```
stackoverfloweth, line 87: here are the new tests

src/utilities/urlAssembly.ts
```diff
@@ -1,10 +1,11 @@
 import { Param, Resolved, Route } from '@/types'
 import { setParamValuesOnUrl } from '@/utilities/paramsFinder'
 
-export function assembleUrl(route: Resolved<Route>, values: Record<string, unknown> = {}): string {
+export function assembleUrl(route: Resolved<Route>, values: Record<string, unknown[]> = {}): string {
   const params = Object.entries<Param[]>(route.params)
+  const pathWithQuery = route.query.length ? `${route.path}?${route.query}` : route.path
```
stackoverfloweth, line 6: I considered moving this out into it's own utility

src/utilities/urlAssembly.spec.ts
```diff
@@ -1,83 +1,170 @@
...
+      name: 'simple',
+      path: '/',
+      query,
+      component,
+    } satisfies Route
+    const [resolved] = resolveRoutes([route])
+
+    const url = assembleUrl(resolved, {})
+
+    expect(url).toBe('/?simple=')
```
pleek91, line 119: I'm not sure what the expected behavior here would be. This test makes sense for sure. But I wonder if users will expect the key to not be present when the value is absent. But I honestly like how simple this is.

src/utilities/urlAssembly.ts
pleek91: This whole utility is pretty simple. So this ternary doesn't bother me.

src/utilities/urlAssembly.ts
```diff
@@ -1,10 +1,11 @@
...
 
   return params.reduce((url, [name, params]) => {
     return setParamValuesOnUrl(url, { name, params, values: values[name] })
```
pleek91, line 9: Super cool you didn't actually have to write any additional logic for filling in the params

pleek91 (approved):

src/utilities/urlAssembly.ts
stackoverfloweth: yeah, kudos to you for finding the syntax that just works and feels good

src/utilities/urlAssembly.spec.ts
stackoverfloweth: agreed. And unless you felt strongly I was going to just go with whatever is the most cut and dry and let our users tell us what improvements they want to see here

PR #67, opened 2024-01-26
Add new types and implementation for router.push

pleek91:
Adds new types and implementation for router.push that accepts
- A route with params object `{ route: 'parent.child', params: { paramA: true } }`
- A route method `routes.parent.child({ paramA: true })`
- A string `http://example.com` or `/relative/path`


src/types/routePaths.ts
```diff
@@ -0,0 +1,55 @@
...
+> = FlatArray<{
+  [K in keyof TRoutes]: TRoutes[K] extends Route
+    ? [RoutePath<TRoutes[K], TPrefix>, ChildrenPaths<TRoutes[K], TPrefix>]
+    : never
+}, 1>
+
+type RoutePath<
+  TRoute extends Route,
+  TPrefix extends string
+> = TRoute extends Public<TRoute> & { name: infer Name extends string }
```
stackoverfloweth, line 33: public just means it's callable yeah? I think we just need a name for a parent to show up in RoutePath

src/utilities/createRouterPush.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { expect, test, vi } from 'vitest'
+import { createRouterPush } from '@/utilities/createRouterPush'
+import { resolveRoutes } from '@/utilities/resolveRoutes'
+import { createRouterNavigation } from '@/utilities/routerNavigation'
+import { routes } from '@/utilities/testHelpers'
+
+test('push calls onLocationUpdated', () => {
```
stackoverfloweth, line 7: I feel like this aligns with our general strategy so far with unit testing, just wanted to checkin with you to see if we think there are additional good tests that we could do here?

src/types/routePaths.ts
```diff
@@ -0,0 +1,55 @@
...
+  TRouteMethods,
+  TRoutePath extends string
+> =
+  TRoutePath extends `${infer F}.${infer R}`
+    ? F extends keyof TRouteMethods
+      ? NestedObjectValue<TRouteMethods[F], R>
+      : never
+    : TRoutePath extends keyof TRouteMethods
+      ? ExtractRouteMethodParams<TRouteMethods[TRoutePath]>
+      : never
```
stackoverfloweth, line 19: bit of a nit but we can avoid one of these nested ternaries by checking extends after the inference of F

```suggestion
  TRoutePath extends `${infer F extends string & keyof TRouteMethods}.${infer R}`
    ? NestedObjectValue<TRouteMethods[F], R>
    : TRoutePath extends keyof TRouteMethods
      ? ExtractRouteMethodParams<TRouteMethods[TRoutePath]>
      : never
```

src/utilities/createRouterPush.ts
```diff
@@ -1,15 +1,67 @@
...
+    }
+
+    if (isRecord(source)) {
+      const match = resolved.find((resolvedRoute) => flattenParentMatches(resolvedRoute) === source.route)
+
+      if (!match) {
+        throw `No route found: "${String(source)}"`
+      }
+
+      const params = get<Record<string, unknown>>(source, 'params')
```
stackoverfloweth, line 52: I've 100% written a type like this when I got annoyed by Typescript but ultimately I think we're just hiding the `as Record<string, unknown>` in a method. I personally would drop the excess in favor of just casting here

```suggestion
      const params = source.params as Record<string, unknown>
```

src/utilities/flattenParentMatches.ts
```diff
@@ -0,0 +1,9 @@
+import { Resolved, Route, isNamedRoute, isPublicRoute } from '@/types'
+
+export function flattenParentMatches(routes: Resolved<Route>): string {
+  const namedAndPublicRoutesNames = routes.matches
+    .filter(route => isNamedRoute(route) && isPublicRoute(route))
```
stackoverfloweth, line 5: same question applies here, not sure we need name and public, possible we're not applying that business rule correctly throughout

src/utilities/createRouterPush.ts
stackoverfloweth: never mind, I see now how this avoid TS checking at all. Unfortunate but makes sense now

src/utilities/guards.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  return true
+}
+
+const test = { bar: true }
+
+if (hasProperty(test, 'foo', Boolean)) {
+  test
+}
```
stackoverfloweth, line 27: guessing this is by mistake

```suggestion
```

src/utilities/guards.ts
```diff
@@ -0,0 +1,27 @@
+export function isRecord(value: unknown): value is Record<PropertyKey, unknown> {
+  return typeof value === 'object' && value !== null
```
stackoverfloweth, line 2: this type would return `true` given `[]`, it would still fail when you check for `key in ...` but could be misused in the future?

stackoverfloweth (commented):
few small things

src/utilities/guards.ts
pleek91: Good catch. Updated

src/types/routePaths.ts
pleek91: Yeah this is incorrect. Removed the `Public<TRoute>` check.

src/utilities/flattenParentMatches.ts
pleek91: Yeah why'd you do that lol. Removed

src/utilities/createRouterPush.spec.ts
pleek91: I thought about additional tests. Even wrote some, but ended up deleting them as they didn't seem very useful and mostly fell under the "test js does what js is supposed to do". I agree though it feels like there should be more tests around this. Any suggestions?

stackoverfloweth: nope, just struck me as odd to have a single test for a function that is not tiny.

stackoverfloweth (approved):

PR #69, opened 2024-01-28
Abstract resolving a url from a route source into router.resolve

pleek91:
# Description
Moves the logic for determining the url from a route source out of createRouterPush and into createRouterResolve. Also adds `resolve` to the `Router` type so it can be used in other applications like in `RouterLink`


src/utilities/createRouterResolve.browser.spec.ts
pleek91, line 1: I wanted to do these tests by using `createRouterResolve` directly but ran into type issues so aborted that. Going to investigate some type improvements and hopefully come back to this.

stackoverfloweth (approved):
This makes a lot of sense. Nice change

PR #70, opened 2024-01-28
Separate out type checking from the type tests

pleek91:
# Description
Better devx so test:types can run without picking up source errors.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,6 +20,6 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true
   },
-  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.vue"],
+  "include": ["src/**/*.ts"],
   "exclude": ["node_modules", "dist"]
 }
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -20,6 +20,6 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true
   },
-  "include": ["src/**/*.ts", "src/**/*.d.ts", "src/**/*.vue"],
```
stackoverfloweth, line 23: This change surprised me. I thought we just wanted to remove type checking from spec files?

pleek91: Well we don't have any .vue files and *.ts includes *.d.ts. So idk that anything here changed. I was just cleaning up cause I came to look at the tsconfig.

pleek91: I didn't actually remove the .spec files from the tsconfig (in order to do that we'd have to convert them to .js). Just removed source checking from the type testing command

stackoverfloweth (approved):

PR #71, opened 2024-01-28
RFC: Implementation types for internal utilities

pleek91:
# Description
Adding more general implementation types for internal utilities so that we don't have to pass so many generics around. I think this will make development and testing simpler. 

I'm not attached to this idea. But it felt kinda nice when making these changes. 

The idea is that internal utilities like `createRouteMethods`, and `createRouterPush` return wider `Implementation` types that don't use generics. I think this means there will be less typescript depth but not totally sure. Either way several methods got simpler.

### DIFF
--- a/src/types/router.ts
+++ b/src/types/router.ts
@@ -1,8 +1,8 @@
 import { App, DeepReadonly } from 'vue'
 import { Resolved } from '@/types/resolved'
-import { RouteMethods } from '@/types/routeMethods'
+import { RouteMethods, RouteMethodsImplementation } from '@/types/routeMethods'
 import { Route, Routes } from '@/types/routes'
-import { RouterPush, RouterPushOptions } from '@/utilities/createRouterPush'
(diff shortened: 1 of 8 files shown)


src/types/router.ts
```diff
@@ -26,4 +26,16 @@ export type Router<
...
+export type RouterImplementation = {
+  routes: RouteMethodsImplementation,
+  route: DeepReadonly<Resolved<Route>>,
+  resolve: RouterResolve,
+  push: RouterPushImplementation,
+  replace: RouterReplace,
+  back: () => void,
+  forward: () => void,
+  go: (delta: number) => void,
+  install: (app: App) => void,
```
stackoverfloweth, line 40: I like the syntax of having all the "...Implementation" properties coming together here. Thoughts on following that  even further, maybe created Implementation types for resolve, replace? Just thinking out loud, def not a blocker

pleek91: We could, but they'd be the exact same types. Those types don't have any generics as is.

stackoverfloweth (approved):

PR #72, opened 2024-01-30
Rejections

pleek91:
# Description
Adds types and support for rejections and implements the NotFound rejection type.

### DIFF
--- a/src/utilities/createRouterReject.ts
+++ b/src/utilities/createRouterReject.ts
@@ -1,0 +1,19 @@
+import { NotFound } from '@/components'
+import { RegisteredRejection, RouteComponent } from '@/types'
+
+export const builtInRejections = ['NotFound'] as const
+export type BuiltInRejection = typeof builtInRejections[number]
+
+export const builtInRejectionComponents: Record<BuiltInRejection, RouteComponent> = {
+  NotFound,
+}
+
+export type RouterRejection = BuiltInRejection | RegisteredRejection
+
+export type RouterReject = (type: RouterRejection) => void
+
+export function createRouterReject(): RouterReject {
+  return (type) => {
+
+  }
+}
--- a/src/utilities/createRouterReject.ts
+++ b/src/utilities/createRouterReject.ts
@@ -21,6 +21,7 @@ export type RouterReject = (type: RouterRejection) => void
 
 type GetRejectionComponent = (type: RouterRejection) => RouteComponent
 type GetRejectionRoute = (type: RouterRejection) => Resolved<Route>
+type ClearRejection = () => void
 export type RouterRejectionComponent = Ref<null | { type: RouterRejection, component: RouteComponent }>
 
 type CreateRouterRejectContext = {
@@ -30,6 +31,7 @@ type CreateRouterRejectContext = {
 type CreateRouterReject = {
   reject: RouterReject,
   rejection: RouterRejectionComponent,
+  clearRejection: ClearRejection,
   getRejectionRoute: GetRejectionRoute,
   getRejectionComponent: GetRejectionComponent,
 }
@@ -67,6 +69,10 @@ export function createRouterReject({
     return resolved
   }
 
+  const clearRejection: ClearRejection = () => {
+    rejection.value = null
+  }
+
   const reject: RouterReject = (type) => {
     const component = getRejectionComponent(type)
 
@@ -78,6 +84,7 @@ export function createRouterReject({
   return {
     reject,
     rejection,
+    clearRejection,
     getRejectionRoute,
     getRejectionComponent,
   }
--- a/src/utilities/createRouterReject.ts
+++ b/src/utilities/createRouterReject.ts
@@ -1,28 +1,27 @@
 import { Ref, ref } from 'vue'
 import { NotFound } from '@/components'
-import { RegisteredRejection, Resolved, Route, RouteComponent } from '@/types'
+import { RegisteredRejectionType, Resolved, Route, RouteComponent } from '@/types'
 
 export const builtInRejections = ['NotFound'] as const
-export type BuiltInRejection = typeof builtInRejections[number]
+export type BuiltInRejectionType = typeof builtInRejections[number]
 
-export const builtInRejectionComponents: Record<BuiltInRejection, RouteComponent> = {
+export const builtInRejectionComponents: Record<BuiltInRejectionType, RouteComponent> = {
   NotFound,
 }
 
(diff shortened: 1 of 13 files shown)


src/utilities/createRouterReject.ts
```diff
@@ -0,0 +1,91 @@
...
+export const builtInRejectionComponents: Record<BuiltInRejection, RouteComponent> = {
+  NotFound,
+}
+
+export type RouterRejection = BuiltInRejection | RegisteredRejection
+
+type BuiltInRejectionComponents = Partial<Record<BuiltInRejection, RouteComponent>>
+
+export type RouterRejectionComponents = RegisteredRejection extends never
+  ? { rejections?: BuiltInRejectionComponents }
```
stackoverfloweth, line 17: just to clarify here, if the developer doesn't have any additional RegisteredRejections, the rejections property itself is optional. Presumably because the only rejection (NotFound) is handled but could be overridden. If the user does provide their own rejections, the rejections property is required. Any user defined rejections are required, the NotFound is still optional. 

this is beautiful

src/utilities/createRouterReject.ts
```diff
@@ -0,0 +1,91 @@
...
+      matches: [route],
+      name: type,
+      depth: 0,
+      path: '',
+      query: '',
+      params: {},
+    }
+
+    return resolved
+  }
```
stackoverfloweth, line 70: it seems the only purpose of this function is make sure the router has a valid "current route"? I have 2 comments

1. Since this route is generated outside of user-land, is it worth tagging it with some symbol that we can theoretically use elsewhere to tell user that current route is a rejection? 
1.A. Alternatively we could say router.route is either a route or a rejection? That would mean we can avoid the slightly awkward depth, path, query, params that are all not really applicable but undoubtably would have other implications?
2. Should we just have the `reject` method return this route instead of having a separate method?

That would mean we can update the syntax in router to be

```ts
if (!route) {
   return reject('NotFound')
}
```

src/utilities/createRouterReject.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const rejection: RouterRejectionComponent = ref(null)
+
+  return {
+    reject,
+    rejection,
+    clearRejection,
+    getRejectionRoute,
+    getRejectionComponent,
```
stackoverfloweth, line 89: I also wonder if this is something that needs to be exported?

src/utilities/createRouterReject.ts
```diff
@@ -0,0 +1,91 @@
...
+export type RouterRejectionComponents = RegisteredRejection extends never
+  ? { rejections?: BuiltInRejectionComponents }
+  : { rejections: BuiltInRejectionComponents & Record<RegisteredRejection, RouteComponent> }
+
+export type RouterReject = (type: RouterRejection) => void
+
+type GetRejectionComponent = (type: RouterRejection) => RouteComponent
+type GetRejectionRoute = (type: RouterRejection) => Resolved<Route>
+type ClearRejection = () => void
+export type RouterRejectionComponent = Ref<null | { type: RouterRejection, component: RouteComponent }>
```
stackoverfloweth, line 25: IMO this type name is a bit misleading since it's both the type and the component. Suggestion would be just `RouterRejection` and probably rename the existing `RouterRejection` to be `RouterRejectionType`?

stackoverfloweth (commented):
some suggestions, love this solution so much. I really think things like this will really win hearts and minds of our future users

src/utilities/createRouterReject.ts
pleek91: Yep exactly. ❤️

src/utilities/createRouterReject.ts
pleek91: I struggled with this, and your suggestion seems so obvious 😂 will update!

src/utilities/createRouterReject.ts
pleek91: Yeah I had these same questions/thoughts as I was building this and actually wrote some of these originally. 

> Since this route is generated outside of user-land, is it worth tagging it with some symbol that we can theoretically use elsewhere to tell user that current route is a rejection?

This is really only needed for the NotFound rejection when the url pushed to the router (or the initial route) doesn't match any user defined routes. I think some symbol or flag to tell us/users this is a special route for a rejection would be nice.

> Alternatively we could say router.route is either a route or a rejection? That would mean we can avoid the slightly awkward depth, path, query, params that are all not really applicable but undoubtably would have other implications?

I thought about this but ultimately I thought the slightly awkward route for a rejection was worth the annoying devx of having to type guard if router.route is a route or a rejection every time you want to use it. That's a non starter for me personally. 

> Should we just have the reject method return this route instead of having a separate method?

If a rejection happens in middleware or in a component then router.route is still a valid user defined route. But the rejection is shown. I originally did have reject return the route but realized not every rejection actually updates the route. So decided to separate out that and only use `getRejectionRoute` for the somewhat unique `NotFound` rejection that happens when a url doesn't match a route. 


This is a bit nebulous. Let's chat about this sync if this doesn't make sense.

src/utilities/createRouterReject.ts
pleek91: The `getRejectionComponent` piece? Yeah probably not, I think that's left over from a previous iteration where the router handled more of this itself. But I ended up putting pretty much all the logic here in this create method.

src/utilities/createRouterReject.ts
stackoverfloweth: I think I follow you, though if NotFound is so unique I'm kinda leaning towards exporting that responsibility from rejection handling and just saying every router instance automatically has a "NotFound" route. Thoughts on that?

pleek91: I'm hoping to make a distinction between rejections and routes. Route are only associated with a specific (possibly dynamic) url. Vs rejections can appear and be rendered for any url even if that url matches a route. Having NotFound be an actual route means it would have to have an actual url. Which I don't think we want. But lets chat about that. I may have a different opinion after thinking about it for more than 2 minutes.

pleek91: Not found can function exactly like any other rejection. It just also is responsible for a url that doesn't match any route. And the rejection route thing is only there as a stop gap so we can say that router.route always exists. But I wouldn't expect any end user to actually see the special route. They'd have to create their own NotFound component AND use router.route within it. Which is perfectly valid, but still not a main stream use case IMO and I'm fine with a little oddity around that special route.

stackoverfloweth (approved):

PR #73, opened 2024-02-01
added support for router.refresh method

stackoverfloweth:
After all the docs I've written I felt like writing some code.


pleek91:
What do you think about moving refresh into navigation? Otherwise LGTM

stackoverfloweth:
> What do you think about moving refresh into navigation? Otherwise LGTM

@pleek91 how would you imagine that working? I don't think actually want the browser to reload, so I think the best we could do from the navigation service is simulate it with a popstate, which I think is just more confusing?

pleek91 (approved):

PR #74, opened 2024-02-03
Refactor RouterLink and Router.resolve

pleek91:
# Description
Rewriting RouterLink as a sfc because there's some weird typescript generics stuff around defineComponent. The sfc works better and reads a lot nicer. I admit I was not right lol. 

Updated router.resolve to be typesafe with an implementation type. 

Introduced RegisteredRoutes to use for the router link prop. Need to test that works in another project. Be good to figure out how to test the Registered type in unit tests.


src/.eslintrc
pleek91, line 1: Disabling these rules because ts is already checking for this an these rules kinda get annoying.

src/components/routerLink.browser.spec.ts
```diff
@@ -70,7 +70,9 @@ test.each([
 
   app.find('a').trigger('click')
 
-  expect(spy).toHaveBeenLastCalledWith(routeB.path, { replace })
+  const [, arg2] = spy.mock.lastCall ?? []
+
+  expect(arg2).toMatchObject({ replace, query: undefined })
```
pleek91, line 75: This was annoying. Because push is being called with a route method response (which includes functions) `expect(spy).toBeLastCalledWith(router.routes.routeB(), { replace, query: undefined })` failed with the message "Values have no visual difference". They were the same but expect couldn't verify the functions were the same or something. 

Getting the args from last call is pretty messy looking. But it means
- I can check just the options arg
- I can use `toMatchObject` which works best when comparing objects and which there is no `toHaveBeenCalledWith` equivalent.

src/components/routerLink.vue
pleek91, line 1: I love how simple this component became once router.push and router.resolve and routerLink.to all accepted the same type.

src/types/routeWithParams.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type RouteWithParams<
+  TRoutes extends Routes,
+  TRoutePath extends string
+> = {
+  route: TRoutePath & RoutePaths<TRoutes>,
+  params: ExtractRoutePathParameters<RouteMethods<TRoutes>, TRoutePath>,
+}
+
+export type RegisteredRouteWithParams<T extends string> = RouteWithParams<RegisteredRoutes, T>
```
pleek91, line 14: Quick note on this type. I didn't have this type originally and just used `RouteWithParams<RegisteredRoutes, T>` directly in RouterLink. However when inspecting the types output in the build vite was infering the `[]` from `RegisteredRoutes` and inserting that directly. Which means the registered routes would never be used. Probably related to vue-tsc's custom implementation of defineProp typs. Not using `RegisteredRoutes` directly helped keep it in tact.

src/components/routerLink.vue
stackoverfloweth: agreed, this is beautiful. When (if ever) should we add the class support for current match, exact match, etc?

src/components/routerLink.vue
```diff
@@ -0,0 +1,32 @@
...
+  const props = defineProps<{
+    to: string | RegisteredRouteWithParams<T> | ReturnType<RouteMethod>,
+  } & RouterPushOptions>()
+
+  const router = useRouter()
+
+  function resolve(): string {
+    const thing = props.to
+
+    return router.resolve(thing)
```
stackoverfloweth, line 22: I'm assuming `thing` was for console logging

```suggestion
    return router.resolve(props.to)
```

src/components/routerLink.vue
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  function onClick(event: MouseEvent): void {
+    event.preventDefault()
+
+    const { to, ...options } = props
+
+    router.push(props.to, options)
```
stackoverfloweth, line 30: you already have `to` from destructure?

```suggestion
    router.push(to, options)
```

src/components/routerLink.vue
pleek91: Anytime!

src/components/routerLink.vue
pleek91: 🤦‍♂️

src/utilities/createRouterResolve.browser.spec.ts
pleek91, line 1: This was the file that was showing the type error that has something to do with there only being a single route. I spent a lot of time trying to figure out what going on there. But for this PR I realized since I added a RouterResolveImplementation type I could just test createRouterResolve on its own without using createRouter. Which I think is better anyway. And side steps the ts errors for this PR.

PR #75, opened 2024-02-04
Update RouteWithParam types

pleek91:
# Description
Doesn't fix the single route issue but fixes a few other things and cleans up some types a big.

### DIFF
--- a/src/types/routeWithParams.ts
+++ b/src/types/routeWithParams.ts
@@ -7,10 +7,15 @@ export type RouteWithParams<
   TRoutes extends Routes,
   TRoutePath extends string
 > = {
-  route: TRoutePath & RoutePaths<TRoutes>,
-  params: ExtractRoutePathParameters<RouteMethods<TRoutes>, TRoutePath>,
-}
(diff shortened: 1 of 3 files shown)


src/types/routeWithParams.ts
pleek91, line 1: Makes sure that the param property is optional if there are not params or all params are optional

src/utilities/createRouter.ts
```diff
@@ -7,7 +7,7 @@ import { createRouterPush } from '@/utilities/createRouterPush'
 import { createRouterReject } from '@/utilities/createRouterReject'
 import { createRouterResolve } from '@/utilities/createRouterResolve'
 
-export function createRouter<T extends Routes>(routes: T, options: RouterOptions = {}): Router<T> {
+export function createRouter<const T extends Routes>(routes: T, options: RouterOptions = {}): Router<T> {
```
pleek91, line 10: This didn't really fix anything but makes sense in case users inline their routes.

stackoverfloweth (approved):

PR #76, opened 2024-02-05
vitepress docs (first pass)

stackoverfloweth:
Theres plenty of work left to do here, but I thought this was a good point to get it merged and continue with smaller PRs to update docs.

https://docs-init--warm-phoenix-ba1fbe.netlify.app/

What's left in priority order
- need example that demonstrates value of route methods under "navigating.md"
- Rejections page
- Middleware page
- Route Meta page
- Composables page
- Transitions page
- finish vue-router transition page
- build out API of types ([faker](https://fakerjs.dev/api/) is my favorite example of this)
- stackblitz demo
- specialized kitbag logo for router

### DIFF
--- a/docs/core-concepts/query-params.md
+++ b/docs/core-concepts/query-params.md
@@ -1,6 +1,6 @@
 # Query Params
 
-Kitbag/router has first class support for query params.
+Kitbag Router has first class support for query params.
 
 ```ts
 const routes = [
@@ -20,7 +20,7 @@ The router supports dynamic params declared in the query as well.
 ```ts
 import { 
   Routes,
-} from 'kitbag/router'
+} from '@kitbag/router'
 
 const routes = [
   {
@@ -48,7 +48,7 @@ Again like params declared in `path`, these params default as writable `string`,
 import { 
   Routes,
   query, // [!code ++]
-} from 'kitbag/router'
+} from '@kitbag/router'
 
 const routes = [
   {
--- a/docs/core-concepts/route-params.md
+++ b/docs/core-concepts/route-params.md
@@ -5,7 +5,7 @@ When defining your route `path`, you can use a colon `:` to denote dynamic param
 ```ts
 import { 
   Routes,
-} from 'kitbag/router'
+} from '@kitbag/router'
 
 const routes = [
   {
@@ -28,13 +28,13 @@ When defined this way, these params are reactive strings. If you update a param,
 
 ## Param Types
 
-With the `path` function, kitbag/router supports parsing params to types other than `string`.
+With the `path` function, Kitbag Router supports parsing params to types other than `string`.
 
 ```ts
 import { 
   Routes,
   path, // [!code ++]
-} from 'kitbag/router'
+} from '@kitbag/router'
 
 const routes = [
   {
@@ -48,7 +48,7 @@ const routes = [
 
 This will automatically parse the param from `string` in the URL to `number` in `route.params`. If the value cannot be parsed, the route will not be considered a match.
 
-kitbag/router ships with support for `String` (default), `Boolean`, `Number`, and `RegExp`.
+Kitbag Router ships with support for `String` (default), `Boolean`, `Number`, and `RegExp`.
 
 ### RegExp Params
 
@@ -66,7 +66,7 @@ const routes = [
 
 ### Custom Param
 
-You're not limited to the param types that ship with kitbag/router, use `ParamGetter<T>` or `ParamGetSet<T>` to parse params to whatever type you need.
+You're not limited to the param types that ship with Kitbag Router, use `ParamGetter<T>` or `ParamGetSet<T>` to parse params to whatever type you need.
 
 ```ts
 type IdFormat = `${number}-${number}`
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,15 +2,15 @@
 
 ## Installation
 
-Install `kitbag/router` with your favorite package manager
+Install Kitbag Router with your favorite package manager
 
 ```bash
 # bun
-bun add kitbag/router
+bun add @kitbag/router
 # yarn
-yarn add kitbag/router
+yarn add @kitbag/router
 # npm
-npm install kitbag/router
+npm install @kitbag/router
 ```
 
 ## Define Basic Routes
@@ -79,11 +79,11 @@ This component can be mounted anywhere you want route components to be mounted. 
 Use RouterLink for navigating between routes.
 
 ```html {3-4}
-<div id="app">
+<template>
   ...
   <!-- router-link renders as <a> with href -->
   <router-link :to="{ route: 'home' }">Go somewhere</router-link>
-</div>
+</template>
 ```
 
 This component gives the router the power to change the URL without reloading the page.
--- a/docs/migrating-vue-router.md
+++ b/docs/migrating-vue-router.md
@@ -13,7 +13,7 @@
 
 ## Child Routes
 
-Child routes in vue-router have very different behavior depending on if the path starts with "/" or not. In kitbag/router, the behavior is always the same, so add slashes where you want them and leave them off where you don't.
+Child routes in vue-router have very different behavior depending on if the path starts with "/" or not. In Kitbag Router, the behavior is always the same, so add slashes where you want them and leave them off where you don't.
 
 ## Route Regex
 
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -1,21 +1,21 @@
 # Introduction
 
-kitbag/router introduces a fresh, developer-centric approach to routing in Vue.js applications.
+Kitbag Router introduces a fresh, developer-centric approach to routing in Vue.js applications.
 
-At the heart of kitbag/router lies a commitment to enhancing the developer experience.  First and foremost that means **type safety**, but also **better parameter experience**, **support for query**, **rejection handling**, simple intuitive syntax, and an extensible design written with modern Typescript.
+At the heart of Kitbag Router lies a commitment to enhancing the developer experience.  First and foremost that means **type safety**, but also **better parameter experience**, **support for query**, **rejection handling**, simple intuitive syntax, and an extensible design written with modern Typescript.
 
 ## Type Safety
 
-You already know what routes exist, so why are we using magic strings and hoping it works out? With kitbag/router, the routes that are available to you couldn't be clearer. If the routes change, Typescript will tell you the links that need to be updated.
+You already know what routes exist, so why are we using magic strings and hoping it works out? With Kitbag Router, the routes that are available to you couldn't be clearer. If the routes change, Typescript will tell you the links that need to be updated.
 
 ## Better Route Params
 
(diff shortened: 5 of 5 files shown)


.vscode/settings.json
```diff
@@ -4,4 +4,10 @@
     "source.fixAll.eslint": "explicit"
   },
   "typescript.tsdk": "node_modules/typescript/lib",
+  "cSpell.words": [
+    "composables",
+    "Typesafe",
+    "Typesafety",
```
pleek91, line 10: Are these really one word?

docs/getting-started.md
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+```
+
+## RouterView
+
+Give your route components a place to be mounted
+
+```html {3-4}
+<div id="app">
```
pleek91, line 67: Assuming this is the `App.vue` components I don't think it would have `id="app"`. That would be in your main html file that the vue app is mounted to.

docs/getting-started.md
```diff
@@ -0,0 +1,88 @@
...
+## RouterLink
+
+Use RouterLink for navigating between routes.
+
+```html {3-4}
+<div id="app">
+  ...
+  <!-- router-link renders as <a> with href -->
+  <router-link :to="{ route: 'home' }">Go somewhere</router-link>
+</div>
```
pleek91, line 85: Should we do example like this as templates? Like
```suggestion
<template>
  ...
  <!-- router-link renders as <a> with href -->
  <router-link :to="{ route: 'home' }">Go somewhere</router-link>
</template>
```

docs/core-concepts/defining-routes.md
```diff
@@ -0,0 +1,102 @@
+# Defining Routes
+
+## Typesafety
+
+For typesafety to work properly we need to make sure Typescript doesn't widen your routes.
```
pleek91, line 5:
```suggestion
For typesafety to work properly we need to make sure Typescript doesn't widen your routes by using Typescripts `const` and `satisfies` keywords. 
```

docs/core-concepts/route-params.md
```diff
@@ -0,0 +1,156 @@
...
+const routes = [
+  {
+    name: 'users',
+    path: '/users/:id',
+    component: ...
+  }
+] as const satisfies Routes
+```
+
+This means the route will expect 1+ extra string characters (anything but a forward slash `/`) in order to be considered a match. The value of these extra characters is captured and exposed in the `route.params`
```
pleek91, line 19: > anything but a forward slash `/`

Is that true? Or will the regex that's generated match anything? 

On a side note, if my param was a regex that INCLUDED slashes could I match a parameter across slashes?

docs/core-concepts/route-params.md
```diff
@@ -0,0 +1,156 @@
...
+
+```ts
+const route = useRoute()
+
+route.params.id
+```
+
+When defined this way, these params are reactive strings. If you update a param, the url will be updated accordingly.
```
pleek91, line 27: I'm not sure if this will be true or not. But fine to leave in for now. `route` is currently `DeepReadonly`

docs/core-concepts/route-params.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Param Types
```
pleek91, line 29: This is where the AHA moment happens ❤️ The examples are so simple, yet so powerful. Love it. This is the secret sauce to the router. And its so easy to explain.

docs/core-concepts/route-params.md
```diff
@@ -0,0 +1,156 @@
...
+    name: 'users',
+    path: path('/users/:id', { id: /^A[0-9]$/  }), // [!code focus]
+    component: ...
+  }
+] as const satisfies Routes
+```
+
+### Custom Param
+
+You're not limited to the param types that ship with kitbag/router, use `ParamGetter<T>` or `ParamGetSet<T>` to parse params to whatever type you need.
```
pleek91, line 69: I think after the intro we should just refer to this as "router" rather than "kitbag/router". But I don't feel strongly. So keep it if you like it.

docs/core-concepts/route-params.md
```diff
@@ -0,0 +1,156 @@
...
+    name: 'users',
+    path: path('/users/:id', { id: myTypeParam }), // [!code focus]
+    component: ...
+  }
+] as const satisfies Routes
+```
+
+With this getter defined, now our route will only match if the param matches our rules above.
+
+As a `ParamGetter`, the value in `route.params` is still writable, but the set will assume `value.toString()` is sufficient. Alternatively if you use `ParamGetSet`, you can provide the same validation on value set as well.
```
pleek91, line 107: Since `ParamGetter` expects to be able to call `toString` we should probably have this example work with that. And maybe use this example for `ParamGetSet`

docs/core-concepts/query-params.md
```diff
@@ -0,0 +1,92 @@
...
+
+```ts
+const route = useRoute()
+
+route.params.sort
+```
+
+## Param Types
+
+Again like params declared in `path`, these params default as writable `string`, but can be overridden with built in `Number`, `Boolean`, or any `ParamGetter`/`ParamGetSet` you define.
```
pleek91, line 45:
```suggestion
Again like params declared in `path`, these params default as writable `string`, but can be overridden with built in `Number`, `Boolean`, `RegExp`, or any `ParamGetter`/`ParamGetSet` you define.
```

docs/core-concepts/navigating.md
```diff
@@ -0,0 +1,133 @@
...
+
+const router = useRouter(routes)
+
+router.push({ route: 'user.settings' }) // [!code --]
+router.push({ route: 'user.settings', params: { id: 42, tab: 'github' } }) // [!code ++]
+```
+
+### Query
+
+With the optional 2nd argument, you add additional query params. Note if you know your route will expect certain parameters, [query params](/core-concepts/query-params) is a much better developer experience for dealing with the query.
```
pleek91, line 78: We might want to distinguish between the raw query and query params. 
```suggestion
With the optional 2nd argument, you add additional values to the query. Note if you know your route will expect certain parameters, [query params](/core-concepts/query-params) is a much better developer experience for dealing with the query.
```

docs/advanced-concepts/reusing-param-names.md
```diff
@@ -0,0 +1,79 @@
...
+            component: ... 
+          },
+        ]
+      }
+    ]
+  }
+] as const satisfies Routes
+```
+
+Notice how in this example the param `id` is used at every level. The router treats this as a single param with a value that is a tuple with each occurrence having a spot. So calling ``, would expect `params: { id: [string, string, string] }`.
```
pleek91, line 29: So calling what?

docs/advanced-concepts/reusing-param-names.md
pleek91, line 1: Such a good page. Like I hope people don't use this feature. But its explained clearly and it "just makes sense".

docs/migrating-vue-router.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Child Routes
+
+Child routes in vue-router have very different behavior depending on if the path starts with "/" or not. In kitbag/router, the behavior is always the same, so add slashes where you want them and leave them off where you don't.
+
+## Route Regex
+
+### Hash History Mode
+
+## Repeatable Params
```
pleek91, line 22: I'm curious if this is supported with a regex param. Like couldn't you just write a regex that would accomplish this?

pleek91 (approved):
This is so awesome to see. Nice job. That was A LOT of work 👏

.vscode/settings.json
```diff
@@ -4,4 +4,10 @@
     "source.fixAll.eslint": "explicit"
   },
   "typescript.tsdk": "node_modules/typescript/lib",
+  "cSpell.words": [
+    "composables",
```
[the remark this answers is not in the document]
stackoverfloweth: agreed. I fought it at first but trying to stay current

.vscode/settings.json
stackoverfloweth: Probably not, I'll update

docs/getting-started.md
stackoverfloweth: I did [what vue-router did](https://router.vuejs.org/guide/#HTML), but I agree with you. I'll move from id= to class=? encourage BEM lol

docs/getting-started.md
stackoverfloweth: I don't think so, at least not here. I think it's worth explaining to some developer that's new to the concept that this needs to be in root-level and while it's subtle, `<div class="app">` does convey more information than a generic `<template/>`

docs/core-concepts/route-params.md
stackoverfloweth: dude! This is why we need docs, I didn't even know how the thing I wrote works lol. You're definitely right, wrote tests to prove it. This is super cool that params are so flexible and regex params aren't restricted. You could basically circumvent our entire route match logic by just having a single param with your own complex regex pattern.

docs/core-concepts/route-params.md
stackoverfloweth: okay, I'll add an issue for us to reconcile this difference before going live

stackoverfloweth: https://github.com/kitbagjs/router/issues/81

docs/core-concepts/route-params.md
stackoverfloweth: I had a similar thought but decided to keep kitbag there because vue-router always refers to their router as "Vue Router", tanstack always refers to their roouter as "Tanstack Router", just seems like the norm. 

I would be open to using "Kitbag Router" instead of "kitbag/router" if you like that more?

docs/core-concepts/route-params.md
stackoverfloweth: yep, great call

docs/advanced-concepts/reusing-param-names.md
stackoverfloweth: agreed. I was a little disappointed to even have to write it because I don't want to shine a light on it, might even find another category like `core-concepts`, `advanced-concepts`, and `mostly-ignoreable-concepts` lol

docs/advanced-concepts/reusing-param-names.md
stackoverfloweth: oops

docs/migrating-vue-router.md
stackoverfloweth: we talked about this, you're right that with params that extend beyond the forward slash we can theoretically support repeatable params, but that's going to change and we're not sure we will support that at launch. Keeping this under not supported for now

docs/getting-started.md
pleek91: I'm not following you here. What needs to be root level?

docs/core-concepts/route-params.md
pleek91: IMO that's more readable 👍

docs/getting-started.md
stackoverfloweth: oops, sorry I assumed this was the `<router-view/>` piece. I agree with you now that I actually read it more closely 😅

PR #79, opened 2024-02-08
rearrangement suggestion

stackoverfloweth:
I personally found the number of args for `createRouterRoute` to be a bit of a deal breaker. IMO that implied that it was doing too much so I set out to move tasks back into the router. I'm not opposed to breaking some of those things out again into separate services, just not `createRouterRoute`. Even with the extra logic back in the router, I find this arrangement is more of what I expected.

I think it also will ensure that middleware runs for initial route and not found routes

it also establishes notFoundRoute as a const so checking if the current route is the not found route doesn't use the route name. I think this is an improvement because our end users could also define a route called `NotFound`.


src/components/routerView.browser.spec.ts
```diff
@@ -114,7 +114,7 @@ test('updates components when route changes', async () => {
   expect(app.html()).toBe(childA.component.template)
 })
 
-it.each([
+test.each([
```
stackoverfloweth, line 117: this was the only file using `it`, just opting for consistency here

src/utilities/createRouterReject.ts
```diff
@@ -5,6 +5,22 @@ import { RegisteredRejectionType, Resolved, Route, RouteComponent } from '@/type
 export const builtInRejections = ['NotFound'] as const
 export type BuiltInRejectionType = typeof builtInRejections[number]
 
+const notFoundRoute = {
+  name: Math.random().toString(),
```
stackoverfloweth, line 9: the point here is that the name doesn't matter.. We could probably use a more elegant way of "random string"

src/utilities/createRouterReject.ts
```diff
@@ -47,26 +61,6 @@ export function createRouterReject({
...
-      depth: 0,
-      path: '',
-      query: '',
-      params: {},
-    }
-
-    return resolved
-  }
-
   const clearRejection: ClearRejection = () => {
```
stackoverfloweth, line 64: I also liked how this change simplifies this contract, no longer needs to export `getRejectionRoute`, which was unique to NotFound

src/utilities/createRouterReject.ts
pleek91: I feel pretty strongly we shouldn't use a random string for this because it is possible an end user could see this. Something specific like `NotFoundRejectionRoute` should be fine.

src/utilities/createRouterReject.ts
pleek91: So the problem with this is it doesn't consider that the user can provide their own component. Which is basically the only reason this utility existed in the first place.

src/utilities/createRouterRoute.ts
```diff
@@ -2,49 +2,35 @@ import { DeepReadonly, reactive, readonly } from 'vue'
...
-    const matched = routeMatch(resolved, url)
-
-    if (!matched) {
-      reject('NotFound')
-      setRoute(getRejectionRoute('NotFound'))
-      return
-    }
+  const updateRoute: RouterRouteUpdate = (matched: Resolved<Route>) => {
+    setRoute(matched)
+  }
```
pleek91, line 29: These two methods have the exact same type. We probably don't need both. Can just put the object.assign right here in this method.

src/utilities/createRouterRoute.ts
```diff
@@ -2,49 +2,35 @@ import { DeepReadonly, reactive, readonly } from 'vue'
...
 
+  const executeMiddleware: RouterRouteMiddleware = async (matched) => {
```
pleek91, line 31: We should just pull this out and pass the current route into it as another arg. Then this can be its own utility.

src/utilities/createRouter.ts
```diff
@@ -2,32 +2,56 @@ import { App } from 'vue'
...
+
+    try {
+      await executeMiddleware(matched)
+      updateRoute(matched)
+      clearRejection()
+    } catch (error) {
+      if (error instanceof RouterRejectionError) {
+        reject(error.type)
+      }
+    }
```
pleek91, line 35: This isn't quite right. When a reject happens the route should still be updated. Which I think I had wrong in my PR as well. 

Also we cannot catch and silence every error. So we'll need to re throw if the error is not a specific instance we're expecting. I'm also thinking since we already have navigation here, we might as well just catch the middleware errors directly here and do navigation. Rather than passing navigation into createRouterRoute.

pleek91 (commented):
I think theres some good stuff here. Wanna build on this. I'm going to make a PR against this PR.

src/utilities/createRouterReject.ts
stackoverfloweth: I don't see how I broke that functionality? Tests all passed?

src/utilities/createRouter.ts
stackoverfloweth: I think you resolved this in your PR?

src/utilities/createRouterRoute.ts
stackoverfloweth: agreed. I think you resolved this in your PR?

src/utilities/createRouterRoute.ts
stackoverfloweth: agreed. I think you resolved this in your PR?

src/utilities/createRouterReject.ts
pleek91: That's because I don't think there's a test for that lol. The rejection works just fine. But if you were to look at router.route for the not found route you'd see the library's component and not the custom component.

src/utilities/createRouterRoute.ts
pleek91: I did

src/utilities/createRouterRoute.ts
pleek91: Yep I did

src/components/routerView.browser.spec.ts
```diff
@@ -181,10 +185,13 @@ it('Renders custom NotFound component when the initialUrl does not match', () =>
     },
   })
 
-  expect(app.text()).toBe(template)
+  const route = mount(router.route.matched.component)
+
+  expect(app.text()).toBe(NotFound.template)
+  expect(route.text()).toBe(NotFound.template)
```
pleek91, line 191: @stackoverfloweth 
This is the failing test to show what I'm talking about with the rejection route. Both `rejection` and `route` get set. In this branch the rejection is set correctly. But the `route` still has the library defined not found component rather than the user defined one.

pleek91 (approved):

PR #80, opened 2024-02-09
rearrangement suggestion suggestions

pleek91:
# Description
Continuing on the ideas/progress from https://github.com/kitbagjs/router/pull/79

### DIFF
--- a/src/utilities/createRouter.ts
+++ b/src/utilities/createRouter.ts
@@ -1,15 +1,16 @@
-import { App } from 'vue'
+import { App, readonly } from 'vue'
 import { RouterLink, RouterView } from '@/components'
 import { routerInjectionKey, routerRejectionKey } from '@/compositions'
 import { Routes, Router, RouterOptions, RouterImplementation } from '@/types'
-import { RouterRejectionError } from '@/types/errors'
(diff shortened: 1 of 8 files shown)


src/utilities/createRouter.ts
```diff
@@ -20,19 +21,42 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+        return
+      }
+
+      if (error instanceof RouterReplaceError) {
+        const [source] = error.to
+        const url = resolve(source)
+
+        navigation.update(url, { replace: true })
+        return
+      }
```
pleek91, line 51: IMO moving this here makes a lot of sense. Leaves the router itself responsible for navigation.

src/utilities/createRouter.ts
```diff
@@ -20,19 +21,42 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+
+      if (!(error instanceof RouterRejectionError)) {
+        throw error
+      }
```
pleek91, line 55: making sure we still throw in case its a user implementation error.

stackoverfloweth (approved):

PR #87, opened 2024-02-15
Clean up how private symbol properties work in Resolved

pleek91:
# Description
Adds a new utility called `createResolvedRoute` that accepts all the properties of a resolved route (including things like `isRejection` and `depth`) and returns `Resolved<T>` where `isRejection` and `depth` are their symbol properties. 

This will allow us to easily add functionality to the `Resolved` type while not exposing non public values.


src/utilities/createResolvedRoute.ts
```diff
@@ -0,0 +1,34 @@
...
+type ResolvedRouteProperties<T extends Route> = {
+  matched: T,
+  matches: Route[],
+  name: string,
+  path: string,
+  query: string,
+  params: Record<string, Param[]>,
+  depth: number,
+  isRejection?: true,
+}
```
stackoverfloweth, line 13: Many of these properties are the same as `Resolved<Route>` right? Would it be better to define this as resolved route & the extra properties used to assign to symbols?

stackoverfloweth (approved):
Looks really nice. Great idea

src/utilities/createResolvedRoute.ts
pleek91: I tried that originally but the omit and add type was kinda hard to read. This seemed simpler even if it's less dry

PR #88, opened 2024-02-16
Simplify `Resolved<T>` type to `ResolvedRoute`

pleek91:
# Description
Cleans up `Resolved` type to be simpler based on decision to make it an internal only type. Will follow this with a PR to remove Resolved from things like router.route and middleware types

- Removes generic `Route` since there's no reason to type guard this type
- Removes symbol properties in favor of strings
- Renames to `ResolvedRoute`

### DIFF
--- a/src/utilities/createResolvedRoute.ts
+++ b/src/utilities/createResolvedRoute.ts
@@ -1,5 +1,5 @@
 import { markRaw } from 'vue'
-import { Route, Param, Resolved, routeDepthSymbol, isRejectionSymbol } from '@/types'
+import { Route, Param, Resolved } from '@/types'
 
 type ResolvedRouteProperties<T extends Route> = {
   matched: T,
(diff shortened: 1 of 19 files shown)


src/utilities/createResolvedRoute.ts
```diff
@@ -1,34 +1,26 @@
...
   path: string,
   query: string,
   params: Record<string, Param[]>,
   depth: number,
-  isRejection?: true,
+  isRejection: boolean,
 }
 
-export function createResolvedRoute<T extends Route>(route: ResolvedRouteProperties<T>): Resolved<T> {
+export function createResolvedRoute(route: ResolvedRouteProperties): ResolvedRoute {
```
pleek91, line 15: I decided to keep this only because currently we're using it as a place to mark the components as raw.

stackoverfloweth (approved):
this looks like what I expected, though tbh I still like the syntax of `Resolved<Route>` more than `ResolvedRoute`. Even if we're not really utilizing the underlying generic

pleek91:
> this looks like what I expected, though tbh I still like the syntax of `Resolved<Route>` more than `ResolvedRoute`. Even if we're not really utilizing the underlying generic

The naming syntax is nice but it means we're passing Route as the generic in a lot of places and it doesn't provide any utility. So seems like we should remove it.

PR #89, opened 2024-02-17
Router link classes

stackoverfloweth:
added `router.find` method

takes `source`, same type as `router.push` and `RouterLink.to`

first calls `resolve` to convert source into url, then passes that url into existing `routeMatch` logic

returns `ResolvedRoute | undefined`

### DIFF
--- a/src/components/routerLink.vue
+++ b/src/components/routerLink.vue
@@ -23,7 +23,7 @@
   const isMatched = computed(() => !!route.value && router.route.matched === route.value)
 
   const classes = computed(() => ({
-    'router-link--matches': inMatches.value,
+    'router-link--match': inMatches.value,
     'router-link--exact-match': isMatched.value,
   }))
 
--- a/src/components/routerLink.vue
+++ b/src/components/routerLink.vue
@@ -1,10 +1,11 @@
 <template>
-  <a :href="resolve()" @click="onClick">
+  <a :href="resolve()" :class="classes" @click="onClick">
     <slot />
   </a>
 </template>
 
 <script setup lang="ts" generic="T extends string">
+  import { computed } from 'vue'
   import { useRouter } from '@/compositions'
   import { RouteMethod } from '@/types/routeMethod'
   import { RegisteredRouteWithParams } from '@/types/routeWithParams'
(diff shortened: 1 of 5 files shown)


src/components/routerLink.vue
```diff
@@ -16,6 +17,16 @@
...
+  const route = computed(() => router.find(props.to)?.matched)
+
+  const inMatches = computed(() => !!route.value && router.route.matches.includes(route.value))
+  const isMatched = computed(() => !!route.value && router.route.matched === route.value)
+
+  const classes = computed(() => ({
+    'router-link--matches': inMatches.value,
+    'router-link--exact-match': isMatched.value,
+  }))
+
```
pleek91, line 29: Maybe --match and --exact-match?

src/components/routerLink.vue
```diff
@@ -16,6 +17,16 @@
 
   const router = useRouter()
 
+  const route = computed(() => router.find(props.to)?.matched)
+
+  const inMatches = computed(() => !!route.value && router.route.matches.includes(route.value))
+  const isMatched = computed(() => !!route.value && router.route.matched === route.value)
```
pleek91, line 23: I love how simple this is.

src/utilities/createRouterFind.ts
pleek91, line 1: Such a nice little utility

pleek91 (commented):
New code looks good. Looks like there's a test we need to fix though.

src/utilities/createRouterFind.ts
stackoverfloweth: agreed. Simple enough that I didn't bother with unit tests

src/types/router.ts
```diff
@@ -19,6 +20,7 @@ export type Router<
   resolve: RouterResolve<TRoutes>,
   push: RouterPush<TRoutes>,
   replace: RouterReplace<TRoutes>,
+  find: RouterFind<TRoutes>,
```
pleek91, line 23: Since this utility returns ResolvedRoute types, do we even want to add this to our public api? That would mean ResolvedRoute is public and we'd have to make any changes to it backwards compatible. I'm hoping ResolvedRoute can be completely internal

pleek91: Well I guess this would just return RouterRoute actually 🤔 ignore me

pleek91 (approved):

PR #90, opened 2024-02-18
`RouterRoute` type and implementation

pleek91:
# Description
Updates the type definition and implementation of `router.route` to use a new `RouterRoute` type. `ResolvedRoute` is now a completely internal type and should never be exposed publicly. 

Note: This does not yet solve the route specific types issue. Which will require some more work. I have some ideas I want to try out for that. But this PR was big enough on its own.


src/utilities/createRouterReject.ts
```diff
@@ -50,26 +57,29 @@ export function createRouterReject({
...
+      [isRejectionRouteSymbol]: true,
+    }
 
     return resolved
   }
 
+  const isRejectionRoute: IsRejectionRoute = (route: RouterRejectionRoute) => {
+    return route[isRejectionRouteSymbol] === true
+  }
+
```
pleek91, line 82: I thought this turned out nicely. The type for a rejection route is slightly wider than RouterRoute. But since the `isRejectionRouteSymbol` is optional it accepts `RouterRoute` just fine.

src/utilities/createRouterRouteQuery.ts
pleek91, line 1: `RouterRouteQuery` is basically a readonly version of `URLSearchParams`. I think the syntax and devx is really nice. And we can add things to it if needed. Which is a nice benefit of not just using `Record<string, string | string[] | null>` like vue-router does.

src/utilities/paramValidation.ts
pleek91, line 1: I separated getting route params from validating route params. `getRouteParams` can still throw an error and `routeParamsAreValid` catches that error and always returns a boolean. But this way after we've matched a round we can call `getRouteParams` to get the actual param values knowing that it won't error because the params have already been validated. 

Might be nice to rework this at some point so we don't have to call getRouteParams twice. But for now this works nicely.

src/utilities/routeMatch.ts
pleek91, line 1: this got renamed to `getRouterRouteForUrl` and moved to `utilities/routes.ts`

src/utilities/paramValidation.ts
```diff
@@ -1,18 +1,31 @@
...
 
-    try {
-      paramsTuple.forEach((param, index) => getParamValue(stringValues[index], param))
-    } catch {
-      return false
+    if (paramsTuple.length === 1) {
+      const [value] = values
+      params[key] = value
+      continue
     }
```
pleek91, line 25: Note that if only a single param is expected its "unwrapped"

src/utilities/routes.browser.spec.ts
```diff
@@ -163,7 +163,58 @@ test('given route with equal matches, returns route with highest score', () => {
...
+    component,
+  } as const satisfies Route
+
+  const resolved = resolveRoutes([route])
+  const response = getRouterRouteForUrl(resolved, '/A?paramB=B')
+
+  expect(response?.params).toMatchObject({
+    paramA: 'A',
+    paramB: 'B',
+  })
```
pleek91, line 219: this test is failing because of a bug in finding param values. I thinks the value for `paramA` is `A?paramB=B`

src/utilities/paramValidation.ts
stackoverfloweth: `nit`, I always try to avoid "continue". How would you feel about just using `else` instead?

```ts
if (paramsTuple.length === 1) {
  const [value] = values
  params[key] = value
} else {
  params[key] = values
}
```

src/utilities/middleware.browser.spec.ts
```diff
@@ -2,50 +2,73 @@ import { expect, test, vi } from 'vitest'
...
 
-  const [resolvedA] = resolveRoutes([routeA])
-  const execute = (): Promise<void> => executeMiddleware({ to: resolvedA, from: null })
+  const routerRoute: RouterRoute = {
+    matched: route,
+    matches: [route],
+    name: route.name,
+    query: createRouterRouteQuery(),
+    params: {},
+  }
```
stackoverfloweth, line 69: `nit` do you think we should have an internal utility like `resolveRoutes`, but for turning a `Route` into `RouterRoute`? I've seen blocks like this in several places in several files

src/utilities/middleware.browser.spec.ts
```diff
@@ -2,50 +2,73 @@ import { expect, test, vi } from 'vitest'
...
-  executeMiddleware({ to: resolvedA, from: resolvedB })
+  executeMiddleware({ to: routerRouteA, from: routerRouteB })
 
   expect(middleware).toHaveBeenCalledOnce()
 
   const [to, { from }] = middleware.mock.lastCall
-  expect(to).toMatchObject(resolvedA)
-  expect(from).toMatchObject(resolvedB)
+  expect(to).toMatchObject(routerRouteA!)
+  expect(from).toMatchObject(routerRouteB!)
```
stackoverfloweth, line 48: pretty sure these aren't necessary

```suggestion
  expect(to).toMatchObject(routerRouteA)
  expect(from).toMatchObject(routerRouteB)
```

src/utilities/middleware.browser.spec.ts
stackoverfloweth: something else to consider which tripped me up is that we have a thing called `createRouterRoute` but it doesn't do what I suggested, it creates `router.route` not `RouterRoute` 🤔

src/utilities/createRouterRoute.ts
```diff
@@ -1,15 +1,24 @@
...
-type RouterRoute = {
-  route: ResolvedRoute,
+type CreateRouterRoute = {
+  route: RouterRoute,
   updateRoute: RouterRouteUpdate,
 }
 
-export function createRouterRoute(fallbackRoute: ResolvedRoute): RouterRoute {
-  const route = reactive<ResolvedRoute>(fallbackRoute)
+export function createRouterRoute(fallbackRoute: RouterRoute): CreateRouterRoute {
```
stackoverfloweth, line 20: if we're going to have a thing called `RouterRoute`, I think we should rename this file & function. Perhaps `createCurrentRoute` or `createReactiveRoute`.

src/utilities/createRouterReject.ts
stackoverfloweth: yes! Love that ❤️

src/utilities/paramValidation.ts
pleek91: Going to prefer continue over else to follow the "return early" pattern.

src/utilities/middleware.browser.spec.ts
pleek91: We have many `createRouterX` methods that are factories for the various router properties. So this is consistent with that pattern. 

To create a `RouterRoute` you really need a resolved route AND the query and params. And in the codebase we only do that in one place (`getRouterRouteForUrl`). But in tests I created some manually so that way I could test functionality separately without depending on other utilities or services. I considered a test utility for this but ultimately decided the tests were simple enough.

src/utilities/createRouterRoute.ts
pleek91: how is this any different than `createRouterPush` or the other factory methods used for creating the router object?

src/utilities/routes.ts
```diff
@@ -1,5 +1,36 @@
...
+  const matches = routes
+    .filter(route => rules.every(test => test(route, url)))
+    .sort(sortByRouteScore)
+
+  if (matches.length === 0) {
+    return undefined
+  }
+
+  const [route] = matches
+  const [, search] = url.split('?')
```
stackoverfloweth, line 22: I think using browser `URL` is probably preferred here?

```suggestion
  const { search } = createMaybeRelativeUrl(url)
```

src/utilities/createRouterRoute.ts
stackoverfloweth: because `RouterRoute` is a thing, when I see a function `createRouterRoute` I expect it to return `RouteRoute`, not take `RouterRoute` as an arg and return `CreateRouterRoute`. Our naming convention is causing conflicts here

src/utilities/routes.ts
pleek91: Doesn't that require a full url with a domain?

stackoverfloweth: that's why I suggested using `createMaybeRelativeUrl`, which accounts for relative URLs where `new URL()` does not

src/utilities/createRouterRoute.ts
pleek91: I mean it does return a RouterRoute and a method to update that route. So I don't see any conflict here. But we could change our naming convention away from creatRouterX to something like routerXFactory which would leave createX available for the use case you want. 

But we've already talked about swapping RouterRoute with ResolvedRoute so does that help at all?

Either way, the code is solid yeah? Ultimately we're not gonna be able to name everything perfect. And id like our router.thing to all follow the same naming convention. So unless we're changing that convention IMO this should be called createRouterRoute. 

But I'd be fine with any other convention as long as we change them all.

src/utilities/routes.ts
pleek91: But it just throws an error

pleek91: And this needs to work in node in order to render the initialUrl

src/utilities/createRouterRoute.ts
stackoverfloweth: yes the code is solid. I think these discussions over naming are worthwhile, despite how tedious it may seem. This function returns a thing called `CreateRouterRoute`, so following our convention of methods like `createRouterPush`, `createRouterFind`, etc this should be `createCreateRouterRoute` as ugly as that is. Perhaps there's a better term for this thing? What do you think of naming the type `RouterCurrentRoute` and the method `createRouterCurrentRoute`? or `RouterActiveRoute`, something like that?

pleek91: The current pattern is all of the properties of `Router` are made in factories following the `createRouterX` pattern where `X` is the router property. Some of those factories return multiple values to support the primary `X` being created. They return `X` directly OR `X` is one of the values it returns. Both `createRouterRoute` and `createRouterReject` return `CreateRouterX` because they return more than just the router property they're responsible for creating. Which all makes sense to me. 

If you feel strongly that any `createX` method should return `X` and only `X` then we need a different naming convention for these factory methods. I suggested `routerXFactory` (`routerRouteFactory`, `routerPushFactory`, etc) where `X` is the router property NOT the type being returned. Do you have any thoughts on that suggestion?

stackoverfloweth (approved):

PR #92, opened 2024-02-19
Matched route type refactor

stackoverfloweth:
updates `ResolvedRoute.params` to be `pathParams` and `queryParams` so that we can work with their values in the URL independently


src/utilities/paramValidation.browser.spec.ts
stackoverfloweth, line 1: this had to go to browser because it calls param validation now calls `createMaybeRelativeUrl`, which only works in browser context

src/utilities/paramValidation.ts
```diff
@@ -1,21 +1,30 @@
...
 
-export const getRouteParams = (route: ResolvedRoute, url: string): Record<string, unknown> => {
+export const getRouteParamValues = (route: ResolvedRoute, url: string): Record<string, unknown> => {
+  const { pathname, search } = createMaybeRelativeUrl(url)
+
+  return {
+    ...getRouteParams(route.pathParams, route.path, pathname),
+    ...getRouteParams(route.queryParams, route.query, search),
+  }
+}
```
stackoverfloweth, line 21: I left this function returning the combination of both, but this will need to be reworked. We can't just spread the values together, currently I believe query params of the same name as path would just override

src/utilities/routes.ts
```diff
@@ -21,7 +21,7 @@ export function getRouterRouteForUrl(routes: ResolvedRoute[], url: string): Rout
   const [route] = matches
   const [, search] = url.split('?')
   const query = createRouterRouteQuery(search)
-  const params = getRouteParams(route, url)
+  const params = getRouteParamValues(route, url)
```
stackoverfloweth, line 24: just renamed the exported function, now `getRouteParam` is the internal function that generically pulls values off either path or query params

src/utilities/urlAssembly.ts
```diff
@@ -1,8 +1,9 @@
 import { Param, ResolvedRoute } from '@/types'
+import { mergeParams } from '@/utilities/mergeParams'
 import { setParamValuesOnUrl } from '@/utilities/paramsFinder'
 
 export function assembleUrl(route: ResolvedRoute, values: Record<string, unknown[]> = {}): string {
-  const params = Object.entries<Param[]>(route.params)
+  const params = Object.entries<Param[]>(mergeParams(route.pathParams, route.queryParams))
```
stackoverfloweth, line 6: this should work fine, but given that these are now separate on the resolved route, I probably would have assigned values when building the url separately. Seems silly to merge together but required less refactoring

src/utilities/paramValidation.ts
stackoverfloweth: https://github.com/kitbagjs/router/issues/93

src/utilities/paramValidation.browser.spec.ts
pleek91: Doesn't that mean we cannot load any routes in node? This is something we need to address. The majority of our tests should be running in node and not browser environments. Testing in a browser environment should be reserved for functionality we expect to only work in a browser such has history.

stackoverfloweth: this is addressed by [this change](https://github.com/kitbagjs/router/pull/94)

src/utilities/paramValidation.ts
stackoverfloweth: great idea! https://github.com/kitbagjs/router/pull/92/files#diff-8875ff68754a59b04f6ca606e3f5d6a54351ee99c43a568ad292d4905e6d3c3cR105

pleek91 (approved):

PR #96, opened 2024-02-20
Expand middleware concept into hooks

pleek91:
# Description
Makes what we now call middleware much more flexible by allowing for global, route, and component level middleware. There are currently three lifecycle hooks
- onBeforeRouteEnter
- onBeforeRouteUpdate
- onBeforeRouteLeave

The plan is to have three additional `after` hooks
- onAfterRouteEnter
- onAfterRouteUpdate
- onAfterRouteLeave

You can define a hook for any lifecycle at any level (global, route, component). Which is a gripe I've often had with vue-router. 

I'm putting this PR up to discuss before going any further. But things that still need to be done are
- implement `after` hooks
- test global and component level hooks actually executed as expected
- rename the `Middleware` to `RouteHook`


src/components/routerView.ts
pleek91, line 1: Nothing changed functionally here. I just converted the depth injection into a composition to use in the component level hooks.

src/types/routes.ts
pleek91, line 1: Need to clean up these types. But didn't want to address that in this PR.

src/utilities/createRouter.ts
```diff
@@ -5,58 +5,76 @@ import { Routes, Router, RouterOptions, RouterImplementation } from '@/types'
...
+        ...hooks.before,
+        ...getRouteHooks(to, 'before'),
+      ],
+      to,
+      from,
+      onMiddlewareError,
+    })
+
+    if (!success) {
+      return
```
pleek91, line 74: executeMiddleware use to just accept `to`, and `from` and then it would grab the middleware from the `to` route and throw errors. 

Now you have to tell it what middleware to execute AND how to handle errors. I was looking for a way to abstract the error handling because I'll need to execute middleware in two places (for before and after hooks) but use the same error handling logic. 

I think passing in an error handler makes sense because its expected that this function will throw errors. So it should never be used without an error handler.

src/utilities/createRouterHooks.ts
pleek91, line 1: This is a good place to explain my thinking behind how hooks are registered and executed. There are two types of hooks `before` and `after`. The type determines where in the `onLocationUpdate` function a hook gets called. 

When registering a hook, its wrapped in a conditional check to determine if based on the specific lifecycle requested. So when middleware is executed, it executes all the `before` hooks at the same time. But that wrapping condition determines if the user provided middleware function should actually be run.

src/utilities/createRouterHooks.ts
```diff
@@ -0,0 +1,96 @@
...
+  return to.matches[1] !== from?.matches[1]
+}
+
+function isGlobalUpdate(to: RouterRoute, from: RouterRoute | null): boolean {
+  return to.matches.some((route, depth) => route === from?.matches[depth])
+}
+
+function isGlobalLeave(to: RouterRoute, from: RouterRoute | null): boolean {
+  return to.matches[1] !== from?.matches[1]
+}
```
pleek91, line 33: These are the lifecycle conditions for global hooks. Currently enter and leave are the same. But seemed worth separating out for consistency and easy naming.

src/utilities/createRouterHooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const isGlobalUpdate: GlobalHookCondition = (to, from) => {
+  return to.matches.some((route, depth) => route === from?.matches[depth])
+}
+
+const isGlobalLeave: GlobalHookCondition = (to, from) => {
+  return to.matches[1] !== from?.matches[1]
+}
```
pleek91, line 36: These are the lifecycle conditions for global hooks. Currently enter and leave are the same. But seemed worth separating out for consistency and easy naming.

src/utilities/hooks.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const removeAll: RouteHookRemove = () => remove.forEach(fn => fn())
+
+    onUnmounted(removeAll)
+
+    return removeAll
+  }
+}
+
+const isComponentEnter: ComponentHookCondition = (to, from, depth) => {
```
pleek91, line 49: Starting here you can see the lifecycle conditions for component level hooks. I think these same conditions will be used for route level hooks.

src/utilities/createRouterHooks.ts
```diff
@@ -0,0 +1,77 @@
+import { InjectionKey } from 'vue'
+import { RouteMiddleware } from '@/types'
+import { asArray } from '@/utilities/array'
+import { RouterRoute } from '@/utilities/createRouterRoute'
+
+export type RouteHookRemove = () => void
+export type AddRouteHook = (hook: RouteMiddleware) => RouteHookRemove
+export type RouteHookType = 'before' | 'after'
+export type RouteHookLifeCycle = 'onBeforeRouteEnter' | 'onBeforeRouteLeave' | 'onBeforeRouteUpdate'
```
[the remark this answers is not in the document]
pleek91: I moved these into `/types/hooks.ts`

stackoverfloweth (approved):

PR #97, opened 2024-02-20
naming refactor 2, mostly ResolvedRoute and RouterRoute swap

stackoverfloweth:


pleek91 (approved):
Nice, that was a lot more than I imagined. 🙇

PR #98, opened 2024-02-21
Add hook conditions to route level hooks

pleek91:
# Description
Updates the `getRouteHooks` method to return only hooks that should be run based on the `RouteHookType`


pleek91:
@stackoverfloweth I think this is ready based on what we talked about. I didn't implement the onBeforeLocationUpdate in this PR but that will be the next piece I work on. We'll need a little more functionality in `navigation` to account for actually aborting navigation. So should have a fairly small PR to implement that. I also need to write tests for the component and global hooks. So this will probably be a few PRs before this is buttoned up.

stackoverfloweth (approved):

PR #101, opened 2024-02-25
added slot props to router-link and router-view

stackoverfloweth:
this is untested


stackoverfloweth:
looking for a way to provide type context for router-view being that it's not a SFC and can't use `defineSlots`

https://stackoverflow.com/questions/78058105/vue-defineslots-functional-component

src/components/routerView.ts
```diff
@@ -35,6 +35,10 @@ export default defineComponent({
         return h(rejection.value.component)
       }
 
+      if (slots.default) {
+        return slots.default({ component, route: router.route })
+      }
```
pleek91, line 40: What are you thinking here?

src/components/routerLink.vue
```diff
@@ -15,21 +15,29 @@
...
-    'router-link--exact-match': isMatched.value,
+    'router-link--match': match.value,
+    'router-link--exact-match': exactMatch.value,
   }))
 
-  function resolve(): string {
-    return router.resolve(props.to)
-  }
+  const resolved = computed(() => router.resolve(props.to))
+  const isExternal = computed(() => resolved.value.startsWith('http'))
```
pleek91, line 40: Idk that we can ensure this is true

src/components/routerView.ts
stackoverfloweth: Following the pattern vue-router uses. Gives developers a chance to put "middleware" components between router-view and the end component.

pleek91: Sure 🤷

pleek91:
> looking for a way to provide type context for router-view being that it's not a SFC and can't use `defineSlots`
> 
> https://stackoverflow.com/questions/78058105/vue-defineslots-functional-component

We could probably write router-view as a sfc using `<component />` rather than a render function. Might be worth trying to convert it over and see if all the tests still pass. Might just work

stackoverfloweth:
> > looking for a way to provide type context for router-view being that it's not a SFC and can't use `defineSlots`
> > https://stackoverflow.com/questions/78058105/vue-defineslots-functional-component
> 
> We could probably write router-view as a sfc using `<component />` rather than a render function. Might be worth trying to convert it over and see if all the tests still pass. Might just work

@pleek91 seems to have worked and IMO is much easier to understand

pleek91:
> @pleek91 seems to have worked and IMO is much easier to understand

Nice, sold.

src/components/routerLink.vue
stackoverfloweth: I thought this might be useful since our router-view supports `to='https://www.google.com'` unlike vue-router. Thinking of how prefect added the little icon, I thought this could be useful. Do you think we should just drop it?

pleek91: I think its useful. I just don't think checking for http is sufficient. Since there's no reason fully qualified url couldn't be passed that's same origin.

stackoverfloweth: oh right, but notice this is checking `resolved`, which if given a relative path will return a full qualified URL

[here's a line of a test](https://github.com/kitbagjs/router/blob/component-slot-props/src/components/routerLink.browser.spec.ts#L39) that should prove what I'm saying

pleek91: that doesn't seem right to me 🤔 But even if it is, that means resolved will be fully qualified and won't `resolved.value.startsWith('http')` always be true?

stackoverfloweth: oh duh. Fixed

pleek91 (approved):

PR #102, opened 2024-02-28
Add support for running route hooks in before location update hook

pleek91:
# Description
Updates navigation to provide context methods for the `onBeforeLocationUpdate` hook to allow for control over what happens next. Currently providing `done` and `abort` methods. `done` instructions navigation to finish updating the location but then to stop prior to running `onAfterLocationUpdate`. `abort` prevents the location from being updated at all. 

I haven't yet added support for `abort` in hooks with a timing of `before`. That will be in a future PR.

### DIFF
--- a/src/errors/navigationAbortError.ts
+++ b/src/errors/navigationAbortError.ts
@@ -1,0 +1,1 @@
+export class NavigationAbortError extends Error {}
--- a/src/utilities/createRouter.ts
+++ b/src/utilities/createRouter.ts
@@ -62,10 +62,11 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
 
     if (!to) {
       reject('NotFound')
+
       return false
     }
 
-    const success = await executeRouteHooks({
+    return await executeRouteHooks({
       hooks: [
         ...hooks.before,
         ...getRouteHooks(to, from, 'before'),
@@ -74,8 +75,6 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
       from,
       onRouteHookError,
     })
-
-    return success
   }
 
   const onAfterLocationUpdate: AfterLocationUpdate = async (url) => {
@@ -83,10 +82,14 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
     // const from = isRejectionRoute(route) ? null : route
 
     if (!to) {
-      return reject('NotFound')
+      reject('NotFound')
+      return
     }
 
-    // const success = await executeRouteHooks({
+    updateRoute(to)
+    clearRejection()
+
+    // await executeRouteHooks({
     //   hooks: [
     //     ...hooks.after,
     //     ...getRouteHooks(to, from, 'after'),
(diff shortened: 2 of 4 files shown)


src/utilities/createRouter.ts
```diff
@@ -83,9 +77,32 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+    //     ...getRouteHooks(to, from, 'after'),
+    //   ],
+    //   to,
+    //   from,
+    //   onRouteHookError,
+    // })
+
+    // if (!success) {
+    //   return
+    // }
```
pleek91, line 104: Commenting this out because `getRouteHooks` with `after` timing isn't implemented yet.

src/utilities/createRouter.ts
```diff
@@ -83,9 +77,32 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+      done()
+    }
+  }
+
+  const onAfterLocationUpdate: AfterLocationUpdate = async (url) => {
+    const to = getResolvedRouteForUrl(routerRoutes, url)
+    // const from = isRejectionRoute(route) ? null : route
+
+    if (!to) {
+      return reject('NotFound')
```
pleek91, line 89: Technically this isn't needed for anything other than type narrowing. Would it be better to throw an error here since we would expect a not found rejection to be caught in the `onBeforeLocationUpdate` callback? 

This kinda makes me want to add a third callback into navigation that runs first and handles not found routes and returns the resolved route which would then be passed into the before and after hook.

src/utilities/routerNavigation.ts
pleek91, line 1: I don't really like how verbose and repetitive the try catch is in three places here. But its also pretty simple and follows the same patter we've used for hooks with the push, replace, and reject errors.

src/errors/navigationAbortError.ts
pleek91, line 1: The thought crossed my mind to call these "signals" rather than "errors". Like `NavigationAbortSignal`. And anything that is a signal is something we're expecting to be caught and trigger additional functionality. Idk, what do you think?

src/utilities/createRouter.ts
stackoverfloweth: I don't hate the duplication here, feels clear when reading, even if this particular case is not possible. 

Regarding how possible it is, if I change the location in a before hook to something that's gonna be "not found", does that not found rejection happen in the before hook triggered by my location change?

stackoverfloweth (commented):
I think this is a good solution to a tricky problem. I'm not sure I love having `done` though, I think I would prefer we only introduce `abort` for now

src/utilities/createRouter.ts
stackoverfloweth: regarding the 3rd callback, I don't really have a good concept of what that would look like but I'm not opposed to it if you wanna do a poc

pleek91:
> I think this is a good solution to a tricky problem. I'm not sure I love having `done` though, I think I would prefer we only introduce `abort` for now

`done` is what allows reject, push, and replace in a hook to work. Like we talked this morning we need the location to actually be updated. 

`abort` is what would be used to prevent the location and the route from being updated at all. Canceling the current navigation. 

Without `done` this won't work right.

stackoverfloweth (approved):

PR #104, opened 2024-03-02
Refactor router history using history library

pleek91:
# Description
Updates the navigation implementation using the [history](https://github.com/remix-run/history) library. This means we can offer exact same functionality between node and the browser. We could also easily offer a hash navigation mode since the history library has that built in.

Ended up here kinda on accident. I was looking at how to improve the interactions between the router and navigation and kept getting hung up on the differences between the node and browser environment. And realizing anything I added would mean two implementations. 

I think this is the sort of thing that makes sense to offload to a library so we can focus on the router itself. Though I am hesitant to introduce a dependency.


src/utilities/routerNavigation.browser.spec.ts
```diff
@@ -16,59 +15,35 @@ test('when go is called, forwards call to window history', () => {
 })
 
 test('when back is called, forwards call to window history', () => {
-  vi.spyOn(window.history, 'back')
+  vi.spyOn(window.history, 'go')
```
pleek91, line 18: the history library calls `go` for its `back` and `forward` methods.

src/utilities/routerNavigation.browser.spec.ts
```diff
@@ -16,59 +15,35 @@ test('when go is called, forwards call to window history', () => {
...
-test('when update is called, calls updateBrowserUrl', () => {
-  vi.spyOn(utilities, 'updateBrowserUrl')
-
-  const onAfterLocationUpdate = vi.fn()
-  const url = random.number().toString()
-  const history = createRouterNavigation({ onAfterLocationUpdate })
-
-  history.update(url)
-
-  expect(utilities.updateBrowserUrl).toHaveBeenCalledWith(url, undefined)
```
pleek91, line 49: Not testing, because we're not responsible. But we could probably mock the history library and test that the right history method is called.

src/utilities/routerNavigation.browser.spec.ts
```diff
@@ -16,59 +15,35 @@ test('when go is called, forwards call to window history', () => {
...
-test('when update is called and not same origin does not call onAfterLocationUpdate ', () => {
-  const onAfterLocationUpdate = vi.fn()
-  vi.spyOn(utilities, 'isSameOrigin').mockReturnValue(true)
-
-  const url = random.number().toString()
-  const history = createRouterNavigation({ onAfterLocationUpdate })
-
-  history.update(url)
-
-  expect(onAfterLocationUpdate).not.toHaveBeenCalled()
```
pleek91, line 73: I feel like this test was just broken anyway. Because it was mocking the return value of `isSameOrigin` to `true` but the test says its testing when NOT same origin. 

Anyway from what I can tell the history library will automatically call `window.location.assign` if its not same origin. Will probably want to verify that

src/utilities/routerNavigation.spec.ts
```diff
@@ -2,21 +2,14 @@ import { expect, test, vi } from 'vitest'
...
 import { random } from '@/utilities/testHelpers'
 
-test('Browser like navigation is not supported', () => {
-  const onAfterLocationUpdate = vi.fn()
-  const navigation = createRouterNavigation({ onAfterLocationUpdate })
-
-  expect(() => navigation.back()).toThrowError()
-  expect(() => navigation.forward()).toThrowError()
-  expect(() => navigation.go(1)).toThrowError()
-})
```
pleek91, line 12: But now it is!

package.json
```diff
@@ -46,5 +46,8 @@
   },
   "peerDependencies": {
     "vue": "^3.4.5"
+  },
+  "dependencies": {
+    "history": "^5.3.0"
```
stackoverfloweth, line 51: crazy that this is our first "true" dependency!

src/utilities/createRouter.ts
```diff
@@ -113,6 +113,7 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
   const notFoundRoute = getRejectionRoute('NotFound')
   const { route, updateRoute } = createCurrentRoute(notFoundRoute)
 
+  // todo this isn't correct anymore because we need to run before hooks on the initialUrl
```
stackoverfloweth, line 116: I can pickup this action item, unless you already have started?

src/utilities/routerNavigation.browser.spec.ts
stackoverfloweth: yep, makes sense to me

src/utilities/routerNavigation.browser.spec.ts
stackoverfloweth: should we write a test for that? Probably more of an "ingetegration test"

src/utilities/routerNavigation.spec.ts
stackoverfloweth: are there even more .browser test suites that no longer NEED to be?

src/utilities/routerNavigation.ts
```diff
@@ -61,42 +62,28 @@ function createBrowserNavigation({ onBeforeLocationUpdate, onAfterLocationUpdate
       }
     }
 
-    await updateBrowserUrl(url, options)
+    action?.()
```
stackoverfloweth, line 65: should this be awaited? we will need to use `if(action)` rather than the safe navigation if we're going to await it

stackoverfloweth (commented):
I'm still good either way with this. I like effectively zero dependencies but I also think it's you're right that this is the kind of dependency we should be open to. Especially since it's getting like 8M weekly downloads 🤯

Just a couple small comments for you to consider

src/utilities/createRouter.ts
pleek91: I have not, go for it!

src/utilities/routerNavigation.ts
pleek91: its not async so I don't think there's a reason to await it

src/utilities/routerNavigation.spec.ts
pleek91: I went through the `.browser.spec.ts` files and cleaned up what could.

src/utilities/routerNavigation.browser.spec.ts
pleek91: possibly though I'm not 100% sure how to write that accurately in a emulated browser environment.

PR #107, opened 2024-03-09
Refactor navigation

pleek91:
# Description
Removes the concept of navigation. Replaces it with a simpler `history` concept that is a wrapper for the history npm package. Then the router itself has an `update` method which is the core logic of the router. The `push` and `replace` utilities were also removed in and brought into the router itself. They are basically just wrappers around `resolve` and `update` and don't need to be external utilities.


stackoverfloweth (approved):
😍

PR #108, opened 2024-03-12
Refactor of how hooks are run to account for different timing and outcomes

pleek91:
# Description
Prior to this there was a single utility for executing route hooks that would return a boolean based on if something like a `push`, `replace`, or `reject` happened. This wasn't enough information for the `update` method in `createRouter` to know what to do. 

I've added stronger types for `BeforeRouteHook` and `AfterRouteHook` and defined types for `BeforeRouteHookResponse` and `AfterRouteHookResponse`. Each response has a status like `abort`, `push`, `reject`, or `success`. The `abort` status is new altogether and is only included for before hooks.

### DIFF
--- a/src/utilities/createRouter.ts
+++ b/src/utilities/createRouter.ts
@@ -1,9 +1,7 @@
 import { App, readonly } from 'vue'
 import { RouterLink, RouterView } from '@/components'
 import { routerInjectionKey, routerRejectionKey } from '@/compositions'
-import { NavigationAbortError } from '@/errors/navigationAbortError'
-import { Routes, Router, RouterOptions, RouterImplementation } from '@/types'
-import { RouterPushError, RouterRejectionError } from '@/types/errors'
+import { Routes, Router, RouterOptions, RouterImplementation, RouterReject, RouteHookResponse } from '@/types'
 import { RouterPushImplementation } from '@/types/routerPush'
 import { RouterReplaceImplementation } from '@/types/routerReplace'
 import { createCurrentRoute } from '@/utilities/createCurrentRoute'
@@ -17,7 +15,7 @@ import { createRouterRoutes } from '@/utilities/createRouterRoutes'
 import { getInitialUrl } from '@/utilities/getInitialUrl'
 import { getResolvedRouteForUrl } from '@/utilities/getResolvedRouteForUrl'
 import { getRouteHooks } from '@/utilities/getRouteHooks'
-import { OnRouteHookError, executeRouteHooks } from '@/utilities/hooks'
+import { runAfterRouteHooks, runBeforeRouteHooks } from '@/utilities/hooks'
 
 type RouterUpdateOptions = {
   replace?: boolean,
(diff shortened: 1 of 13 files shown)


src/utilities/createRouterReject.ts
pleek91, line 1: this was a little side quest because I was worried about readability in the router `update` method. I read nicely to call `setRejection(type)` and `setRejection(null)` rather than `reject(type)` and `clearRejection`. And I like the syntax of just exporting `setRejection` from this utility.

src/utilities/hooks.ts
pleek91, line 1: I ended up separating out running `before` and `after` hooks. Basically just because of `abort` only existing in `before` hooks. More code but seems simpler and more straight forward.

src/utilities/createRouter.ts
```diff
@@ -38,67 +35,60 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
-      // })
-
-    } catch (error) {
-      if (error instanceof NavigationAbortError) {
+    const from = route
+
+    const beforeResponse = await runBeforeRouteHooks({ to, from, hooks })
+
+    const setHistory = (): void => history.update(url, { replace })
+    const setRoute = (): void => updateRoute(to)
```
stackoverfloweth, line 43: ~one thing I noticed with this change is now that we're not using an exception, the abort doesn't actually interrupt execution of other hooks. Not sure that really matters?~

never mind, looks like inside of hooks it still uses abort error, which WOULD in fact interrupt other hooks?

stackoverfloweth: I personally hate the syntax of a `const` function with the type defined inline like this. It's 1000% a nit, but I would have written it as 

```suggestion
    function setHistory(): void {
      history.update(url, { replace })
    }

    function setRoute(): void {
      updateRoute(to)
    }
```

Or possibly not even defined these, since calling the underlying functions is pretty straight forward already

stackoverfloweth (commented):
big fan of where you're going with these hooks 👏

src/utilities/createRouter.ts
```diff
@@ -38,67 +35,60 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+      // On success update history, set the route, and clear the rejection
+      case 'SUCCESS':
+        setHistory()
+        setRoute()
+        setRejection(null)
+        break
+
+      default:
+        const exhaustive: never = beforeResponse
+        throw new Error(`Switch is not exhaustive for before hook response status: ${JSON.stringify(exhaustive)}`)
```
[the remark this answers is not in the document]
pleek91: Oh that's clever. Gonna steal that

src/utilities/createRouter.ts
pleek91: Yeah I don't feel strongly here. I was mostly defining these so that the switch had zero logic. Just called `setThing() and setOtherThing()` reads nicely. 

Logic wise it's the same either way. 

> never mind, looks like inside of hooks it still uses abort error, which WOULD in fact interrupt other hooks?

Correct. We're still throwing and catching exceptions, that just is an implementation detail of running hooks and not something you know about inside or outside of hooks.

stackoverfloweth (approved):

PR #109, opened 2024-03-22
Refactor hooks registration

pleek91:
This build off of https://github.com/kitbagjs/router/pull/108 to fix the types and also implement how hooks are "registered" and then how we get the in order to run them.


src/utilities/createRouter.ts
```diff
@@ -135,10 +128,8 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
     back: history.back,
     go: history.go,
     install,
-    onBeforeRouteEnter,
-    onBeforeRouteLeave,
-    onBeforeRouteUpdate,
     initialized,
+    ...routeHookMethods,
```
pleek91, line 132: This feels a little code golfy. Could easily be convinced to not rest/spread these. (This is all the `onBeforeRoute...` and `onAfterRoute...` methods.

src/utilities/hooks.ts
```diff
@@ -1,27 +1,32 @@
...
-    ...hooks.before,
-    ...getBeforeRouteHooks(to, from),
+    ...global.onBeforeRouteEnter,
+    ...global.onBeforeRouteUpdate,
+    ...global.onBeforeRouteLeave,
+    ...route.onBeforeRouteEnter,
+    ...route.onBeforeRouteUpdate,
+    ...route.onBeforeRouteLeave,
+    ...component.onBeforeRouteUpdate,
+    ...component.onBeforeRouteLeave,
```
pleek91, line 29: I had a more complex order in mind something like
```
global before enter
route before enter
global before update
route before update
component before update
component before leave
route before leave
component before leave
```
But once I put them down it seemed to much. The order above seems much more intuitive and explainable. But I don't feel strongly.

pleek91 (commented):
@stackoverfloweth this turned out pretty good I think.

src/utilities/createRouter.ts
stackoverfloweth: yeah I probably wouldn't have, but I don't feel strongly

src/utilities/hooks.ts
stackoverfloweth: perhaps we should try extracting this logic so that we can change this easily in the future? looks like we have something similar in `runAfterRouteHooks`

stackoverfloweth: despite the way this looks, if a route changes, `global beforeRouteEnter` would be first but `route beforeRouteEnter` would be next right? not `global beforeRouteUpdate`

stackoverfloweth: also, are you still planning on unit tests to assert correct ordering?

stackoverfloweth (approved):
So nice! 🤩

src/utilities/hooks.ts
pleek91: I updated the order to be (before) enter -> update -> leave in  global -> route -> component order and then (after) leave -> update -> enter in component -> route -> global order.

src/utilities/createRouter.ts
pleek91: updated

pleek91:
@stackoverfloweth 

> perhaps we should try extracting this logic so that we can change this easily in the future? looks like we have something similar in runAfterRouteHooks

They are similar but different. I think this is easy to change as is. You just move a line up or down to change its order. So I wouldn't abstract this.

pleek91:
@stackoverfloweth 

> also, are you still planning on unit tests to assert correct ordering?

Yeah need to test order and actual execution. Maybe in the next PR

stackoverfloweth (approved):

PR #110, opened 2024-03-26
setup basic readme

stackoverfloweth:
follows pattern established by @kitbag/mapper and @kitbag/events

- really only covers what's in "getting-started" of vitepress site
- links to vitepress site at the top
- note both `npm` and `zip` badges will not work until we have any version of @kitbag/router published on npm

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,97 @@
-# Router
+# @kitbag/router
 
-Type safe router for Vue.js
+A simple and versatile mapping utility for Typescript.
+
+[![Npm Version](https://img.shields.io/npm/v/@kitbag/router.svg)](https://www.npmjs.org/package/kitbag/router)
+[![Zip Size](https://img.badgesize.io/https:/unpkg.com/@kitbag/router/dist/kitbag-router?compression=gzip)](https:/unpkg.com/@kitbag/router/dist/kitbag-router)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,3 +1,97 @@
...
+bun add @kitbag/router
+# yarn
+yarn add @kitbag/router
+# npm
+npm install @kitbag/router
+```
+
+## Define Basic Routes
+
+Create an array of possible routes. Learn more about [defining routes](/core-concepts/defining-routes).
```
pleek91, line 26: Does this link need to be fully qualified?

README.md
```diff
@@ -1,3 +1,97 @@
...
+  { name: 'home', path: '/', component: Home },
+  { name: 'path', path: '/about', component: About },
+] as const satisfies Routes 
+```
+
+## Plugin
+
+Create a router instance and pass it to the app as a plugin
+
+```js {2-3,6,9}
```
pleek91, line 45: Do these highlights work on GitHub?

pleek91 (approved):
Think some links need to be updated but otherwise this is great.

README.md
stackoverfloweth: yes. great catch

README.md
stackoverfloweth: no, they also didn't break anything 🤷 I removed to be safe

PR #111, opened 2024-03-30
Update some packages to latest versions

pleek91:
Notice vitepress released an official 1.0 which prompted this


pleek91:
uggghhhhhhh that excessively deep type error is back....

stackoverfloweth (approved):

PR #112, opened 2024-04-03
refactored params

stackoverfloweth:
Big Things
- param names are now expected to be unique, you can't have the same name for a param in path, query, or across path and query.
  - currently throws runtime error
- as a result, you should no longer see `Record<string, Param[]>` when looking at param definitions or `Record<string, unknown[]>` when looking at param implementations (where unknown is the actual value coming from end user)

Smaller Things
- Path and Query types retain the actual `Param` type they are given, only trading that in for `ReturnType` when building RouteMethods
  - this enables future merging of multiple paths into 1, since `MergeParams<TParentParams, TChildParams> extends PathParams<'${TParentPath}${TChildPath}'>` will now be true
- removed a handful of utilities and complexity around having possibly tuples and needing to normalize


src/types/params.ts
```diff
@@ -43,11 +43,11 @@ export type ExtractPathParamType<
   TParams extends Record<string, Param | undefined>
 > = TParam extends `?${infer OptionalParam}`
   ? OptionalParam extends keyof TParams
-    ? ExtractParamType<TParams[OptionalParam]> | undefined
-    : string | undefined
+    ? TParams[OptionalParam] | undefined
+    : StringConstructor | undefined
```
stackoverfloweth, line 47: note Path and Query params are expected to extend `Param` not the return value of the param

src/utilities/createRouterRoutes.ts
```diff
@@ -10,13 +10,17 @@ type ParentContext = {
...
+export function createRouterRoutes<T extends Routes>(routes: T, parentContext: ParentContext = {}): RouterRoute[] {
   const { parentPath = [], parentQuery = [], parentMatches = [], parentDepth = 0 } = { ...parentContext }
 
   return routes.reduce<RouterRoute[]>((value, route) => {
     const path = typeof route.path === 'string' ? createPath(route.path, {}) : route.path
     const query = typeof route.query === 'string' ? createQuery(route.query, {}) : route.query ?? { query: '', params: {} }
 
+    if (hasDuplicateParams(path, query)) {
+      throw new DuplicateParamsError()
+    }
```
stackoverfloweth, line 22: calling `path` and `query` will also throw this type of error if there are duplicates WITHIN the path or query. This line ensures that if path has `foo` and query also has `foo` we get an exception there too

src/utilities/params.ts
```diff
@@ -84,30 +84,30 @@ const numberParam: ParamGetSet<unknown> = {
   },
 }
 
-export function getParamValue<T extends Param>(value: string, param: T): ExtractParamType<T>
-export function getParamValue<T extends Param>(value: string, param: T): unknown {
+export function getParamValue<T extends Param>(value: string | undefined, param: T): ExtractParamType<T>
+export function getParamValue<T extends Param>(value: string | undefined, param: T): unknown {
```
stackoverfloweth, line 88: this was always technically true, but when I updated `getParamValuesFromUrl` which used to return string[] to be `getParamValueFromUrl` which now returns `string | undefined`, it correctly was upset with me

stackoverfloweth: it's safe to use `!` for value below because we expect the `undefined` to mean the param was optional, which is handled by optional ParamGetSet. If somehow `undefined` gets used with param type like `String` or any non-optional, those params will handle it gracefully by throwing `invalid()`

src/utilities/query.spec.ts
stackoverfloweth, line 1: same sort of tests in path.spec.ts, not sure why we never made a query version

src/utilities/params.ts
pleek91: This seems suspicious to me. `ParamGet` says the value must be a string. For example `stringParam` accepts and returns the value passed in. But it expects it to be a string, and it does not call `invalid` if its not a string.

src/utilities/createRouterRoutes.ts
```diff
@@ -51,6 +55,17 @@ export function createRouterRoutes(routes: Routes, parentContext: ParentContext
   }, [])
 }
 
-function reduceParams(entries: Path[] | Query[]): Record<string, Param[]> {
-  return entries.reduce((params, entry) => mergeMaybeTuples(params, entry.params as Record<string, Param | Param[]>), {})
+function hasDuplicateParams(path: Path, query: Query): boolean {
+  const keys = [...Object.keys(path.params), ...Object.keys(query.params)]
+
+  return new Set(keys).size !== keys.length
```
pleek91, line 61: Creating a `set` is crazy slow. How many times are we expecting this utility to get called when creating routes? Once per route?

src/utilities/createRouterRoutes.ts
pleek91: You mentioned path and query will throw errors. But I'm not finding that in this PR. Is that a future plan or what are you thinking here. Possible I just missed it but I looked lol.

src/utilities/createRouterRoutes.ts
```diff
@@ -10,13 +10,17 @@ type ParentContext = {
...
-export function createRouterRoutes(routes: Routes, parentContext: ParentContext = {}): RouterRoute[] {
+export function createRouterRoutes<T extends Routes>(routes: T, parentContext: ParentContext = {}): RouterRoute[] {
   const { parentPath = [], parentQuery = [], parentMatches = [], parentDepth = 0 } = { ...parentContext }
 
   return routes.reduce<RouterRoute[]>((value, route) => {
     const path = typeof route.path === 'string' ? createPath(route.path, {}) : route.path
     const query = typeof route.query === 'string' ? createQuery(route.query, {}) : route.query ?? { query: '', params: {} }
 
+    if (hasDuplicateParams(path, query)) {
+      throw new DuplicateParamsError()
```
pleek91, line 21: We probably want to error with the name of the param that was duplicated and where it was (in the path vs the query).

pleek91 (commented):
Overall I think this is a good refactor. I do think eventually we'll bring back duplicates, but I think this puts us in a good spot to tackle other more necessary complexities. And then only if there is demand for it we can figure out duplicates.

src/utilities/createRouterRoutes.ts
stackoverfloweth: here's an alternative that doesn't use `Set`. Do you prefer this?

```ts
  const pathKeys = Object.keys(path)
  const queryKeys = Object.keys(query)

  return pathKeys.some(key => queryKeys.includes(key))
  ```

Comment (pleek91) [reply src/utilities/createRouterRoutes.ts:61]: That would be significantly faster especially considering there are likely very few keys. I'd go with that version.

File: src/utilities/createRouterRoutes.ts (hunk 2 above)
Comment (stackoverfloweth) [reply src/utilities/createRouterRoutes.ts:22]: both `path` and `query` call `getParamsForString` to populate params. If you look in that utility, you'll see the following

```ts
const paramPattern = /:\??([\w]+)(?=\W|$)/g
const matches = Array.from(string.matchAll(paramPattern))

return matches.reduce<Record<string, Param>>((value, [match, paramName]) => {
  const isOptional = match.startsWith(':?')
  const param = getParam(params, paramName)

  if (paramName in value) {
    throw new DuplicateParamsError(paramName)
  }

  value[paramName] = isOptional ? optional(param) : param

  return value
```

src/utilities/params.ts
stackoverfloweth: I think you're right. I will revisit this

src/utilities/createRouterRoutes.ts
pleek91: ah gotcha 👍  ty

src/errors/duplicateParamsError.ts
```diff
@@ -1 +1,5 @@
-export class DuplicateParamsError extends Error {}
\ No newline at end of file
+export class DuplicateParamsError extends Error {
+  public constructor(paramName: string) {
+    super(`Invalid Param "${paramName}": Router can not have duplicate param names in path and query.`)
```
pleek91, line 3: For some reason github isn't letting me make a suggestion here. But slight tweak

```
`Invalid Param "${paramName}": Router does not support multiple params by the same name. All param names must be unique.`
```

src/utilities/params.ts
```diff
@@ -6,8 +6,20 @@ export function getParam<P extends Record<string, Param | undefined>>(params: P,
...
+  get: (value: string | undefined, extras: ParamExtras) => TValue,
+}
+
+function isOptionalParam(param: Param | OptionalParamGetSet<Param>): param is OptionalParamGetSet<Param> {
+  return optionalKey in param
+}
+
+export function optional<TParam extends Param>(param: TParam): OptionalParamGetSet<TParam> {
   return {
+    [optionalKey]: true,
```
pleek91, line 22: Dude, yes. I love this pattern

pleek91 (approved):
Nice work man

PR #114, opened 2024-04-11
update docs after ome recent refactors

stackoverfloweth:


docs/core-concepts/defining-routes.md
```diff
@@ -1,38 +1,31 @@
...
-/* type safe */
-export const routes = [
-  { name: 'home', path: '/', component: Home },
-  { name: 'path', path: '/about', component: About },
-] as const satisfies Routes 
+import { createRoutes } from '@kitbag/router'
 
-/* not type safe */
-export const routes: Routes = [
+export const routes = createRoutes([
```
pleek91, line 10:
```suggestion
const routes = createRoutes([
```

docs/core-concepts/route-params.md
```diff
@@ -136,27 +134,88 @@ const idFormatParam: ParamGetSet<IdFormat> = {
...
+
+## Param Name
+
+There are very few constraints on the name you choose for a param. Your param name can include any special character except forward slash `/`. Kitbag Router uses forward slash (or the end of the string) to denote the end of a param name in a path.
+
+```ts
+const routes = createRoutes([
+  {
+    name: 'users',
+    path: '/users/:can-have#what.ever!?',
```
pleek91, line 170: Im curious if that's even valid for a url. My guess is you can't use a # or a ? Because those mean the start of a hash or query.

docs/core-concepts/route-params.md
```diff
@@ -136,27 +134,88 @@ const idFormatParam: ParamGetSet<IdFormat> = {
...
+
+Keep in mind that any special characters in the param name will make accessing the value slightly less pretty.
+
+```ts
+const route = useRoute()
+
+route.params['can-have#what.ever!?']
+```
+
+If your param name starts with a question mark `?`, Router will assume your param was intended to be optional.
```
pleek91, line 184: I would reword this to just state that prefixing your param name with a ? will make it optional.

pleek91 (approved):

docs/core-concepts/route-params.md
stackoverfloweth: that's just the name of the param, not the value that will end up in a url

docs/core-concepts/route-params.md
stackoverfloweth: I tried 🤷 see what you think now

docs/core-concepts/route-params.md
pleek91: oh right 👍

pleek91 (approved):

PR #115, opened 2024-04-11
implemented useRoute method

stackoverfloweth:
side quest: combine 2 different approaches to grouping error types


src/types/register.ts
```diff
@@ -21,4 +23,7 @@ export type RegisteredRejectionType = Register extends { rejections: infer TReje
 
 export type RegisteredRouterState = Register extends { state: infer TState }
   ? TState
-  : {}
\ No newline at end of file
+  : {}
+
+export type RegisteredRouteMap = RoutesMap
```
pleek91, line 28: I think we should just get rid of the RoutesMap type and define RegisteredRoutesMap here in this file directly.

src/compositions/useRoute.ts
```diff
@@ -1,13 +1,15 @@
...
+import { RegisteredRouteMap, ResolvedRoute } from '@/types'
 
-type Route<T extends RouterRoute> = {
-  name: string,
-  params: ExtractRouterRouteParamTypes<T>,
-  query: unknown,
-  hash: string,
-}
+export function useRoute<TRouteName extends keyof RegisteredRouteMap>(routeName: TRouteName): ResolvedRoute<TRouteName>
+export function useRoute(routeName: string): ResolvedRoute {
```
pleek91, line 6: I think we should probably overload this and make `routeName` optional. If its not passed in the user just wants the route no matter what it is. Vs if they pass a name they want to to type guard.

src/compositions/useRoute.ts
```diff
@@ -1,13 +1,15 @@
...
+  const router = useRouter()
+  const route = router.find(routeName)
 
-export function useRoute<T extends RouterRoute>(): Identity<Route<T>> {
-  throw 'not implemented'
+  if (!route || route.name !== routeName) {
+    throw new UseRouteInvalidError(routeName, router.route.name)
+  }
```
pleek91, line 12: Few thoughts here

1. I don't think we actually need to look up the route for `routeName`. We can check against `route.matches`. 
2. We don't need to throw if the `routeName` is a parent of the current route. The user could use `useRoute('settings')` even if the current route is `settings.user`
3. Should this be reactive if the route changes? Like if the user asks for `settings` and when the component mounts that's true, but then the route changes and its no longer `settings`?

src/compositions/useRoute.ts
```diff
@@ -1,13 +1,46 @@
...
+    const routeNames = router.route.matches.map(route => route.name)
+    const routeAncestryNames = getRouteDotNotationNames(routeNames)
+    const routeNameIsValid = routeAncestryNames.includes(routeName)
+
+    if (!routeNameIsValid) {
+      throw new UseRouteInvalidError(routeName, router.route.name)
+    }
+  }
+
+  watch(router.route, () => checkRouteNameIsValid, { immediate: true, deep: true })
```
pleek91, line 26:
```suggestion
  watch(router.route, checkRouteNameIsValid, { immediate: true, deep: true })
```

pleek91: I'm not sure but we might want to call `checkRouteNameIsValid` manually first rather than using immediate true. Otherwise I think setup will finish. We'll have to test that

src/types/register.ts
```diff
@@ -21,4 +23,8 @@ export type RegisteredRejectionType = Register extends { rejections: infer TReje
 
 export type RegisteredRouterState = Register extends { state: infer TState }
   ? TState
-  : {}
\ No newline at end of file
+  : {}
+
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
```
pleek91, line 28:
```suggestion
// Because RegisteredRoutes defaults to `[]` it thinks passing it is unnecessary
// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
```

pleek91 (approved):

src/compositions/useRoute.ts
stackoverfloweth: I'll add a note for myself to explicitly test this when we start testing the router in a test app

PR #117, opened 2024-04-13
Remove implementation types

pleek91:
# Description
With the simpler and non recursive types provided by `createRoutes` we don't have the same problem we had before where typescript would try to compare the deeply nested types and break. So I think its safe to remove the implementation types.


src/types/hooks.ts
pleek91, line 1: Hooks got a little weird. Not sure I like this solution. The issue was because hooks run with the registered types but `createRouter` wants the types for push args to match its push parameters.

src/utilities/createRouterResolve.spec.ts
```diff
@@ -29,5 +29,6 @@ test('throws an error if route with params cannot be matched', () => {
   const routerRoutes = createRoutes(routes)
   const resolve = createRouterResolve(routerRoutes)
 
+  // @ts-expect-error
```
pleek91, line 32: This was a positive thing I ran into. This test is testing that resolve throws and error when invalid parameters are passed. With removing the implementation types, typescript also errored.

src/utilities/hooks.ts
pleek91, line 1: Wasn't really happy with the changes in this file. Felt weird, though its probably more accurate.

src/utilities/hooks.ts
```diff
@@ -1,146 +1,161 @@
...
-    await Promise.all(results)
-
-  } catch (error) {
-    if (error instanceof RouterPushError) {
-      return {
-        status: 'PUSH',
-        to: error.to,
-      }
-    }
+type RouteHookBeforeRunner<T extends RouterRoutes> = (context: BeforeContext) => Promise<BeforeRouteHookResponse<T>>
```
pleek91, line 22: I didn't like that it requires T to be passed because it cannot be inferred as it is currently written.

stackoverfloweth (approved):

src/types/routerReplace.ts
```diff
@@ -10,6 +10,5 @@ export type RouterReplace<
   TRoutePath extends string
 >(source: string | RouteWithParams<TRoutes, TRoutePath>, options?: RouterPushOptions) => Promise<void>
 
-export type RegisteredRouterReplace = RouterReplace<RegisteredRoutes>
-
-export type RouterReplaceImplementation = (source: string | RouteWithParamsImplementation, options?: RouterReplaceOptions) => Promise<void>
\ No newline at end of file
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
+export type RegisteredRouterReplace = RouterReplace<RegisteredRoutes>
```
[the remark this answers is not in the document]
pleek91: Yeah for sure. I didn't move anything in this PR

PR #116, opened 2024-04-13
implementing new push syntax

stackoverfloweth:
https://github.com/kitbagjs/router/assets/6098901/a6b43d33-8c6a-4b85-b986-e497dfebc565

PR #118, opened 2024-04-17
new-push-syntax take 2

stackoverfloweth:
replaces https://github.com/kitbagjs/router/pull/116


src/components/routerLink.vue
```diff
@@ -4,14 +4,17 @@
...
-  import { RegisteredRouteMap, RegisteredRouteWithParams } from '@/types/register'
+  import { RegisteredRouteMap, RegisteredRoutes } from '@/types/register'
   import { RouterPushOptions } from '@/types/routerPush'
+  import { RouteParamsByName } from '@/types/routeWithParams'
+  import { Url, isUrl } from '@/types/url'
 
   const props = defineProps<{
-    to: string | RegisteredRouteWithParams<T>,
+    to: T,
+    params?: T extends keyof RegisteredRouteMap ? RouteParamsByName<RegisteredRoutes, T> : undefined,
```
pleek91, line 17: params are always optional? That's probably not good. But we're going to struggle here because of vue's type support for props...

src/types/register.ts
```diff
@@ -11,7 +12,7 @@ export interface Register {
 
 export type RegisteredRouter = Register extends { router: infer TRouter }
   ? TRouter
-  : Router
+  : Router<any>
```
pleek91, line 15: Curious why this any is needed

src/types/index.ts
```diff
@@ -6,9 +6,14 @@ export * from './register'
 export * from './resolved'
 export * from './resolvedQuery'
 export * from './routeMatchRule'
+export * from './routeWithParams.spec-d'
 export * from './routeWithParams'
 export * from './router'
+export * from './routerPush'
+export * from './routerReplace'
+export * from './routerRoute.spec-d'
```
pleek91, line 14:
```suggestion
export * from './routeWithParams'
export * from './router'
export * from './routerPush'
export * from './routerReplace'
```

src/types/routerPush.ts
```diff
@@ -8,9 +9,8 @@ export type RouterPushOptions = {
...
-> = <
-  TRoutePath extends string
->(source: string | RouteWithParams<TRoutes, TRoutePath>, options?: RouterPushOptions) => Promise<void>
-
-// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
-export type RegisteredRouterPush = RouterPush<RegisteredRoutes>
\ No newline at end of file
+> = {
+  <TSource extends RouteKeysThatHaveOptionalParams<TRoutes>>(source: TSource, params?: RouteParamsByName<TRoutes, TSource>, options?: RouterPushOptions): Promise<void>,
+  <TSource extends RouteKeysThatHaveRequireParams<TRoutes>>(source: TSource, params: RouteParamsByName<TRoutes, TSource>, options?: RouterPushOptions): Promise<void>,
```
pleek91, line 14: Does this still show all the route keys in the type hint even though they are separated into two different overloads?

src/types/routerPush.ts
pleek91, line 1: For these types that accept RouterRoutes and then convert them into a RouteMap, could we just pass the route map directly?

src/types/routerReplace.ts
```diff
@@ -1,14 +1,14 @@
...
->(source: string | RouteWithParams<TRoutes, TRoutePath>, options?: RouterPushOptions) => Promise<void>
-
-// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
-export type RegisteredRouterReplace = RouterReplace<RegisteredRoutes>
\ No newline at end of file
+> = {
+  <TSource extends RouteKeysThatHaveOptionalParams<TRoutes>>(source: TSource, params?: RouteParamsByName<TRoutes, TSource>, options?: RouterPushOptions): Promise<void>,
+  <TSource extends RouteKeysThatHaveRequireParams<TRoutes>>(source: TSource, params: RouteParamsByName<TRoutes, TSource>, options?: RouterPushOptions): Promise<void>,
+  (source: Url, options?: RouterPushOptions): Promise<void>,
+}
```
pleek91, line 14:
```suggestion
> = {
  <TSource extends RouteKeysThatHaveOptionalParams<TRoutes>>(source: TSource, params?: RouteParamsByName<TRoutes, TSource>, options?: RouterReplaceOptions): Promise<void>,
  <TSource extends RouteKeysThatHaveRequireParams<TRoutes>>(source: TSource, params: RouteParamsByName<TRoutes, TSource>, options?: RouterReplaceOptions): Promise<void>,
  (source: Url, options?: RouterReplaceOptions): Promise<void>,
}
```

src/types/utilities.ts
```diff
@@ -13,24 +13,6 @@ export type IsAny<T> = 0 extends (1 & T) ? true : false
 
 export type IsEmptyObject<T> = T extends Record<string, never> ? (keyof T extends never ? true : false) : false
 
-// Copied from [type-fest](https://github.com/sindresorhus/type-fest/blob/main/source/union-to-intersection.d.ts)
```
pleek91, line 16: happy to be removing this type

src/utilities/createRouter.ts
```diff
@@ -90,21 +94,65 @@ export function createRouter<const T extends RouterRoutes>(routes: T, options: R
...
+  function push(
+    source: any,
+    paramsOrOptions?: any,
+    maybeOptions?: RouterPushOptions,
+  ): Promise<void> {
+    const options: RouterPushOptions = (isUrl(source) ? paramsOrOptions : maybeOptions) ?? {}
+    const url = resolve(source, paramsOrOptions, maybeOptions)
+
+    return update(url, { replace: options.replace })
   }
```
pleek91, line 120: I believe since you've defined the types with the overloads, if you use the RouterPush type here but add the implementation types same as you have here it'll work without having to redefine the overloads
```
  const push: RouterPush<T> = (source: any, paramsOrOptions?: any, maybeOptions?: RouterPushOptions): Promise<void> => {
    const options: RouterPushOptions = (isUrl(source) ? paramsOrOptions : maybeOptions) ?? {}
    const url = resolve(source, paramsOrOptions, maybeOptions)

    return update(url, { replace: options.replace })
  }
```

src/utilities/createRouter.ts
```diff
@@ -90,21 +94,65 @@ export function createRouter<const T extends RouterRoutes>(routes: T, options: R
...
+    maybeOptions?: RouterReplaceOptions,
+  ): Promise<void> {
+    if (isUrl(source)) {
+      const options: RouterPushOptions = paramsOrOptions ?? {}
+      return push(source, { ...options, replace: true })
+    }
+
+    const params: any = paramsOrOptions
+    const options: RouterPushOptions = maybeOptions ?? {}
+    return push(source, params, { ...options, replace: true })
```
pleek91, line 148: Replace could be defined the same way. Here you have more specific implementation types, but if we're consistent with `push` it seems to work from my testing
```
   const replace: RouterReplace<T> = (source: any, paramsOrOptions?: any, maybeOptions?: RouterReplaceOptions): Promise<void> => {
    if (isUrl(source)) {
      const options = paramsOrOptions ?? {}
      return push(source, { ...options, replace: true })
    }

    const params = paramsOrOptions
    const options = maybeOptions ?? {}

    return push(source, params, { ...options, replace: true })
  }
```

src/utilities/createRouterFind.ts
```diff
@@ -1,24 +1,40 @@
...
+> = {
+  <TSource extends RouteKeysThatHaveRequireParams<TRoutes>>(source: TSource, params: RouteParamsByName<TRoutes, TSource>): ResolvedRoute | undefined,
+  <TSource extends RouteKeysThatHaveOptionalParams<TRoutes>>(source: TSource, params?: RouteParamsByName<TRoutes, TSource>): ResolvedRoute | undefined,
+  (source: Url): ResolvedRoute | undefined,
 }
 
-export function createRouterFind<const T extends RouterRoutes>({ routes, resolve }: CreateRouterFindContext<T>): RouterFind<T> {
-  return (source) => {
-    const url = resolve(source)
+export function createRouterFind<const TRoutes extends RouterRoutes>(routes: TRoutes): RouterFind<TRoutes> {
```
pleek91, line 17: Same thought here (and for all the rest of the methods declared like this).

src/components/routerLink.browser.spec.ts
pleek91, line 1: Was something broken about this test that you fixed?

pleek91 (commented):
This is really good. Some thoughts but nothing major. 

One thing I noticed while testing is if you select a route from the typehint in router.push you immediately get a type error because it thinks you're using the `url` overload but you didn't pass in a url. Hopefully we can do something about that.

src/components/routerLink.vue
stackoverfloweth: ah yeah, feels like a lose-lose situation. you'll get a param invalid error if you don't pass in the params required, maybe we can do that check inside of router-link and provide a more helpful error?

src/types/register.ts
stackoverfloweth: internally we use `useRouter()` in RouterLink which will never have Registered router and `Router` without the `<any>` has type never for `router.push` source.

image

my thinking is, if a developer doesn't provide register interface, the router they get back from `useRouter` should probably be pretty wide?

src/types/index.ts
stackoverfloweth: thank you. been meaning to update my barrel generator extension to exclude `.spec-d`, just did that now

src/types/routerPush.ts
stackoverfloweth: yeah it works pretty well, I'll see if there's a way that provides an even better syntax

https://github.com/kitbagjs/router/assets/6098901/dcfaca6a-b3c0-4833-a48c-e7a41c08d406

src/types/routerPush.ts
stackoverfloweth: not sure what we gain by making the implementer to the RouteMap creation. The places I can think of like RouterFind and RouterResolve are creating by Router, which has `TRoutes` but doesn't currently have any use for the `RouterMap` type

src/utilities/createRouter.ts
stackoverfloweth: ah yeah, good call. I was hoping to actually have the types inside of implementation but ended up needing to use any to get around type complaints

src/types/routerPush.ts
stackoverfloweth: this type seems to be a bit better dx?

https://github.com/kitbagjs/router/assets/6098901/e9f4d0c9-7844-49b4-99f6-81969802a8e9

image

src/utilities/createRouterFind.ts
stackoverfloweth: for RouterResolve and RouterReplace these functions actually have the correct types, not sure why those work well but find/replace inside of createRouter require a bit more coercing

src/components/routerLink.browser.spec.ts
stackoverfloweth: yeah, the syntax for `to` here was the old `{ route: string, params: {} }` syntax and the expected keys (if not Url) must be in RegisteredMap keys. But on investigating this helped me see how my `Router<any>` was slightly incomplete because RegisteredRoutes needed something similar

src/types/routerPush.ts
pleek91: yeah this looks great

pleek91 (approved):

PR #119, opened 2024-04-20
new syntax for router-link

stackoverfloweth:


src/types/register.ts
```diff
@@ -16,7 +18,7 @@ export type RegisteredRouter = Register extends { router: infer TRouter }
 
 export type RegisteredRoutes = Register extends { router: Router<infer Routes extends RouterRoutes> }
   ? Routes
-  : RouterRoute<string, any, any, false>[]
+  : RouterRoute<string, Path<'', {}>, Query<'', {}>, false>[]
```
pleek91, line 21: Interesting, I imagined this would end up being like `Path<string, {}>` so that any path was allowed.

pleek91: Does this not matter because we only use the key generic which is `string`? Are the path and query used for anything?

pleek91 (approved):

src/types/register.ts
stackoverfloweth: This only matters for registered route params when register interface is not used (like in our unit tests for router-link). Unfortunately it's still imperfect. As is it expects no params now.

PR #121, opened 2024-04-20
renamed Route to RouteProps

stackoverfloweth:
part 1 of bigger rename


pleek91 (approved):
Another name that came to mind is RouteDefinition. But ultimately I don't think it matters much here.

PR #127, opened 2024-04-21
RouterView as default route component

stackoverfloweth:
previously, if a child route rendered RouterView it would cause infinite loop. This PR moves responsibility for defaulting component to `createRoutes` method. 

side quests
- fixes vue external dep on vite build
- fixed bug with incorrect sort for matches, should be greatest ancestor -> least descendent
- attempts to fix type error when using plugin (could have been related to npm link)
- bumps version back to what's published on npm
- moves vue to peerDep

### DIFF
--- a/src/components/routerView.vue
+++ b/src/components/routerView.vue
@@ -10,7 +10,7 @@
 </template>
 
 <script lang="ts" setup>
-  import { AsyncComponentLoader, DeepReadonly, computed, defineAsyncComponent, provide, resolveComponent } from 'vue'
+  import { AsyncComponentLoader, DeepReadonly, computed, defineAsyncComponent, provide } from 'vue'
   import { useRejection } from '@/compositions/useRejection'
   import { useRouter } from '@/compositions/useRouter'
   import { useRouterDepth } from '@/compositions/useRouterDepth'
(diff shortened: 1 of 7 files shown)


src/components/routerView.vue
```diff
@@ -31,16 +30,15 @@
...
     if (routeComponent) {
       if (typeof routeComponent === 'function') {
         return defineAsyncComponent(routeComponent as AsyncComponentLoader)
       }
-
       return routeComponent
     }
 
-    return routerView
+    return 'span'
```
pleek91, line 42: Why a span?

src/types/router.ts
```diff
@@ -29,12 +29,11 @@ export type Router<
...
   onBeforeRouteEnter: AddBeforeRouteHook,
   onBeforeRouteLeave: AddBeforeRouteHook,
   onBeforeRouteUpdate: AddBeforeRouteHook,
   onAfterRouteEnter: AddAfterRouteHook,
   onAfterRouteLeave: AddAfterRouteHook,
   onAfterRouteUpdate: AddAfterRouteHook,
   initialized: Promise<void>,
-}
\ No newline at end of file
+} & Plugin
```
pleek91, line 39: Nit: could this go first? plugin & {...}

pleek91 (approved):

src/components/routerView.vue
stackoverfloweth: what should it be? alternatively I could have component return `undefined` and add a `v-if` in the template?

pleek91: Oh this is the noop? I think you can just turn null. Then it won't render anything

PR #130, opened 2024-04-21
Remove compositions and update readonly route type

pleek91:
# Description
Step one in making params writable. Updating params does not update the route but the tyypes and implementation allow for updating individual params (while still keeping the rest of the route object readonly)


src/types/resolved.ts
```diff
@@ -8,4 +9,6 @@ export type ResolvedRoute<TRoute extends Route = Route> = {
   key: TRoute['key'],
   query: ResolvedRouteQuery,
   params: ExtractRouteParamTypes<TRoute>,
-}
\ No newline at end of file
+}
+
+export type ReadonlyResolvedRoute<TRoute extends ResolvedRoute = ResolvedRoute> = DeepReadonly<Omit<TRoute, 'params'>> & Readonly<{ params: TRoute['params'] }>
```
pleek91, line 14: I'm not sold on 
1. Introducing another route type
2. The name.

stackoverfloweth: Should ResolvedRoute be read only by default?

stackoverfloweth (approved):

src/types/resolved.ts
pleek91: Possibly? Once we resolve a route do we ever need to update it internally?

stackoverfloweth: I think that's fine, I would proceed with that assumption and we can pivot in the future if we need to

stackoverfloweth (approved):

PR #131, opened 2024-04-21
using listener from history

stackoverfloweth:
documentation for history 
https://github.com/remix-run/history/blob/dev/docs/getting-started.md


src/utilities/createRouter.ts
```diff
@@ -41,38 +41,35 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
-        setHistory()
-        setRoute()
+        history.update(url, { replace })
         setRejection(beforeResponse.type)
         break
 
       // On success update history, set the route, and clear the rejection
       case 'SUCCESS':
-        setHistory()
-        setRoute()
```
pleek91, line 68: Guessing this is why tests are failing. You'll still need to set the route on reject and success. 

The setHistory and setRoute methods are here because the actions don't care about the logic of how to do these these. Just that it gets done. These actions should also always do the same thing. By removing these we can end up with subtle differences in functionality where there is not supposed to be. So I'd rather not remove them unless you have a strong reason to do so.

src/utilities/createRouter.ts
```diff
@@ -41,38 +41,35 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+        history.update(url, { replace })
         setRejection(null)
         break
 
       default:
         throw new Error(`Switch is not exhaustive for before hook response status: ${JSON.stringify(beforeResponse satisfies never)}`)
     }
+  }
 
+  async function afterUpdate(to: ResolvedRoute, from: ResolvedRoute): Promise<void> {
```
pleek91, line 72: What is your thought process behind separating update and afterUpdate? Updating is a single action which includes running before and after hooks.

src/utilities/createRouter.ts
```diff
@@ -133,6 +130,15 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
   const initialUrl = getInitialUrl(options.initialUrl)
   const initialized = update(initialUrl, { replace: true })
 
+  history.listen(({ location }) => {
+    const url = `${location.pathname}${location.search}${location.hash}`
```
pleek91, line 134: is there not a `toString` that does this already?

src/components/routerView.browser.spec.ts
```diff
@@ -67,7 +67,7 @@ test('renders components for initial route', async () => {
   expect(app.html()).toBe('Child')
 })
 
-test('updates components when route changes', async () => {
+test.only('updates components when route changes', async () => {
```
pleek91, line 70:
```suggestion
test('updates components when route changes', async () => {
```
Assuming that was for development

src/utilities/createRouter.ts
```diff
@@ -23,7 +24,15 @@ type RouterUpdateOptions = {
...
   const resolve = createRouterResolve(routes)
-  const history = createRouterHistory({ mode: options.historyMode })
+  const history = createRouterHistory({
+    mode: options.historyMode,
+    listener: () => {
+      const url = createPath(location)
+
+      update(url)
+    },
+  })
```
pleek91, line 34: Well I love this

src/utilities/createRouter.ts
```diff
@@ -36,37 +45,36 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
       // On abort do nothing
       case 'ABORT':
         return
 
       // On push update the history, and push new route, and return
       case 'PUSH':
-        setHistory()
-        push(...beforeResponse.to)
+        history.update(url, { replace })
+        await push(...beforeResponse.to)
```
pleek91, line 63: This makes sense but explain why cause I'm not sure it makes a difference

src/utilities/createRouter.ts
```diff
@@ -130,6 +140,8 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
   const notFoundRoute = getRejectionRoute('NotFound')
   const { route, updateRoute } = createCurrentRoute(notFoundRoute)
 
+  history.startListening()
```
pleek91, line 143: Should listening just be the default starting state?

src/utilities/createRouterHistory.ts
```diff
@@ -42,12 +39,24 @@ export function createRouterHistory({ mode }: RouterHistoryOptions = {}): Router
     return history.replace(url)
   }
 
+  let removeListener: (() => void) | undefined
```
pleek91, line 42: Blasphemy

pleek91 (approved):
This looks great to me. Assuming it works I love the direction

src/utilities/createRouter.ts
stackoverfloweth: The await? I just wanna make sure we don't end up in a thread that starts listening, then we come back to this thread and it was left on.

src/utilities/createRouter.ts
stackoverfloweth: I went back and forth on that. I'm happy to move it in. I had the thought that we might for some reason want to delay starting until 'install' or something but no good reason to.

src/utilities/createRouterHistory.ts
stackoverfloweth: Agreed 😖

stackoverfloweth:
> This looks great to me. Assuming it works I love the direction

Glad you like it. It does work in both unit tests and in my demo app. I'm glad we collaborated on this, it feels right to me as well.

src/utilities/createRouter.ts
pleek91: I don't feel strongly. I'm just in favor of keeping createRouter as simple as possible. 

I do kinda like putting it in the install 🤔

pleek91 (approved):

PR #132, opened 2024-04-22
Fixed bug with history start/stop listener

stackoverfloweth:


src/utilities/createRouterHistory.ts
```diff
@@ -42,7 +42,8 @@ export function createRouterHistory({ mode, listener }: RouterHistoryOptions = {
   let removeListener: (() => void) | undefined
 
   const startListening: () => void = () => {
-    if (!!listener && removeListener === undefined) {
+    if (listener) {
+      removeListener?.()
```
pleek91, line 46: IMO listener should be required and then we don't even need this if statement.

pleek91 (approved):

src/utilities/createRouterHistory.ts
stackoverfloweth: listener comes from an options arg for history, I _really_ like options to be as optional as possible

pleek91: For user facing stuff for sure. But router literally doesn't work if you don't provide a listener here. It's not optional functionality wise

PR #135, opened 2024-04-22
Writable route params

pleek91:
# Description
Makes route.params writable so that it updates the route when params are updated.


src/types/resolved.ts
pleek91, line 1: I pivoted here and went this direction
1. `ResolvedRoute` is a completely readonly value
2. `ResolvedRoute` is a route populated with actual values. But its not necessarily the "current" route. For example in hooks both `to` and `from` are `ResolvedRoute`. Readonly values you cannot modify. They also don't specifically represent the current route (in a before hook `from` is the current route, in an after hook `to` is the current route). 
3. `RouterRoute` is the current route that you get from `router.route` and `useRoute`. It has writable params and is designed for the user to interact with. I'm imagining that type being more rich than `ResolvedRoute`.

src/utilities/createRouter.ts
```diff
@@ -154,8 +153,23 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
...
+          set(target, prop, value) {
+            router.push(route.key, { ...target, [prop]: value })
+            return true
+          },
+        })
+      }
+
+      return Reflect.get(target, prop, receiver)
+    },
+  })
```
pleek91, line 169: WIP

Notes: 
- Setting a param calls `router.push` and does not update the current route directly. This means the url is the source of truth. But it could be problematic if the route has async before hooks. Because the page won't react until after the route has been updated. But that would only happen in an onBeforeRouteUpdated hook. Which seems rare. 
- I'm using `router.push` rather than just `push` because otherwise we cannot spy on router.push in unit tests. 
- I'm thinking we might introduce a router setting for the push functionality here. I feel like everyone has different opinions on whether this should be a push or a replace. And we can easily allow users to customize this behavior.

pleek91: Thought: This probably breaks if you update multiple params at the same time. Line 161 will overwrite the previous updates...

src/utilities/createRouter.spec.ts
```diff
@@ -86,6 +92,23 @@ test('route is readonly except for individual params', async () => {
...
   route.params.param = 'goodbye'
+
+  expect(spy).toHaveBeenCalledOnce()
+
+  await flushPromises()
+
   // @ts-expect-error
   expect(route.params.param).toBe('goodbye')
+
+  // @ts-expect-error
```
stackoverfloweth, line 103: is this expect-error here because we actually expect an error or because types weren't setup correctly yet?

src/utilities/createRouter.spec.ts
```diff
@@ -86,6 +92,23 @@ test('route is readonly except for individual params', async () => {
...
+  const { param } = toRefs(route.params)
+
+  param.value = 'again'
+
+  expect(spy).toHaveBeenCalledTimes(2)
+
+  await flushPromises()
+
+  // @ts-expect-error
```
stackoverfloweth, line 112: same question here

src/utilities/createRouter.spec.ts
pleek91: Right it's because params is type {} which means it has no properties so you cannot access anything. The two you flagged should go away once route.params is Record<string, unknown>

stackoverfloweth (approved):

PR #136, opened 2024-04-23
setup default rejection component for custom rejections

stackoverfloweth:
I noticed that when a dev sets custom rejection type the options correctly will require a component for router options arg if passed but the whole options arg can be omited. That puts the dev in a position where rejections can happen but the user ends up at a blank screen. 

This PR pivots from having a default NotFound component to a generic rejection component that supports both NotFound as well as any custom types the dev may define. Overriding this default component still works the same way, but components are always optional now that there is a default.

https://github.com/kitbagjs/router/assets/6098901/b183c58c-a441-47d2-a6f4-ce1bd14569a9

https://github.com/kitbagjs/router/assets/6098901/1c9188fc-5dcd-417e-b978-3c113c57fe62


pleek91 (approved):
Great idea

PR #139, opened 2024-04-24
updated createCurrentRoute to export both ResolvedRoute and RouterRoute

stackoverfloweth:


src/utilities/createCurrentRoute.ts
```diff
@@ -1,22 +1,24 @@
 import { readonly, reactive } from 'vue'
-import { ResolvedRoute } from '@/types/resolved'
+import { ResolvedRoute, RouterRoute } from '@/types/resolved'
 
 type ResolvedRouteUpdate = (route: ResolvedRoute) => void
 
-type CurrentRoute = {
-  route: ResolvedRoute,
+type CurrentRouteContext = {
+  resolvedRoute: ResolvedRoute,
```
pleek91, line 7: I'd probably call this currentRoute or something other than resolvedRoute. Otherwise we're going to have problems talking about which "resolved route"

pleek91 (approved):

PR #141, opened 2024-04-26
added router.update method

stackoverfloweth:
likely to be the core of `route.update`, can enable editable params as well.

side-quest
- I noticed everywhere we used `Router` without generic we used `<any>`. I moved that into default
- renamed private `update` method in `createRouter` to `set`, since `update` means something new


src/types/routerUpdate.ts
```diff
@@ -0,0 +1,15 @@
+import { ResolvedRoute } from '@/types/resolved'
+
+export type RouteUpdateOptions = {
+  replace?: boolean,
+}
```
pleek91, line 5: Should we add query to this?

pleek91 (approved):
Nice, this looks great. Needs test coverage but we can add that later

PR #142, opened 2024-04-26
added update method to route itself

stackoverfloweth:


src/utilities/createRouterRoute.ts
```diff
@@ -1,6 +1,32 @@
 import { readonly } from 'vue'
-import { ResolvedRoute, RouterRoute } from '@/types'
+import { ResolvedRoute } from '@/types/resolved'
+import { RouteUpdate, RouteUpdateOptions, RouterUpdate } from '@/types/routerUpdate'
 
-export function createRouterRoute(route: ResolvedRoute): RouterRoute {
-  return readonly(route)
+export type RouterRoute<TRoute extends ResolvedRoute = ResolvedRoute> = Omit<ResolvedRoute, 'params'> & Readonly<{
+  params: { -readonly [P in keyof TRoute['params']]: TRoute['params'][P] },
```
pleek91, line 6: Nit: we should abstract this into a type utility called `Writable` or something similar.

src/utilities/createRouterRoute.ts
```diff
@@ -1,6 +1,32 @@
...
+  function routeUpdate(keyOrParams: PropertyKey | Partial<ResolvedRoute['params']>, valueOrOptions: any, maybeOptions?: RouteUpdateOptions): Promise<void> {
+    if (typeof keyOrParams === 'object') {
+      return update(route, keyOrParams, valueOrOptions)
+    }
+
+    const updatedParams = readonly({
+      ...route.params,
+      [keyOrParams]: valueOrOptions,
+    })
+    return update(route, updatedParams, maybeOptions)
```
pleek91, line 20: I'm guessing the types for update need the writable type so that here we don't have to first pass the params into readonly.

src/utilities/createRouterRoute.ts
```diff
@@ -1,6 +1,32 @@
...
+  }
+
+  return new Proxy(route as RouterRoute<TRoute>, {
```
pleek91, line 23: I think Proxy takes a generic. Does this work?
```suggestion
  return new Proxy<RouterRoute<TRoute>>(route, {
```

pleek91 (approved):
Nicely done 👏

src/utilities/createRouterRoute.ts
stackoverfloweth: yep, good catch

src/utilities/createRouterRoute.ts
stackoverfloweth: I tried that but similar to reduce, the starting value is expected to match the generic, which isn't true here

image

pleek91 (approved):

PR #152, opened 2024-05-02
Route meta documented

stackoverfloweth:
- normalize `matched.meta` to always be empty object so devs don't have to check for undefined

```ts
// without
route.metached.meta?.thing

// with
route.matched.meta.thing
```

- update types for RouteMeta to be `Record<string, unknown>` so regardless of if devs override interface all string keys will be accepted and value defaults to unknown
- docs updated

### DIFF
--- a/docs/advanced-concepts/route-meta.md
+++ b/docs/advanced-concepts/route-meta.md
@@ -1,1 +1,42 @@
 # Route Meta
+
+It may be useful to store additional context on your route to be used in a hook, or within a component. Meta data might be useful for authorization, analytics, and much more. For example, below we will use route meta to configure the document title per route.
+
+```ts
+import { createRoutes, createRouter } from '@kitbag/router'
+
+const routes = createRoutes([
+  { 
+    name: 'home',
+    path: '/',
+    component: Home,
+    meta: {
+      pageTitle: 'Kitbag Home'
+    }
+  },
+  { 
+    name: 'path',
+    path: '/about',
+    component: About,
+    meta: {
+      pageTitle: 'Learn More About Kitbag'
+    }
+  },
+])
+
+const router = createRouter(routes)
+
+router.onAfterRouteEnter(to => {
+  document.title = to.matched.meta.pageTitle
+})
+```
+
+You might notice that the default type for `pageTitle` is `unknown`. By default meta is `Record<string, unknown>` but can be narrowed with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).
+
+```ts
+declare module '@kitbag/router' {
+  interface RouteMeta {
+    pageTitle?: string
+  }
+}
+```
--- a/src/types/route.ts
+++ b/src/types/route.ts
@@ -1,8 +1,10 @@
 import { ExtractParamTypes, MergeParams, Param } from '@/types/params'
-import { RouteProps } from '@/types/routeProps'
+import { RouteMeta, RouteProps } from '@/types/routeProps'
 import { Path, Query, ToPath, ToQuery } from '@/utilities'
 
 export type Routes = Readonly<Route[]>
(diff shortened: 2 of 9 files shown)


docs/advanced-concepts/route-meta.md
```diff
@@ -1 +1,42 @@
...
+})
+```
+
+You might notice that the default type for `pageTitle` is `unknown`. By default meta is `Record<string, unknown>` but can be narrowed with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).
+
+```ts
+declare module '@kitbag/router' {
+  interface RouteMeta {
+    pageTitle?: string
+  }
```
pleek91, line 40: In my head I thought this was part of the Register interface for some reason. Prob makes sense that it's not since it's part of that interface since meta is part of routes which is used to create the router.

src/types/routeProps.ts
```diff
@@ -7,9 +7,7 @@ import { Query } from '@/utilities/query'
 
 export type RouteComponent = Component | DefineComponent | AsyncComponentLoader
 
-export interface RouteMeta {
-
-}
+export interface RouteMeta extends Record<string, unknown> {}
```
pleek91, line 10: Interesting. I'm surprised this works

src/utilities/createRoutes.spec.ts
```diff
@@ -65,4 +65,16 @@ test('given route without component, sets component default to RouterView', () =
   ])
 
   expect(route.matched.component).toMatchObject(RouterView)
+})
+
+test('given route without meta, sets meta equal to empty object', () => {
```
pleek91, line 70: Nice yeah good idea. Much better devx

src/types/route.ts
```diff
@@ -1,17 +1,19 @@
 import { ExtractParamTypes, MergeParams, Param } from '@/types/params'
-import { RouteProps } from '@/types/routeProps'
+import { RouteMeta, RouteProps } from '@/types/routeProps'
 import { Path, Query, ToPath, ToQuery } from '@/utilities'
 
 export type Routes = Readonly<Route[]>
 
+type RoutePropsWithMeta = RouteProps & { meta: RouteMeta }
```
pleek91, line 7: Why isn't meta just part of RouteProps?

src/types/routeProps.ts
stackoverfloweth: I did test it lol

docs/advanced-concepts/route-meta.md
stackoverfloweth: same. I considered combining it but ultimately decided not to. Do you feel strongly?

src/types/route.ts
stackoverfloweth: it's optional in `RouteProps` and should be

docs/advanced-concepts/route-meta.md
pleek91: I'm honestly not even sure it would work. If it does, it kinda be nice to be consistent and have everything go through Register. But I don't feel strongly either way. The devx and outcome is about the same

src/types/routeProps.ts
pleek91: Yeah I assumed it worked. Just wouldn't have thought that

src/types/route.ts
pleek91: Oh right yep that makes sense. Yay another route type 🤪😂

pleek91 (approved):

src/types/route.ts
stackoverfloweth: lol this one is not exported so it doesn't count

pleek91: Haha I can get behind that philosophy

PR #153, opened 2024-05-02
transition docs added

stackoverfloweth:


pleek91 (approved):
Nice, great for now. 

I'm hoping we can improve how transitions work before a 1.0.0. For example currently it won't transition to/from a rejection.

PR #154, opened 2024-05-02
composables docs updated

stackoverfloweth:
this is the last page needed 🎉

### DIFF
--- a/docs/advanced-concepts/composables.md
+++ b/docs/advanced-concepts/composables.md
@@ -1,1 +1,33 @@
 # Composables
+
+## useRoute
+
+Returns the current route with built in type guard.
+
+```ts
+import { useRoute } from '@kitbag/router'
+
+const route = useRoute('route.key.here')
+```
+
+the route key argument is what acts as a route guard, providing back the correct types for `params` and `update`. This route key argument is optional on `useRoute`, if omitted will return the same wider type as `Router.route`. The key must match the current route, otherwise router will throw an error.
+
+## useRouter
+
+Returns the installed router, with types provided by [the `Register` interface](/getting-started#update-registered-router).
+
+```ts
+import { useRouter } from '@kitbag/router'
+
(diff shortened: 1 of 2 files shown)


docs/advanced-concepts/composables.md
```diff
@@ -1 +1,33 @@
...
+
+Returns the current route with built in type guard.
+
+```ts
+import { useRoute } from '@kitbag/router'
+
+const route = useRoute('route.key.here')
+```
+
+the route key argument is what acts as a route guard, providing back the correct types for `params` and `update`. This route key argument is optional on `useRoute`, if omitted will return the same wider type as `Router.route`. The key must match the current route, otherwise router will throw an error.
```
pleek91, line 13: Can we expand the type guard part to include how it works if the key is any part of the current route. It can be a parent.

docs/advanced-concepts/composables.md
```diff
@@ -1 +1,33 @@
...
+
+## useRouter
+
+Returns the installed router, with types provided by [the `Register` interface](/getting-started#update-registered-router).
```
pleek91, line 17: Nit
```suggestion
Returns the router, with types provided by [the `Register` interface](/getting-started#update-registered-router).
```

docs/advanced-concepts/composables.md
```diff
@@ -1 +1,33 @@
...
+
+## hooks
+
+All of the hooks are available as composables, with the exception of `onBeforeRouteEnter`
+
+- onBeforeRouteUpdate
+- onBeforeRouteLeave
+- onAfterRouteLeave
+- onAfterRouteUpdate
+- onAfterRouteEnter
```
pleek91, line 33: I would omit this section and leave it to the hooks documentation. I don't think we should refer to these as compositions.

pleek91 (approved):

PR #156, opened 2024-05-04
folder structure

stackoverfloweth:
- added services folder, restructured files
- only exporting what's intended to be public


pleek91 (approved):
LGTM

PR #159, opened 2024-05-06
fix children not having param context of parent

stackoverfloweth:
drop separate params buckets in favor of combined path and query which also contain params

### DIFF
--- a/src/types/route.ts
+++ b/src/types/route.ts
@@ -19,12 +19,10 @@ export type Route<
   key: TKey,
   path: ToPath<TPath>,
   query: ToQuery<TQuery>,
-  pathParams: ToPath<TPath>['params'],
-  queryParams: ToQuery<TQuery>['params'],
   depth: number,
   disabled: TDisabled extends boolean ? TDisabled : false,
 }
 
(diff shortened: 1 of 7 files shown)


pleek91 (approved):

src/types/route.ts
```diff
@@ -19,12 +19,10 @@ export type Route<
...
   path: ToPath<TPath>,
   query: ToQuery<TQuery>,
-  pathParams: ToPath<TPath>['params'],
-  queryParams: ToQuery<TQuery>['params'],
   depth: number,
   disabled: TDisabled extends boolean ? TDisabled : false,
 }
 
-export type ExtractRouteParamTypes<TRoute extends { pathParams: Record<string, unknown>, queryParams: Record<string, unknown> }> = TRoute extends { pathParams: infer PathParams extends Record<string, Param | undefined>, queryParams: infer QueryParams extends Record<string, Param | undefined> }
+export type ExtractRouteParamTypes<TRoute extends { path: { params: Record<string, unknown> }, query: { params: Record<string, unknown> } }> = TRoute extends { path: { params: infer PathParams extends Record<string, Param | undefined> }, query: { params: infer QueryParams extends Record<string, Param | undefined> } }
```
pleek91, line 26: Could these extends just extend Path and Query?

stackoverfloweth: This subtype of path and query were added as defaults I believe to improve dx when users don't override registered routes.

PR #160, opened 2024-05-08
several fixes for query bugs, mostly pivoting from using regex to using URLSearchParams class

stackoverfloweth:
kept finding bugs in my newest test case where parent had a query param, 1 of 3 children of that parent also added a query param.


src/services/combineQuery.ts
```diff
@@ -22,5 +23,9 @@ export function combineQuery(parentQuery: Query, childQuery: Query): Query {
     throw new DuplicateParamsError(key)
   }
 
-  return query(`${parentQuery.query}${childQuery.query}`, { ...parentQuery.params, ...childQuery.params })
+  const newQueryString = [parentQuery.query, childQuery.query]
+    .filter(stringHasValue)
```
pleek91, line 27: How does this work with an optional param that is undefined vs an empty string?

src/services/routeRegex.spec.ts
```diff
@@ -97,4 +97,32 @@ describe('generateRouteQueryRegexPatterns', () => {
...
+describe('getParamName', () => {
+  test('given string with optional param name syntax, returns param name', () => {
+    const paramName = 'foo'
+
+    const response = getParamName(`:?${paramName}`)
+
+    expect(response).toBe(paramName)
+  })
+
+  test('given string with optional param name syntax, returns param name', () => {
```
pleek91, line 111:
```suggestion
  test('given string with param name syntax, returns param name', () => {
```

src/services/combineQuery.ts
stackoverfloweth: this is inside of `combineQuery`, which is taking a parent route and a child route and creating a new Query that's the combination of both. So there is no undefined or empty strings here

PR #164, opened 2024-05-10
New `component` utility for binding props and attrs

pleek91:
# Description
Adds a new utility to be used when creating routes that allows you to bind props or attributes. The first arg is the component to mount and the second is a callback which must return the props/attributes to bind to the component. This callback can be async. 

I'd like to see if we can pass params to the props callback. But I'll explore that in a follow up.


src/types/routeProps.ts
```diff
@@ -1,15 +1,10 @@
...
 import { AfterRouteHook, BeforeRouteHook } from '@/types/hooks'
 import { Path } from '@/types/path'
 import { Query } from '@/types/query'
 import { Routes } from '@/types/route'
 import { MaybeArray } from '@/types/utilities'
 
-/**
- * Type alias for Vue components, which can be either synchronous or asynchronous components.
- */
-export type RouteComponent = Component | DefineComponent | AsyncComponentLoader
```
pleek91, line 11: I removed this type because I don't think we need it. I believe the built in `Component` type encompasses all of these types. All the tests and types passed without the extras.

src/services/component.browser.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+      name: 'two',
+      path: '/two',
+      component: component(echo, async () => {
+        return await { value: 'two' }
+      }),
+    },
+    {
+      name: 'three',
+      path: '/three',
+      component: component(defineAsyncComponent(() => import('@/components/echo')), () => ({ value: 'three' })),
```
pleek91, line 26: Something interesting to note is I couldn't get this to work with just the dynamic import. I had to wrap it in `defineAsyncComponent`. The types work just fine, but I couldn't get the component to actually render.

src/components/echo.ts
stackoverfloweth, line 1: 👀 what's this?

stackoverfloweth: assuming it's just for testing? thoughts on moving this into `testHelpers.ts`?

src/services/component.ts
stackoverfloweth, line 1: love this. I don't suppose you tested this in an actual vue app?

src/types/routeProps.ts
stackoverfloweth: I'm all for simplicity if we can make things simpler. There will be some doc changes required if both our PRs get merged (note for future us)

src/services/component.ts
pleek91: No just the unit tests

src/components/echo.ts
pleek91: Its a vue component so it has to be a default export. Similar to the `helloWorld.ts` component. All for moving those somewhere else if we want to. But basically just for testing, its a vue component that just spits out whatever value you give it as the `value` prop. Used in the unit test I wrote for the `component` utility to test different use cases.

stackoverfloweth (approved):
🔥🔥🔥

PR #165, opened 2024-05-10
Typedoc with jsdocs

stackoverfloweth:
Primary objective, add API docs for components, composables, errors, functions, interfaces, and types.

80% generated with jsdocs with some small tweaks. Added script for re-generating files, though as I mentioned there was a not insignificant amount of hand crafting to improve generated docs.

Also adds RouteProps type in bundle, adds jsdoc strings to RouteProps and ResolvedRoute.


docs/.vitepress/config.ts
```diff
@@ -15,8 +15,8 @@ export default defineConfig({
     },
     
     nav: [
-      { text: 'Guide', link: '/core-concepts' },
-      { text: 'API', link: '/api' }
+      { text: 'Guide', link: '/core-concepts/defining-routes' },
```
pleek91, line 18: I feel like the intro or getting started pages are better entry points for the guide

pleek91 (approved):
Didn't take the time to read all of this. But its awesome. Docs look so good

PR #178, opened 2024-05-19
fixed bug with non-string optional params

stackoverfloweth:


src/services/params.ts
```diff
@@ -29,7 +29,7 @@ export function optional<TParam extends Param>(param: TParam): OptionalParamGetS
       return getParamValue(value, param)
     },
     set: (value) => {
-      if (!stringHasValue(value)) {
+      if (value === undefined) {
```
pleek91, line 32: Should we have a test for setting optional params?

pleek91 (approved):

src/services/params.ts
stackoverfloweth: added ✅

Issue #179, opened 2024-05-19
update Router.find logic


stackoverfloweth:
I found with external urls, or possibly for any url that shouldn't actually match a route
if you have a route that's only a slash, `router.resolve('https://github.io')` will return that route

My specific test case is route

```ts
  {
    name: 'home',
    path: '/',
    component: HomeView,
  },
```

PR #181, opened 2024-05-22
Fix router.find returning any route with `/` as the path when resolving an external url with no path

pleek91:
# Description
Adds a route match rule for host to make sure that `http://example.com` doesn't match a route with path `/`. This would previously match because the path for `https://example.com` is `/`. So we need to check the host itself. 

Closes https://github.com/kitbagjs/router/issues/179

### DIFF
--- a/src/services/createRouterFind.spec.ts
+++ b/src/services/createRouterFind.spec.ts
@@ -1,0 +1,49 @@
+import { expect, test } from 'vitest'
+import { RouteNotFoundError } from '@/errors/routeNotFoundError'
+import { createRouterFind } from '@/services/createRouterFind'
+import { routes } from '@/utilities'
+
+test('when given a key that matches a route return that route', () => {
+  const find = createRouterFind(routes)
+
+  const route = find('parentB')
+
+  expect(route).toBeDefined()
+  expect(route?.key).toBe('parentB')
+})
+
+test('when given a url that matches a route returns that route', () => {
+  const find = createRouterFind(routes)
+  const route = find('/parentB')
+
+  expect(route).toBeDefined()
+  expect(route?.key).toBe('parentB')
+})
+
+test('when given a key that does not match a route returns undefined', () => {
+  const find = createRouterFind(routes)
+
+  // @ts-expect-error
+  expect(() => find('parentD')).toThrow(RouteNotFoundError)
+})
+
+test('when given a url that does not match a route returns undefined', () => {
+  const find = createRouterFind(routes)
+  const route = find('/parentC')
+
+  expect(route).toBeUndefined()
+})
+
+test('when given a url that does not match a route returns undefined', () => {
+  const find = createRouterFind(routes)
+  const route = find('/does-not-exist')
+
+  expect(route).toBeUndefined()
+})
+
+test('when given an external url that does not match a route returns undefined', () => {
+  const find = createRouterFind(routes)
+  const route = find('https://example.com')
+
+  expect(route).toBeUndefined()
+})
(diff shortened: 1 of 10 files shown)


src/services/createRouterFind.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  expect(route).toBeDefined()
+  expect(route?.key).toBe('parentB')
+})
+
+test('when given a key that does not match a route returns undefined', () => {
+  const find = createRouterFind(routes)
+
+  // @ts-expect-error
+  expect(() => find('parentD')).toThrow(RouteNotFoundError)
```
pleek91, line 27: The new errors in this PR were added because I wanted to test this.

src/services/getResolvedRouteForUrl.ts
```diff
@@ -2,14 +2,21 @@ import { readonly } from 'vue'
...
 import { Routes } from '@/types/route'
 import { RouteMatchRule } from '@/types/routeMatchRule'
 
+const rules: RouteMatchRule[] = [
+  isNamedRoute,
+  routeHostMatches,
+  routePathMatches,
+  routeQueryMatches,
+  routeParamsAreValid,
+]
```
pleek91, line 17: We were creating this const every time `getResolvedRouteForUrl` was called. Pulling it out into its own static const.

src/services/getResolvedRouteForUrl.ts
```diff
@@ -2,14 +2,21 @@ import { readonly } from 'vue'
...
-import { routePathMatches, routeQueryMatches } from '@/services/routeMatchRegexRules'
+import { isNamedRoute, routeHostMatches, routePathMatches, routeQueryMatches } from '@/services/routeMatchRules'
 import { getRouteScoreSortMethod } from '@/services/routeMatchScore'
 import { ResolvedRoute } from '@/types/resolved'
 import { Routes } from '@/types/route'
 import { RouteMatchRule } from '@/types/routeMatchRule'
 
+const rules: RouteMatchRule[] = [
+  isNamedRoute,
+  routeHostMatches,
```
pleek91, line 13: Added this new rule

src/utilities/testHelpers.ts
```diff
@@ -14,7 +14,7 @@ export const component = { template: '<div>This is component</div>' }
 export const routes = createRoutes([
   {
     name: 'parentA',
-    path: '/[paramA]',
+    path: '/parentA/[paramA]',
```
pleek91, line 17: I update this route (and associated tests) because the path for `parentB` is `/parentB` which matched `/[paramA]`. Discovered while writing tests for `createRouterFind`.

src/services/routeMatchRules.ts
```diff
@@ -2,6 +2,17 @@ import { createMaybeRelativeUrl } from '@/services/createMaybeRelativeUrl'
 import { generateRoutePathRegexPattern, generateRouteQueryRegexPatterns } from '@/services/routeRegex'
 import { RouteMatchRule } from '@/types/routeMatchRule'
 
+export const isNamedRoute: RouteMatchRule = (route) => {
+  return 'name' in route.matched && !!route.matched.name
+}
+
+export const routeHostMatches: RouteMatchRule = (route, url) => {
+  const { host: urlHost } = createMaybeRelativeUrl(url)
+  const { host: routeHost } = createMaybeRelativeUrl(route.path.toString())
```
stackoverfloweth, line 11: would a `route` ever have a host? I feel like all this is really testing is that the host is either omitted from url or that it's not `https://localhost` (which is the value we use when given a relative path not absolute URL).

For example, I think if `router.find` was given `https://localhost/parentA` it would think it's always same host even when hosted

src/services/createRouterFind.spec.ts
stackoverfloweth, line 1: can't believe this didn't exist before...

src/services/routeMatchRules.ts
pleek91: Currently no route would never have a host. But once we add support for external urls as defined routes this should just work to match those as well. 

But this will make sure that any url with a host doesn't match a route with a relative path. Otherwise any url with a host could match a route with a relative path.

stackoverfloweth: Ah yes, I forgot it leaves host off when url is relative.

stackoverfloweth (approved):

Issue #144, opened 2024-04-27
route match not working for kebab param names


stackoverfloweth:
url: http://localhost:5173/walk/zoo

routes

```ts
{
    name: 'walk', 
    path: '/walk', 
    component: () => import('../views/Walk.vue'),
    children: createRoutes([
      {
        name: 'walk-my-dogs', 
        path: '/:my-Param',
        component: () => import('../views/DogsList.vue')
      },
      {
        name: 'walk-shelter-dogs', 
        path: 'shelter-dogs',
        component: () => import('../views/DogsList.vue'),
        meta: {
          showShelterDogs: true
        }
      }
    ])
  },
```

if I change path on `walk-my-dogs` to 

```ts
path: '/myParam'
```

it works fine

kktos:
At first glance, it doesn't seem a problem, from my point of view.
I mean, the param name is a variable name at the end, right ?
so it should comply to the same js rules we have for a variable name, don't you think ?

stackoverfloweth:
> At first glance, it doesn't seem a problem, from my point of view. I mean, the param name is a variable name at the end, right ? so it should comply to the same js rules we have for a variable name, don't you think ?

I agree that's it's not a high priority problem but devs coming from vue-router will assume it works this way and it should be relatively easy to convert kebab-case param names into camelCase variable names

pleek91:
Prob fine if we said they had to be camel case. But I agree it be nice if it just worked either way and wasn't opinionated.

PR #182, opened 2024-05-23
Fix param names unable to be snake or kebab case

pleek91:
# Description
Since the param regex only looked for `\w` characters any param name that was snake or kebab case would not work. Adding `-` and `_` to the regex patterns fixes this.

Closes https://github.com/kitbagjs/router/issues/144


stackoverfloweth (approved):
This is great, nice find. Would it be worth the dependency on string-ts to automatically convert kebab into camelCase (in a follow up PR)?

PR #183, opened 2024-05-25
Add `isRoute` type guard with `exact` option

pleek91:
# Description
The current route can be type guarded using `useRoute('route.key')`. Adding and `isRoute` type guard so that any value can be type guarded to determine
A. If it is a `RouterRoute`
B. Which user defined route it is

This allows for this type of logic

```ts
const route = useRoute()

if(isRoute(route, 'settings') {
  // do settings logic
}

if(isRoute(route, 'home') {
  // do home logic
}
```

There is also an optional 3rd argument for options where the `exact` option can be passed. Passing this settings means that if the current route is `settings.foo` then `isRoute(route, 'settings', { exact: true })` will return `false`.


src/guards/routes.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (exact) {
+    const actualRouteKey = keys.at(-1)
+
+    return routeKey === actualRouteKey
+  }
+
+  return keys.includes(routeKey)
+}
+
+function getRouteKeys(route: RouterRoute): string[] {
```
pleek91, line 32: I moved this from `useRoute.ts` and modified it to accept the route itself rather than the names.

src/main.ts
```diff
@@ -2,4 +2,5 @@ export * from './components'
 export * from './compositions'
 export * from './errors'
 export * from './services'
-export * from './types'
\ No newline at end of file
+export * from './types'
+export { isRoute } from './guards/routes'
```
pleek91, line 6: I'd like to start doing this in the main.ts file. That way we have to be more intentional about what is exported publicly.

src/services/createRouterRoute.ts
```diff
@@ -3,11 +3,18 @@ import { RouterPush, RouterPushOptions } from '@/types/routerPush'
 import { RouteUpdate } from '@/types/routeUpdate'
 import { Writable } from '@/types/utilities'
 
+const isRouterRouteSymbol = Symbol('isRouterRouteSymbol')
```
pleek91, line 6: I needed a way to type guard the RouterRoute. This works but I'd like to revisit this. The proxy based approach is going to get more and more complex and fragile as we add stuff to this type.

stackoverfloweth (approved):
You know we have a "mature" project when 75% of the changes are documentation and unit tests 😅

Actual changes look good. I'd like to see us lean into the new "/guards" folder and move all of our guards into there. 

Nice added functionality to useRoute to be able to assert exact.

src/services/createRouterRoute.ts
pleek91: addressed in #199

Issue #174, opened 2024-05-17
add support for Named Views


stackoverfloweth:
https://router.vuejs.org/guide/essentials/named-views.html#Named-Views

PR #185, opened 2024-05-26
Named router views

pleek91:
# Description
Adds the ability for a route to define multiple components rather than a single component. These components can then be rendered using named router views. A router view with no name will render the `default` component which is either `route.component` or `route.components.default`.

Closes https://github.com/kitbagjs/router/issues/174

Needs docs

**main.ts**
```ts
const routes = createRoutes([
  {
    name: 'parent',
    path: '/',
    components: {
      default: { template: '<p>default</p>' },
      one: { template: '<p>one</p>' },
      two: { template: '<p>two</p>' },
    },
  },
])

...
```
**App.vue**
```vue
<template>
    <RouterView name="one" />
    <RouterView />
    <RouterView name="two" />
</template>
```
**Rendered html** 
```html
<p>one</p>
<p>default</p>
<p>two</p>
```

### DIFF
--- a/src/components/routerView.browser.spec.ts
+++ b/src/components/routerView.browser.spec.ts
@@ -260,3 +260,39 @@ test('Renders the route component when the router.push does match after a reject
 
   expect(app.text()).toBe('hello world')
 })
+
+test('Renders the multiple components when using named route views', async () => {
+  const routes = createRoutes([
+    {
+      name: 'parent',
+      path: '/',
+      components: {
+        default: { template: '_default_' },
+        one: { template: '_one_' },
+        two: { template: '_two_' },
+      },
+    },
+  ])
+
+  const router = createRouter(routes, {
+    initialUrl: '/',
+  })
+
+  await router.initialized
+
+  const root = {
+    template: `
+      <RouterView name="one" />
+      <RouterView />
+      <RouterView name="two" />
+    `,
+  }
+
+  const app = mount(root, {
+    global: {
+      plugins: [router],
+    },
+  })
+
+  expect(app.text()).toBe('_one__default__two_')
+})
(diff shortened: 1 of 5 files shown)


src/components/routerView.browser.spec.ts
```diff
@@ -188,6 +189,10 @@ test('Renders custom genericRejection component when the initialUrl does not mat
     },
   })
 
+  if (!isRouteWithComponent(router.route.matched)) {
+    throw 'Matched route does not have a single component'
+  }
```
pleek91, line 194: I'm not 100% sold on the types here. See `routeProps.ts` for why this is necessary

src/components/routerView.vue
pleek91, line 1: This PR is branched off of https://github.com/kitbagjs/router/pull/184 so there are a few unrelated changes that will go away once that is merged.

src/types/routeProps.ts
```diff
@@ -30,10 +30,24 @@ type WithHooks = {
...
+   * Multiple components for named views, which can be either synchronous or asynchronous components.
+   */
+  components: Record<string, Component>,
+}
+
 /**
  * Represents properties common to parent routes in a route configuration, including hooks, path, and optional query parameters.
  */
-export type ParentRouteProps = WithHooks & {
+export type ParentRouteProps = Partial<WithComponent | WithComponents> & WithHooks & {
```
pleek91, line 50: Wrote this in a way where you can define a `component` property or a `components` property but not both. Which means if you want to do anything with either of those properties you have to type guard it first to determine which one is available. 

I think this is a good devx when defining routes using `createRoutes`. But its a little annoying if you need to access the component from `route.match.component` or `route.match.components` because you have to type guard it. But I don't expect users to be accessing those properties. So I think its safe to take on that slight annoyance in our internal codebase to give users a good devx when defining routes.

stackoverfloweth (approved):

PR #191, opened 2024-06-04
found a rabbit hole

stackoverfloweth:
quests:
- remove readonly complexity from types
- fix Error: [vite:dts] Internal Error: Unable to follow symbol for "const"
-


src/services/createRouter.ts
```diff
@@ -171,7 +172,7 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
   function install(app: App): void {
     app.component('RouterView', RouterView)
     app.component('RouterLink', RouterLink)
-    app.provide(routerInjectionKey, router)
+    app.provide(routerInjectionKey, router as RegisteredRouter)
```
stackoverfloweth, line 175: this cast makes sense. `router` is `Router<T>` and the injection key expected registered router

src/services/createRouterFind.ts
```diff
@@ -1,15 +1,17 @@
...
+import { Route, Routes } from '@/types/route'
 import { RoutesKey } from '@/types/routesMap'
 import { RouteParamsByKey } from '@/types/routeWithParams'
 import { Url } from '@/types/url'
 import { AllPropertiesAreOptional } from '@/types/utilities'
 
 type RouterFindArgs<
   TRoutes extends Routes,
-  TSource extends string & keyof RoutesKey<TRoutes>,
+  TSource extends RoutesKey<TRoutes>,
```
stackoverfloweth, line 14: this is an example of an actual error that we were missing. `RoutesKey` was already the keys, so `keyof` again was giving me something like the proto properties on `string` type

src/services/createRouterFind.ts
```diff
@@ -24,11 +26,11 @@ export type RouterFind<
 
 export function createRouterFind<const TRoutes extends Routes>(routes: TRoutes): RouterFind<TRoutes> {
 
-  return <TRoutes extends Routes, TSource extends Url | RoutesKey<TRoutes>>(
+  return <TRoutes extends Routes, TSource extends RoutesKey<TRoutes>>(
     source: TSource,
     params: Record<PropertyKey, unknown> = {},
   ): ResolvedRoute | undefined => {
-    const resolve = createRouterResolve(routes)
+    const resolve = createRouterResolve(routes as Route<string, Path, Query, false>[])
```
stackoverfloweth, line 33: this one was annoying. I had to widen routes to make `resolve` happy. It kept complaining that source wasn't satisfies `Url | RoutesKey<T>`

src/services/createRouterRoute.ts
```diff
@@ -39,16 +42,16 @@ export type RouterRoute<TRoute extends ResolvedRoute = ResolvedRoute> = Readonly
...
   [isRouterRouteSymbol]: true,
-}>
+}
 
 export function isRouterRoute(value: unknown): value is RouterRoute {
   return typeof value === 'object' && value !== null && isRouterRouteSymbol in value && value[isRouterRouteSymbol] === true
 }
 
-export function createRouterRoute<TRoute extends ResolvedRoute>(route: TRoute, push: RouterPush): RouterRoute<TRoute> {
+export function createRouterRoute<TRoute extends ResolvedRoute>(route: TRoute, push: RouterPush<Route<string, Path, Query, false>[]>): RouterRoute<TRoute> {
```
stackoverfloweth, line 54: you'll see this in a couple places, the default `Route` (and `Routes`) assumes key might be undefined and disabled might be true. This version is more accurate of a "resolved route" where the route must be named and not disabled

src/types/register.ts
```diff
@@ -33,7 +33,7 @@ export type RegisteredRouter = Register extends { router: infer TRouter }
  */
 export type RegisteredRoutes = Register extends { router: Router<infer TRoutes extends Routes> }
   ? TRoutes
-  : Route<string, Path<'', {}>, Query<'', {}>, false>[]
+  : Route<string, Path, Query, false>[]
```
stackoverfloweth, line 36: I loosened the default `Path` and `Query` when type argument isn't provided.

PR #194, opened 2024-06-09
Improved path query types

stackoverfloweth:
- Improved the way we capture optional params in type system. 
  - previously we were defining the param inside of `params` as `StringConstructor | undefined` for example. So basically if a Param has `| undefined`, it meant it should be optional. 
  - this adds the complication of having to check for missing values and is redundant since we can derive the fact that it's optional by looking for the optional symbol in the Param.
- I updated the definition for `ResolvedRoute` so that if it's created without a generic (using the default), it will widen params to `Record<string, unknown>`, otherwise it would be `{}`.
- path and query types have stronger typed defaults, when used without defaults will have `Record<string, Param>` for `params`.

### DIFF
--- a/src/services/params.ts
+++ b/src/services/params.ts
@@ -9,8 +9,11 @@ export function getParam<P extends Record<string, Param | undefined>>(params: P,
 
 const optionalKey = Symbol()
 
-type OptionalParamGetSet<TParam extends Param, TValue = ExtractParamType<TParam> | undefined> = ParamGetSet<TValue> & {
+export type IsOptionalParam = {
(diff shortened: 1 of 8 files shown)


pleek91:
@stackoverfloweth seems fine. What problem are you solving here? The benefit isn't obvious to me.

stackoverfloweth:
> @stackoverfloweth seems fine. What problem are you solving here? The benefit isn't obvious to me.

@pleek91, I've updated the readme to include some additional context, hope that helps

pleek91 (approved):

PR #195, opened 2024-06-09
Param defaults

stackoverfloweth:
This PR introduces a feature we've wanted for as long as we've had params, **default values**. 

From out initial release we've always had optional params. A param that is optional changes the route matching logic (regex), introduces the possibility of `undefined` for the type when accessing `route.params.myParam`, and correspondingly allows you to push to the route without supplying the optional param.

Defining an optional param on a route is driven by the question mark (?) inside the param syntax.

```ts
import { path } from '@kitbag/router'

{
  name: 'example-route',
  path: path('/path/[?sort]', { sort: String })
}
```

In this example, the `sort` will be type `string | undefined`.

After this PR, you'll be able to extend that functionality by defining a default value you wish to be used instead of `undefined` if no value is supplied/found. Let's extend the example above to provide a logic default value.

```ts
import { path, withDefault } from '@kitbag/router'

{
  name: 'example-route',
  path: path('/path/[?sort]', { sort: withDefault(String, 'asc' })
}
```

Now when we access the param, the type will be `string`.

```ts
import { useRoute } from '@kitbag/router'

const route = useRoute('example-route')

route.params.sort
```

Notice in my example I used the `withDefault` utility to take any Param and assign a default value. Developers also have the option of assigning a default when creating a custom param.

```ts
const myParam: ParamGetSet<'yes' | 'no'> = {
  get: (value, { invalid }) => {
    if (value !== 'yes' && value !== 'no') {
      invalid()
    }

    return value
  },
  set: value => value.toString(),
  defaultValue: 'no'
}
```


pleek91:
Some thoughts based on this example of the syntax

```ts
path('/simple/[simple]', { 
  simple: withDefault(String, 'abc') 
})
```

- the `withDefault` utility is a little too generic IMO. 
- The syntax might be a little awkward with inline params.
```ts
withDefault({
  get: ...
  set: ...
}, 'abc')
```
- I imagined the param itself would still be marked optional. `'/simple/[simple]'` as a path reads that `[simple]` is required. 

Let's sync up on this. Love to see this coming together. Super cool feature, especially for things like a query param with a default value so that you don't have to deal with `undefined` everywhere in your component code.

src/services/routeMatchRules.spec.ts
```diff
@@ -168,6 +170,26 @@ describe('routeQueryMatches', () => {
     expect(response).toBe(true)
   })
 
+  test.each([
+    ['?default'],
+    ['?default='],
```
pleek91, line 175: What is the param value for these two cases? 

Also should there be a test case where the param is completely missing?

src/services/routeRegex.spec.ts
```diff
@@ -50,6 +53,23 @@ describe('generateRoutePathRegexPattern', () => {
     const expected = new RegExp(`^parent/child/${catchAll}/grand-child/${catchAll}$`, 'i')
     expect(result.toString()).toBe(expected.toString())
   })
+
+
+  test('given path with default params, returns value with params replaced with catchall', () => {
```
pleek91, line 58: this should work exactly the same as any other param right? Do we need to test this? The regex doesn't factor in the param right?

src/services/urlAssembly.spec.ts
```diff
@@ -41,6 +42,20 @@ describe('path params', () => {
...
+        name: 'simple',
+        path: path('/simple/[?simple]', { simple: withDefault(String, 'abc') }),
+        component,
+      },
+    ])
+
+    const url = assembleUrl(route)
+
+    expect(url).toBe('/simple/abc')
+  })
```
pleek91, line 57: Ah yeah I think this answers my previous question about the setter logic. I don't think this should put the default value in the url. The default value should come from the getter recognizing the value is `undefined`.

src/services/urlAssembly.spec.ts
```diff
@@ -60,6 +75,22 @@ describe('path params', () => {
...
+      },
+    ])
+
+    const url = assembleUrl(route, {
+      params: { simple: 'DEF' },
+    })
+
+    expect(url).toBe('/simple/DEF')
+  })
+
```
pleek91, line 93: Seems like maybe we don't need this test. The param having a default doesn't factor into this logic or types does it?

src/services/urlAssembly.spec.ts
```diff
@@ -150,6 +196,22 @@ describe('query params', () => {
...
+        path: '/',
+        query: query('simple=[simple]', { simple: withDefault(String, 'abc') }),
+        component,
+      },
+    ])
+
+    const url = assembleUrl(route, { params: { simple: '' } })
+
+    expect(url).toBe('/?simple=')
+  })
```
pleek91, line 213: Same as the path test. This is the same even if there is no default right? Probably don't need this test do we?

src/services/urlAssembly.spec.ts
```diff
@@ -170,6 +232,24 @@ describe('query params', () => {
...
+      },
+    ])
+
+    const url = assembleUrl(route, {
+      params: { simple: 'DEF' },
+    })
+
+    expect(url).toBe('/?simple=DEF')
+  })
+
```
pleek91, line 252: Same here?

src/services/urlAssembly.ts
```diff
@@ -41,13 +41,17 @@ function assembleQueryParamValues(query: string, params: Record<string, Param>,
...
+    const isNotParam = !paramName
 
-    if (!paramName) {
+    if (isNotParam) {
       return { ...url, [key]: value }
     }
 
     const paramValue = setParamValue(paramValues[paramName], params[paramName])
-    if (isOptionalParamSyntax(value) && paramValues[paramName] === undefined && paramValue === '') {
+    const valueNotProvidedAndNoDefaultUsed = paramValues[paramName] === undefined && paramValue === ''
```
pleek91, line 51: I think this is unrelated to this PR but here's another empty string check that I'm curious why we need.

src/services/withDefault.ts
```diff
@@ -0,0 +1,16 @@
+import { asParamGetSet } from '@/services/asParamGetSet'
+import { ExtractParamType, isParamGetSet } from '@/types/params'
+import { Param, ParamGetSet } from '@/types/paramTypes'
+
+export type ParamWithDefault<TParam extends Param, TParamType extends ExtractParamType<TParam> = ExtractParamType<TParam>> = Required<ParamGetSet<TParamType>>
+
+export function isWithDefaultParam(param: Param): param is ParamWithDefault<Param> {
```
pleek91, line 7:
```suggestion
export function isParamWithDefault(param: Param): param is ParamWithDefault<Param> {
```

src/services/withDefault.ts
```diff
@@ -0,0 +1,16 @@
+import { asParamGetSet } from '@/services/asParamGetSet'
+import { ExtractParamType, isParamGetSet } from '@/types/params'
+import { Param, ParamGetSet } from '@/types/paramTypes'
+
+export type ParamWithDefault<TParam extends Param, TParamType extends ExtractParamType<TParam> = ExtractParamType<TParam>> = Required<ParamGetSet<TParamType>>
```
pleek91, line 5: Seems like we don't need the second generic? That can just be part of the return type right?

src/services/withDefault.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function isWithDefaultParam(param: Param): param is ParamWithDefault<Param> {
+  return isParamGetSet(param) && param.defaultValue !== undefined
+}
+
+export function withDefault<TParam extends Param, TParamType extends ExtractParamType<TParam> = ExtractParamType<TParam>>(param: TParam, defaultValue: TParamType): ParamWithDefault<TParam, TParamType> {
```
pleek91, line 11: Same question about the second generic here.

src/types/paramTypes.ts
```diff
@@ -8,6 +8,7 @@ export type ParamSetter<T = any> = (value: T, extras: ParamExtras) => string
 export type ParamGetSet<T = any> = {
   get: ParamGetter<T>,
   set: ParamSetter<T>,
+  defaultValue?: T,
```
pleek91, line 11: ❤️

pleek91 (commented):
@stackoverfloweth this is looking really good. The only major question is about the setter for a param with a default. We can sync up on that

src/services/params.ts
```diff
@@ -186,6 +153,14 @@ export function getParamValue<T extends Param>(value: string | undefined, param:
 }
 
 export function setParamValue(value: unknown, param: Param): string {
+  if (value === undefined) {
+    if (isOptionalParam(param)) {
+      return param.defaultValue ?? ''
+    }
+
+    throw new InvalidRouteParamValueError()
+  }
```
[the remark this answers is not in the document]
stackoverfloweth: yes. populating a value for `defaultValue` on `ParamGetSet` is what defines a param as a `ParamWithDefault`. That's all the `withDefault` utility does, it turns the param into a `ParamGetSet` (if it wasn't already), and populates this property.

src/services/params.ts
```diff
@@ -138,9 +105,9 @@ const jsonParam: ParamGetSet<unknown> = {
 
 export function getParamValue<T extends Param>(value: string | undefined, param: T): ExtractParamType<T>
 export function getParamValue<T extends Param>(value: string | undefined, param: T): unknown {
-  if (value === undefined) {
+  if (value === undefined || !stringHasValue(value)) {
```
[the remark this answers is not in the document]
stackoverfloweth: this brings the same functionality as was previously established in the `optional` utility I removed

image

Previously, the optional utility wrapped the param and first checked for `!stringHasValue` and then would return `undefined`.  With the new paradigm our logic for what to do with optional params is handled here instead of in that wrapper param.

One difference this introduces is now an empty string will always throw an `InvalidRouteParamValueError` if the param is not optional.

I'm checking both conditions here even though `stringHasValue` also checks `value === undefined`. This is to ensure the type for `value` is narrowed to exclude `undefined` for the conditions below this one.

src/services/routeMatchRules.spec.ts
stackoverfloweth: I would expect the default value `abc` to be the value. Worth testing though

I also would like to test if the param is completely missing. Let's leave this comment unresolved until I do these 2 thigns

src/services/routeRegex.spec.ts
stackoverfloweth: correct, this was from my first pass at defaults where it also made them optional. I will remove. Good catch

src/services/urlAssembly.ts
stackoverfloweth: I updated this bit of code because I found it hard to follow. Basically what it's doing is omitting from the search any params that were optional, not included in params, and do not have default values.

This prevents a `push('route-with-optional-query-params', {})` from giving you a URL like `/route?optional=&another=&third=`

src/services/withDefault.ts
stackoverfloweth: good call, found a couple more than just this file as well 👍

src/services/routeMatchRules.spec.ts
stackoverfloweth: I added cases for "" and "?", I also verified that the default value WILL be present in `route.params` for these cases will be populated with default value

stackoverfloweth:
@pleek91 one more thing I'd like to add to this PR before merging is to see if we can fix the types so that a path like

```ts
{
  name: 'my-example',
  path: path('/path/[?id]', { id: withDefault(String, 'abc') }),
}
```
Accessing the param, `params.id` won't be `string | undefined`.

pleek91:
> @pleek91 one more thing I'd like to add to this PR before merging is to see if we can fix the types so that a path like
> 
> 
> 
> ```ts
> 
> {
> 
>   name: 'my-example',
> 
>   path: path('/path/[?id]', { id: withDefault(String, 'abc') }),
> 
> }
> 
> ```
> 
> Accessing the param, `params.id` won't be `string | undefined`. 

Agreed! That's an important piece.

stackoverfloweth:
@pleek91 I fixed the types issue I was having yesterday. Params with default will not remain optional when calling `push`, `replace`, etc but NOT be possibly undefined when reading `router.route` or `useRoute`. 

I ended up going back on our decision to introduce a separate `ParamGetSet` and `ParamGetSetWithDefault`. That was helpful to discover where I was losing my default type but I was able to achieve the same end without changing as much.

image

I prefer having `defaultValue?: T` on ParamGetSet over defining another top level `Param` and having to include it in our big union of possible `Param` types. I think the devex of defining a param as `:ParamGetSet<MyType>` and not having to make the decision about defaults at that moment is better.

pleek91 (approved):
Nice work, this is awesome 👏

PR #198, opened 2024-06-19
Update component utility to not require suspense unless props are async

pleek91:
# Description
https://github.com/kitbagjs/router/pull/164 introduced the `component` utility which allowed binding props and attrs directly to route components. But it always required suspense because even if the props getter was not async it would wrap it in an async component. 

This PR fixes that by only wrapping the component in an async component IF the getter returns a promise.


stackoverfloweth (approved):
🚀

PR #199, opened 2024-06-21
Fix `RouterRoute` type so that `isRoute` type guard works correctly when given a `RouterRoute`

pleek91:
# Description
The `isRoute` type guard was functionally sound and types work just fine. But `RouterRoute` produced a type that was too specific and to wide. So when passing a `RouterRoute` into `isRoute` the types conflicted and effectively didn't narrow properly. A few changes were necessary to fix this.

1. Update `Route` type to default to a `string` for the key rather than `any`. Because `any & 'key'` is still `any`. But `string & 'key'` is `'key'`. Which is what we want. 
2. Remove `Omit` utility from `RouterRoute` declaration. This produces a much simpler type and one that typescript can reason about without evaluating it.
3. Replace all `DeepReadonly` types with `Readonly` to produce simpler types that typescript can reason about more easily (and produce much more readable types)
4. Remove all uses of `readonly` (to satisfy the `DeepReadonly` removal)

3 & 4 were not strictly necessary. I had a working version that still used the readonly types, but everything looks much better without it. I want to revisit this in a follow up PR. For now I've marked the one test that covered the readonly pieces as `fails`.

### DIFF
--- a/src/services/createCurrentRoute.ts
+++ b/src/services/createCurrentRoute.ts
@@ -1,4 +1,4 @@
-import { readonly, reactive } from 'vue'
+import { reactive } from 'vue'
 import { RouterRoute, createRouterRoute } from '@/services/createRouterRoute'
 import { ResolvedRoute } from '@/types/resolved'
 import { RouterPush } from '@/types/routerPush'
@@ -18,7 +18,7 @@ export function createCurrentRoute(fallbackRoute: ResolvedRoute, push: RouterPus
     Object.assign(route, { ...newRoute })
   }
 
-  const currentRoute = readonly(route)
+  const currentRoute = route
   const routerRoute = createRouterRoute(currentRoute, push)
 
   return {
--- a/src/services/createRouter.spec.ts
+++ b/src/services/createRouter.spec.ts
@@ -82,7 +82,7 @@ test('route update updates the current route', async () => {
 
 })
 
-test('route is readonly except for individual params', async () => {
+test.fails('route is readonly except for individual params', async () => {
   const routes = createRoutes([
     {
       name: 'root',
--- a/src/services/createRouterReject.ts
+++ b/src/services/createRouterReject.ts
@@ -24,8 +24,6 @@ type CreateRouterRejectContext = {
 
 const isRejectionRouteSymbol = Symbol()
 
-type RouterRejectionRoute = ResolvedRoute & { [isRejectionRouteSymbol]?: true }
-
 export type CreateRouterReject = {
   setRejection: RouterSetReject,
   rejection: RouterRejection,
@@ -67,8 +65,8 @@ export function createRouterReject({
     return resolved
   }
 
-  const isRejectionRoute: IsRejectionRoute = (route: RouterRejectionRoute) => {
-    return route[isRejectionRouteSymbol] === true
+  const isRejectionRoute: IsRejectionRoute = (route) => {
+    return isRejectionRouteSymbol in route
   }
 
   const setRejection: RouterSetReject = (type) => {
--- a/src/services/createRouterReject.ts
+++ b/src/services/createRouterReject.ts
@@ -1,4 +1,4 @@
-import { Ref, markRaw, ref, readonly, Component } from 'vue'
+import { Ref, markRaw, ref, Component } from 'vue'
 import { genericRejection } from '@/components/rejection'
 import { createResolvedRouteQuery } from '@/services/createResolvedRouteQuery'
 import { RegisteredRejectionType } from '@/types'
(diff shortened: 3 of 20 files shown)


src/services/createRouterReject.ts
pleek91, line 1: This was unrelated but wanted to unify this to match what I did to fix the build errors for `RouterRoute`

src/services/createRouterRoute.ts
```diff
@@ -1,18 +1,23 @@
...
 
-export type RouterRoute<TRoute extends ResolvedRoute = ResolvedRoute> = Omit<ResolvedRoute, 'params'> & Readonly<{
+export type RouterRoute<TRoute extends ResolvedRoute = ResolvedRoute> = Readonly<{
+  key: TRoute['key'],
+  matched: TRoute['matched'],
+  matches: TRoute['matches'],
+  query: ResolvedRouteQuery,
   params: Writable<TRoute['params']>,
   update: RouteUpdate<TRoute>,
-  [isRouterRouteSymbol]: true,
```
pleek91, line 11: This was the problem that kept popping up in the build. We don't actually want this symbol in the types. We just want to use it internally. Removing it from this type fixes the issue. Which makes total sense because this is a public type and `isRouterRouteSymbol` was not exported.

src/services/getRouteHooks.spec.ts
pleek91, line 1: Moved the util functions into testHelpers so I could use them in another test.

src/types/route.ts
```diff
@@ -10,7 +10,7 @@ export type Routes = Readonly<Route[]>
 /**
  * The Route properties originally provided to `createRoutes`. The only change is normalizing meta to always default to an empty object.
  */
-type RoutePropsWithMeta = RouteProps & { meta: RouteMeta }
+export type RoutePropsWithMeta = RouteProps & { meta: RouteMeta }
```
pleek91, line 13: @stackoverfloweth I exported this because we had some test utils that were recreating this type. Seems like we should just use this one. But curious if there's a reason it might have been duplicated before.

src/services/createCurrentRoute.ts
```diff
@@ -18,7 +18,7 @@ export function createCurrentRoute(fallbackRoute: ResolvedRoute, push: RouterPus
     Object.assign(route, { ...newRoute })
   }
 
-  const currentRoute = readonly(route)
+  const currentRoute = route
```
stackoverfloweth, line 21: we _could_ rename `route` on line 15 to `currentRoute` and skip this const altogether?

src/services/createRouter.spec.ts
```diff
@@ -82,7 +82,7 @@ test('route update updates the current route', async () => {
 
 })
 
-test('route is readonly except for individual params', async () => {
+test.fails('route is readonly except for individual params', async () => {
```
stackoverfloweth, line 85: are you sure we want to assert this fails? I would probably just opt for removing the test

src/services/createRouterRoute.ts
stackoverfloweth: nice job on this, that was certainly stressful to see intermittently.

src/types/route.ts
stackoverfloweth: probably just an attempt to keep a lid on how many "route" types we have available, even if only internally

stackoverfloweth: but I agree that duplicating doesn't feel right either

src/services/createRouterRoute.spec.ts
stackoverfloweth, line 1: I'm surprised to see no previous test file for createRouterRoute. Do you think there are other bits of business logic within that we should add tests for? (in follow-up PR)

stackoverfloweth (approved):
I love this PR, not just for the improved quality of types and devex but I appreciate how thorough you were, comments were super helpful as well.

Should we do a follow-up PR that updates documentation in any way? Probably want to communicate the usefulness of `isRoute`, anything else?

Also, would love to plan another release after this gets merged.

src/services/createRouter.spec.ts
pleek91: My thought is I probably want to bring back this functionality in a future PR. So I'd like to keep it for now.

src/types/route.ts
pleek91: Idk that we need to export this publicly. But we can use it internally.

src/services/createRouterRoute.spec.ts
pleek91: Agreed. I was also surprised.

src/services/createCurrentRoute.ts
pleek91: I was to revisit the readonly functionality of this. So let's just leave it for now.

PR #200, opened 2024-06-22
new `createParam` utility

stackoverfloweth:
This utility effectively already existed under the private `asParamGetSet` function. This PR renames that to `createParam` and adds a `defaultValue` argument with overloads that retain the correct type for `ParamWithDefault`.

This small utility enables developers to create custom params _slightly_ more easily.

```ts
createParam((value, { invalid }) => {
  if (value === 'asc' || value === 'desc') {
    return value
  }

  throw invalid()
}, 'asc')
```

The value of this utility really comes in the setting of default values. Without this utility, developers will need to define their custom ParamGetSets with `satisfies ParamGetSet<MyType>` to prevent losing the fact that the param has a default value. Without doing so, when accessing the routes params developers will find the type is `MyType | undefined` even though we know it has a default value.

The first argument is any `Param`, the second is an optional default value.

```ts
createParam(Boolean, true)
```

It also allows developers to add a default value to existing params.

```ts
createParam(MySortParam, 'asc')
```

If the param is a ParamGetSet with a default value it will be replaced.


src/services/createParam.ts
```diff
@@ -0,0 +1,18 @@
+import { getParamValue, setParamValue } from '@/services/params'
+import { ParamWithDefault } from '@/services/withDefault'
+import { ExtractParamType, isParamGetSet } from '@/types/params'
+import { Param, ParamGetSet } from '@/types/paramTypes'
+
+export function createParam<TParam extends Param>(param: TParam): ParamGetSet<ExtractParamType<TParam>>
+export function createParam<TParam extends Param>(param: TParam, defaultValue: ExtractParamType<TParam>): ParamWithDefault<TParam>
```
pleek91, line 7: Could we add another overload where TParam extends ParamWithDefault? So it could be defined as a single object.

pleek91 (approved):
Nice, I love how simple `createParam(String, 'abc')` is.

pleek91 (approved):

PR #202, opened 2024-06-22
Create param documentation

stackoverfloweth:


docs/core-concepts/route-params.md
```diff
@@ -95,10 +97,10 @@ const routes = createRoutes([
 
 With this getter defined, now our route will only match if the param matches our rules above.
 
-As a `ParamGetter`, the value in `route.params` is still writable, but the set will assume `value.toString()` is sufficient. Alternatively if you use `ParamGetSet`, you can provide the same validation on value set as well.
+As a `ParamGetter`, the value in `route.params` is still writable, but the set will assume `value.toString()` is sufficient. Alternatively if you use the `ParamGetSet` syntax, you can provide the same validation on value set as well.
```
pleek91, line 100:
```suggestion
If you only supply a getter for your param it will assume that `value.toString()` is sufficient when setting the value. Alternatively you can supply both a `get` and a `set` method and provide validation on the value when setting as well.
```

docs/core-concepts/route-params.md
```diff
@@ -190,10 +192,10 @@ const routes = createRoutes([
 ])
 ```
 
-If you already have a `ParamGetSet`, you can also assign the default value right in your param definition
+If you're using `createParam`, you can also assign a default value with the 2nd argument.
```
pleek91, line 195:
```suggestion
You can also assign a default value with the 2nd argument of `createParam`.
```

pleek91 (commented):
Suggestions around leaving out the types and just referencing how `createParam` is used.

pleek91 (approved):

PR #203, opened 2024-06-22
0.4.1

stackoverfloweth:
waiting on https://github.com/kitbagjs/router/pull/202

pushed with `git push --follow-tags`, looks like it might be pushing the tags from previous releases as well

https://stackoverflow.com/a/3745250/3511012

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@kitbag/router",
-  "version": "0.4.0",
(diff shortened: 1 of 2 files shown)


pleek91:
I believe you'll want to redo this after your other PR is merged to main. Tags reference specific commits. So I bet your tag here is referencing the current HEAD of main and won't include the latest commits when merged.

PR #206, opened 2024-06-22
Fixed bug with host route match rule

stackoverfloweth:
I discovered bug while trying to update our stackblitz to latest version (currently on 0.2.0). Initial routes were not passing `routeHostMatches` because url has `localhost:5173` for host and route would have `undefined`. 

This PR updates the logic for host match rule to treat a route without host as automatically relative and passes host check.

At some point as we expand support for external routes we can expect route to have host, however `undefined` will continue to be assumed as automatically relative so this change will still apply.

### DIFF
--- a/src/services/createRouterFind.spec.ts
+++ b/src/services/createRouterFind.spec.ts
@@ -41,8 +41,7 @@ test('when given a url that does not match a route returns undefined', () => {
   expect(route).toBeUndefined()
 })
 
-// until route has host this test is not useful
-test.fails('when given an external url that does not match a route returns undefined', () => {
+test('when given an external url that does not match a route returns undefined', () => {
   const find = createRouterFind(routes)
   const route = find('https://example.com')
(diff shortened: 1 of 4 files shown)


src/services/createRouterFind.spec.ts
```diff
@@ -41,7 +41,8 @@ test('when given a url that does not match a route returns undefined', () => {
   expect(route).toBeUndefined()
 })
 
-test('when given an external url that does not match a route returns undefined', () => {
+// until route has host this test is not useful
+test.fails('when given an external url that does not match a route returns undefined', () => {
```
pleek91, line 45: This doesn't seem right. Because router.push supports any url right?

stackoverfloweth: okay, reverted back to stripping host from initialUrl then

pleek91 (approved):

PR #208, opened 2024-06-24
optional params alt approach

stackoverfloweth:
Alternative approach for tracking optional params without modifying param itself

With this PR, knowing if a param should be optional has nothing to do with the `Param` itself. We've always taken the name of a param from the string part of a path/query. IE `path('/something/[id]', {id: String})` for required string or `path('/something/[?id]', {id: String})` for optional string of the same name. We always strip the param name out from the `[` and `]`, we just now keep the `?` if it's present. 

Previously our optional id param would be stored on the route as 

```ts
path: {
  ...,
  params: {
    requiredFoo: String,
    optionalBar: optional(String),
  }
}
```

Now will be stored as 

```ts
path: {
  ...,
  params: {
    requiredFoo: String,
    '?optionalBar': String,
  }
}
```

Note, this key that now may start with "?" is only true internally. So when calling `path` or `query`, the 2nd argument does not expect a question mark

```
// correct
path('/something/[?id]', {id: String})

// incorrect
path('/something/[?id]', {'?id': String})
```

This is also true for `router.push` type functions as well as reading params from `useRoute`.

```
const route = useRoute('myRoute')

// correct
route.params.id

// incorrect
route.params['?id']
```

This alternative approach means we can delete the entire `optional` function, types, and file altogether.


src/services/combinePath.ts
```diff
@@ -22,5 +21,11 @@ export function combinePath(parentPath: Path, childPath: Path): Path {
...
   }
 
-  return path(`${parentPath.path}${childPath.path}`, { ...parentPath.params, ...childPath.params })
+  const newPathString = `${parentPath.path}${childPath.path}`
+
+  return {
+    path: newPathString,
+    params: { ...parentPath.params, ...childPath.params },
+    toString: () => newPathString,
+  }
```
stackoverfloweth, line 30: when combining parent path and child path, we used to use `path` again. The problem with this approach is the parent `path` has already been called, (child might have been too if done so by user). When calling `path`, the params are expected to be without `?`. This allows developers to call path as expected `path('/something/[?id]', {id: Number})` note the `id` in params doesn't expect `?`. Without this change, the nested `path` call would lose `Number` type because `path.params`  after being called initially would be `{ '?id': Number }`. 

Alternatively, if you don't like this we can have path and query both look for with and without `?` since we don't allow for params of the same name, even if one has `?` and the other doesn't.

src/services/combineQuery.ts
```diff
@@ -33,5 +32,9 @@ export function combineQuery(parentQuery: Query, childQuery: Query): Query {
     .filter(stringHasValue)
     .join('&')
 
-  return query(newQueryString, { ...parentQuery.params, ...childQuery.params })
+  return {
+    query: newQueryString,
+    params: { ...parentQuery.params, ...childQuery.params },
+    toString: () => newQueryString,
+  }
```
stackoverfloweth, line 39: same deal here as what I said WRT [combinePath](https://github.com/kitbagjs/router/pull/208/files#diff-bd0b17ec6a10ab0c97cbfaffc09c5aff16dfcb669dc0e91cbe713880958ef103)

src/services/getParamsForString.ts
```diff
@@ -1,22 +1,27 @@
 import { DuplicateParamsError } from '@/errors'
-import { optional } from '@/services/optional'
 import { getParam } from '@/services/params'
+import { getParamName } from '@/services/routeRegex'
 import { paramEnd, paramStart } from '@/types/params'
 import { Param } from '@/types/paramTypes'
 
 export function getParamsForString<TInput extends string, TParams extends Record<string, Param | undefined>>(string: TInput, params: TParams): Record<string, Param> {
-  const paramPattern = new RegExp(`\\${paramStart}\\??([\\w-_]+)\\${paramEnd}`, 'g')
+  const paramPattern = new RegExp(`\\${paramStart}(\\??[\\w-_]+)\\${paramEnd}`, 'g')
```
stackoverfloweth, line 8: this is important change, moves the capture group back from grabbing name without `?` to grabbing name with `?`

src/services/getParamsForString.ts
```diff
@@ -1,22 +1,27 @@
...
   const matches = Array.from(string.matchAll(paramPattern))
 
-  return matches.reduce<Record<string, Param>>((value, [match, paramName]) => {
-    const isOptional = match.startsWith(`${paramStart}?`)
+  return matches.reduce<Record<string, Param>>((value, [match, key]) => {
+    const paramName = getParamName(match)
+
+    if (!paramName) {
+      return value
+    }
```
stackoverfloweth, line 16: we don't expect `paramName` to actually be undefined, that utility existing before this PR and returns `undefined` if the input doesn't match expected syntax

src/services/getParamsForString.ts
```diff
@@ -1,22 +1,27 @@
...
 
 export function getParamsForString<TInput extends string, TParams extends Record<string, Param | undefined>>(string: TInput, params: TParams): Record<string, Param> {
-  const paramPattern = new RegExp(`\\${paramStart}\\??([\\w-_]+)\\${paramEnd}`, 'g')
+  const paramPattern = new RegExp(`\\${paramStart}(\\??[\\w-_]+)\\${paramEnd}`, 'g')
   const matches = Array.from(string.matchAll(paramPattern))
 
-  return matches.reduce<Record<string, Param>>((value, [match, paramName]) => {
-    const isOptional = match.startsWith(`${paramStart}?`)
+  return matches.reduce<Record<string, Param>>((value, [match, key]) => {
+    const paramName = getParamName(match)
```
stackoverfloweth, line 12: note here, paramName is without `?`, key is with (if optional)

`paramName` => `foo`
`key` => `?foo`

src/services/paramValidation.ts
```diff
@@ -30,10 +30,12 @@ function getPathParams(path: Path, url: string): Record<string, unknown> {
   const decodedValueFromUrl = decodeURIComponent(url)
 
   for (const [key, param] of Object.entries(path.params)) {
+    const isOptional = key.startsWith('?')
+    const paramName = isOptional ? key.slice(1) : key
     const stringValue = getParamValueFromUrl(decodedValueFromUrl, path.toString(), key)
-    const formattedValues = getParamValue(stringValue, param)
+    const formattedValues = getParamValue(stringValue, param, isOptional)
```
stackoverfloweth, line 36: previously the `param` argument was modified to include context of if value is optional. Now that logic exists outside of the `param`, so I added an optional 3rd argument to `getParamValue` and `setParamValue`.

src/services/urlAssembly.ts
stackoverfloweth, line 1: much of these changes are just changing args to pass entire `path` and `query` into functions below. This lets us pass the keys with possible `?`.

src/types/params.ts
```diff
@@ -46,7 +45,7 @@ export function isParamGetSet(value: Param): value is ParamGetSet {
  * @returns The extracted parameter name, or never if the parameter string is empty.
  */
 export type ExtractParamName<
-  TParam extends string
+  TParam extends PropertyKey
```
stackoverfloweth, line 48: this just makes it more convenient to use the utility, doesn't change functionality

src/types/params.ts
```diff
@@ -93,7 +90,7 @@ export type ExtractRouteParamTypes<TRoute> = TRoute extends {
  * @returns A new type with the appropriate properties marked as optional.
  */
 export type ExtractParamTypes<TParams extends Record<string, Param>> = Identity<MakeOptional<{
-  [K in keyof TParams]: ExtractParamType<TParams[K]>
+  [K in keyof TParams as ExtractParamName<K>]: ExtractParamType<TParams[K], K>
```
stackoverfloweth, line 93: this does 2 important things.
- it changes the key to be the name of the param without `?` for devs
- it passes in the unmodified `K` to be used by `ExtractParamType` to determine if type should be optional

src/types/params.ts
```diff
@@ -102,15 +99,19 @@ export type ExtractParamTypes<TParams extends Record<string, Param>> = Identity<
...
     : Type
   : TParam extends ParamGetter
-    ? ReturnType<TParam>
-    : string
+    ? TParamKey extends `?${string}`
+      ? ReturnType<TParam> | undefined
+      : ReturnType<TParam>
+    : TParamKey extends `?${string}`
+      ? string | undefined
+      : string
```
stackoverfloweth, line 114: previously, optional params could only be `ParamGetSet`, after this PR any param could be optional

src/types/path.ts
```diff
@@ -8,11 +8,11 @@ type ExtractParamsFromPathString<
   TPath extends string,
   TParams extends Record<string, Param | undefined> = Record<never, never>
 > = TPath extends `${string}${ParamStart}${infer Param}${ParamEnd}${infer Rest}`
-  ? MergeParams<{ [P in ExtractParamName<Param>]: ExtractPathParamType<Param, TParams> }, ExtractParamsFromPathString<Rest, TParams>>
+  ? MergeParams<{ [P in Param]: ExtractPathParamType<Param, TParams> }, ExtractParamsFromPathString<Rest, TParams>>
```
stackoverfloweth, line 11: params for `path` no longer use `ExtractParamName` because we want to keep `?`

src/types/path.ts
```diff
@@ -8,11 +8,11 @@ type ExtractParamsFromPathString<
...
   : Record<never, never>
 
 export type PathParams<TPath extends string> = {
-  [K in keyof ExtractParamsFromPathString<TPath>]?: Param
+  [K in ExtractParamName<keyof ExtractParamsFromPathString<TPath>>]?: Param
```
stackoverfloweth, line 15: this is what drives the 2nd argument for `path`, should remove `?` from param keys

src/types/query.ts
```diff
@@ -8,11 +8,11 @@ type ExtractQueryParamsFromQueryString<
   TQuery extends string,
   TParams extends Record<string, Param | undefined> = Record<never, never>
 > = TQuery extends `${string}=${ParamStart}${infer Param}${ParamEnd}${infer Rest}`
-  ? MergeParams<{ [P in ExtractParamName<Param>]: ExtractPathParamType<Param, TParams> }, ExtractQueryParamsFromQueryString<Rest, TParams>>
+  ? MergeParams<{ [P in Param]: ExtractPathParamType<Param, TParams> }, ExtractQueryParamsFromQueryString<Rest, TParams>>
   : Record<never, never>
 
 export type QueryParams<T extends string> = {
-  [K in keyof ExtractQueryParamsFromQueryString<T>]?: Param
+  [K in ExtractParamName<keyof ExtractQueryParamsFromQueryString<T>>]?: Param
```
stackoverfloweth, line 15: same changes as [types/path](https://github.com/kitbagjs/router/pull/208/files#diff-962fcd366258587d5d35e33171487e404bd77362b7fbb787a86ef4cb14359231R11)

src/types/routeWithParams.ts
```diff
@@ -21,13 +20,17 @@ type ExtractRouteParamTypesWithoutLosingOptional<TRoute> = TRoute extends {
...
   : TParam extends ParamGetter
-    ? ReturnType<TParam>
-    : string
\ No newline at end of file
+    ? TParamKey extends `?${string}`
+      ? ReturnType<TParam> |undefined
+      : ReturnType<TParam>
+    : TParamKey extends `?${string}`
+      ? string | undefined
+      : string
```
stackoverfloweth, line 36: same changes as [types/params](https://github.com/kitbagjs/router/pull/208/files#diff-1666f5b0d3d38b2c1add488147cfe85eebfbf7f6025dbe51a070c44d08a0f41dR93).

src/utilities/checkDuplicateKeys.ts
```diff
@@ -1,6 +1,6 @@
 export function checkDuplicateKeys(aParams: Record<string, unknown>, bParams: Record<string, unknown>): { key: string, hasDuplicates: true } | { key: undefined, hasDuplicates: false } {
-  const aParamKeys = Object.keys(aParams)
-  const bParamKeys = Object.keys(bParams)
+  const aParamKeys = Object.keys(aParams).map(removeLeadingQuestionMark)
+  const bParamKeys = Object.keys(bParams).map(removeLeadingQuestionMark)
```
stackoverfloweth, line 3: don't love this, but I think it's necessary because technically params could theoretically be stored with and without the `?`. However, I've made the decision to keep our rule of not duplicate param names.

src/services/path.ts
```diff
@@ -25,10 +26,11 @@ import { Identity } from '@/types/utilities'
  * ])
  * ```
  */
-export function path<TPath extends string, TParams extends PathParams<TPath>>(path: TPath, params: Identity<TParams>): Path<TPath, TParams> {
+export function path<TPath extends string, TParams extends PathParamsWithParamNameExtracted<TPath>>(path: TPath, params: Identity<TParams>): Path<TPath, TParams>
+export function path(path: string, params: Record<string, Param | undefined>): Path {
```
stackoverfloweth, line 30: using this overload removes the need for type casting params

src/services/query.ts
```diff
@@ -25,10 +26,11 @@ import { Identity } from '@/types/utilities'
  * ])
  * ```
  */
-export function query<TQuery extends string, TParams extends QueryParams<TQuery>>(query: TQuery, params: Identity<TParams>): Query<TQuery, TParams> {
+export function query<TQuery extends string, TParams extends QueryParamsWithParamNameExtracted<TQuery>>(query: TQuery, params: Identity<TParams>): Query<TQuery, TParams>
+export function query(query: string, params: Record<string, Param | undefined>): Query {
```
stackoverfloweth, line 30: using this overload removes the need for type casting params

pleek91 (approved):

PR #209, opened 2024-06-25
Make `route` a union of all possible routes and improve return types for `isRoute` and `useRoute`

pleek91:
# Description
This PR updates the type for the current route to be a discriminating union of possible routes. This means that when accessing the route via `router.route` or by using the `useRoute` composition the route can easily be narrowed. 

Given the following routes
```ts
const routes = createRoutes([
  {
    name: 'parentA',
    path: '/parentA',
    children: createRoutes([
      {
        component,
        name: 'childA',
        path: '/childA/[paramA]',
      },
    ]),
  },
  {
    name: 'parentB',
    path: '/parentB',
    component,
  },
])

```
`router.route.key` is type `'parentA' | 'parentB' | 'parentA.childA'`. And `router.route.params` is type `{} | { paramA: string }`. Which means you can narrow the route in a type safe way by doing things like this
```ts
const route = router.route

if(route.key === 'parentA.childA') {
  console.log(route.params.paramA) // correctly knows that `paramA` is a valid param
}
```
## `isRoute` and `useRoute`
This gets even more useful when combined with the `isRoute` type guard that is exported by router. By default `isRoute` is not an exact route match, meaning it will match any route AND any children of that route.
```ts
if(isRoute(route, 'parentA')) {
  // route.key is type 'parentA' | 'parentA.childA'
}
```
If you want to match only an exact route you can pass the exact option into `isRoute`
```ts
if(isRoute(route, 'parentA', { exact: true })) {
  // route.key is type 'parentA'
}
```
This works exactly the same when using `useRoute`
```ts
const route = useRoute('parentA')
// route.key is 'parentA' | 'parentA.childA'
```
```ts
const route = useRoute('parentA', { exact: true })
// route.key is 'parentA'
```
This makes working with routes in a type safe way much easier and more flexible.

### DIFF
--- a/src/components/routerView.browser.spec.ts
+++ b/src/components/routerView.browser.spec.ts
@@ -5,6 +5,7 @@ import helloWorld from '@/components/helloWorld'
 import { createRouter } from '@/services/createRouter'
 import { createRoutes } from '@/services/createRoutes'
 import { isRouteWithComponent } from '@/types/routeProps'
+import { routes } from '@/utilities/testHelpers'
 
 test('renders component for initial route', async () => {
   const routes = createRoutes([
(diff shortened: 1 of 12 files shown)


stackoverfloweth (approved):
❤️

PR #211, opened 2024-06-29
Update docs to include a guide about narrowing the route

pleek91:
# Description
Little cleanup of docs in general. 

- Added the missing "Composables" page to core concepts (previously only accessible via search and was listed as an advanced concept). 
- Added "Route Narrowing" to advanced concepts
- Updated api docs for `isRoute` and `useRoute`
- Removed some references to "route methods" that were outdated

### DIFF
--- a/docs/advanced-concepts/route-narrowing.md
+++ b/docs/advanced-concepts/route-narrowing.md
@@ -1,0 +1,71 @@
+# Route Narrowing
+When accessing the current route, by default the type is a union of all possible routes. 
+
+```ts
+import { createRoutes, createRouter } from '@kitbag/router'
+
+const routes = createRoutes([
+  {
+    name: 'home',
+    path: '/',
+    component: ...,
+  },
+  {
+    name: 'user',
+    path: '/user/[userId]',
+    component: ...,
+    children: createRoutes([
+      {
+        name: 'profile',
+        path: '/profile',
+        query: '?tab=[tab]',
+        component: ...,
+      },
+      {
+        name: 'settings',
+        path: '/settings',
+        component: ...,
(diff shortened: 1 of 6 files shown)


docs/advanced-concepts/route-narrowing.md
```diff
@@ -0,0 +1,71 @@
...
+    ])
+  }
+])
+
+const router = createRouter(routes)
+
+router.route.key // "home" | "user" | "user.profile" | "user.settings"
+```
+
+This can be narrowed by using the route key to check which route is the current route.
```
stackoverfloweth, line 38:
```suggestion
This can be narrowed like any union in Typescript, by checking the route key.
```

docs/advanced-concepts/route-narrowing.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+The `exact` option is optional and defaults to `false`.
+
+```ts
+if(router.route, 'user') {
+  router.route.key // "user" | "user.profile" | "user.settings"
+  router.route.params // { userId: string } | { userId: string, tab: string }
+}
+```
```
stackoverfloweth, line 71:
```suggestion
The `isRoute` type guard offers more flexibility with the optional `exact` argument, which defaults to `false` and will return narrow to the target route or any of it's descendants. 

```ts
if(router.route, 'user', { exact: false }) {
  router.route.key // "user" | "user.profile" | "user.settings"
  router.route.params // { userId: string } | { userId: string, tab: string }
```
```

stackoverfloweth (approved):
couple of suggestions on the route-narrowing page

PR #213, opened 2024-07-04
Router add host context

stackoverfloweth:
This PR establishes direct responsibility for tracking host on router. This context will be used to determine if subsequent route changes or router-links are "external", meaning of a different host.

The router will now check the `initialUrl` to see if it includes host context.

```ts
const router = createRouter({
  initialUrl: 'https://router.kitbag/dev/my-home-route',
})
```

As a result of this PR every way of changing the routers location will safely check if the target url is within the same target domain. If it's not and the router will skip hooks and simply call `history.update`. 

There's an outstanding suggestion to check if the history mode is memory and throw an error if an attempt to update is called with an external host. I would like to discuss this a bit more, possibly implement in a follow-up PR. 

This PR also adds `isExternal` property to `router`, which is used by `router-link` to determine `isExternal` flag but is also available to end users to compare a given url to what's stored in router context.


stackoverfloweth:
I will also need a follow-up PR that updates documentation if this gets approved

src/services/createIsSameHost.ts
```diff
@@ -0,0 +1,16 @@
+import { createMaybeRelativeUrl } from '@/services/createMaybeRelativeUrl'
+
+export function createIsSameHost(host: string | undefined): (url: string) => boolean {
```
pleek91, line 3: What do you think about this being "createIsExternal"? I think generally that's what we want to know and it might be nice to carry that logic through to routes and router-link.

src/types/router.ts
```diff
@@ -17,7 +17,7 @@ export type RouterReject = (type: RouterRejectionType) => void
  */
 export type RouterOptions = {
   /**
-   * Initial URL for the router to use. Required if using Node environment.
+   * Initial URL for the router to use. Required if using Node environment. Defaults to window.location is using browser.
```
pleek91, line 20:
```suggestion
   * Initial URL for the router to use. Required if using Node environment. Defaults to window.location when using browser.
```

src/types/router.ts
```diff
@@ -26,14 +26,18 @@ export type RouterOptions = {
    * Specifies the history mode for the router, such as 'hash', 'history', or 'abstract'.
    */
   historyMode?: RouterHistoryMode,
+  /**
+   * Specifies the desired host for the router, defaults to host of initialUrl.
+   */
+  host?: string,
```
pleek91, line 32: I don't feel strongly here but thinking if you can use this or initialUrl maybe we just use one? Fine as is. Just thinking out loud.

pleek91 (approved):

src/types/router.ts
stackoverfloweth: yeah, I liked the explicit nature of having a separate prop but I agree that the fact that it's duplicative and potentially might disagree means we should probably kill it.

src/services/createIsExternal.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expect, test } from 'vitest'
+import { createIsExternal } from '@/services/createIsExternal'
+
+test('given undefined host, returns false', () => {
```
pleek91, line 4: This test seems backwards to me. If host is undefined. Any url with a host (anything non relative) should be considered external right?

stackoverfloweth: I guess my thinking is if you don't supply the context we need to determine if it's external, we always default to the more likely case. Which IMO is that it's not external. Being external also has deeper implications on route changes so I figured it was better to let the router try to handle it like any other route.

stackoverfloweth: synced up on this, agreed to change to defaulting to external because presence of error is more helpful than going to not found

pleek91 (approved):

src/services/createRouter.ts
```diff
@@ -70,6 +72,10 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
   async function set(url: string, { replace }: RouterUpdateOptions = {}): Promise<void> {
     history.stopListening()
 
+    if (isExternal(url)) {
+      return history.update(url, { replace })
+    }
```
pleek91, line 77: For follow up later. We should think about how hooks apply when navigating to an external url. We could run the leave hooks of the current url before calling update.

src/services/createRouter.ts
```diff
@@ -192,6 +200,7 @@ export function createRouter<const T extends Routes>(routes: T, options: RouterO
     go: history.go,
     install,
     initialized,
+    isExternal,
```
pleek91, line 203: Do we need to expose this?

stackoverfloweth: I think it's valuable enough, what do you think?

pleek91: I don't feel strongly. But it's one more thing we have to support publicly.

stackoverfloweth: Do you have a good way to hide it? Are you suggesting we just drop it from the type?

PR #214, opened 2024-07-07
Add support for defined external routes

pleek91:
# Description
Introduced a new `createExternalRoutes` utility for defining external routes. Router already supports url strings (either relative or fully qualified) when pushing to the route or for the `to` prop of `router-link`. Using the `createExternalRoutes` you can define routes that produce url strings. 

```ts
const routes = createExternalRoutes([
  {
    host: 'https://router.kitbag.dev/',
    path: path('/api/components/[component]', {
      component: String
    })
  }
])

```


stackoverfloweth (approved):
seems so simple now, well done 👏

PR #215, opened 2024-07-07
Visual learners rejoice 🎉 

stackoverfloweth:
Primary objective here is just to increase likelihood someone goes from docs/readme to actually installing. We have a lot of good content in docs but needs better support for visual learners. This PR introduces a new "Show Me" page, which walks through

- creating routes and router
- installing and registering router
- using useRouter and router.push
- using RouterView and RouterLink components
- adding params, optional, and typed

this PR also 

- sets up [lazy loading](https://vitepress.dev/guide/markdown#image-lazy-loading) for images by default
- puts "show me" page in side nav at the top above "introduction"
- makes "Show Me" the new landing page for "getting started" button
- adds what I think is the most compelling gif, route-params to readme

In retrospect I wish I recorded these in a simpler, higher contrast theme. I added that to my backlog, open to other suggestions if I'm going to re-record these clips/more clips.


docs/show-me.md
```diff
@@ -0,0 +1,31 @@
+# Show Me
+
+## Create Routes and Router
+
+In this clip, we're going to create (2) top level routes `home` and `settings`. Then we're going to add (2) additional routes under `settings`. Finally, we will create and export our `router`.
```
pleek91, line 5:
```suggestion
First, let's create two top level routes `home` and `settings`. Then we're going to add two additional routes under `settings`. Finally, we will create and export our `router`.
```

docs/show-me.md
```diff
@@ -0,0 +1,31 @@
...
+
+![Create Routes and Router](/create-routes.gif)
+
+## Installing the Router
+
+In this clip, we take the router we install the router and setup the Type Safety in components with the `Register` interface. Read more about [updating the registered router](/getting-started#update-registered-router).
```
pleek91, line 11:
```suggestion
Next, we install the router and register the router type using the `Register` interface. Read more about [updating the registered router](/getting-started#update-registered-router).
```

docs/show-me.md
```diff
@@ -0,0 +1,31 @@
...
+
+![Installing the Router](/install-router.gif)
+
+## Navigating with Router Push
+
+In this clip, we call `useRouter` in setup of `HomeView` and call `router.push` when the button is clicked. Notice how all (4) of our routes show up in autocomplete. Typescript will throw an error if I use a value other than a valid route key or a fully formed external URL. Read more about [router push](/core-concepts/navigating#push).
```
pleek91, line 17:
```suggestion
Now we can use the `useRouter` composable in the setup of the `HomeView` component. We can then call `router.push` when the button is clicked. Notice how all four of our routes show up in autocomplete. Typescript will throw an error if a value other than a valid route key or a fully formed external URL. Read more about [router push](/core-concepts/navigating#push).
```

docs/show-me.md
```diff
@@ -0,0 +1,31 @@
...
+
+![Navigating with Router Push](/router-push.gif)
+
+## RouterView and RouterLink
+
+In this clip, we add `RouterView` component to `SettingsView`, since it has children. We also add a couple `RouterLink` components to the navbar to give the user links to those child pages. Notice how the `to` callback syntax is also type safe and autocompletes all possible routes for us.
```
pleek91, line 23:
```suggestion
We need to add `RouterView` component to `SettingsView`, since it has children. We'll also add a couple `RouterLink` components to the navbar to give the user links to those child pages. Notice how the `to` callback syntax is also type safe and autocompletes all possible routes for us.
```

docs/show-me.md
```diff
@@ -0,0 +1,31 @@
...
+
+![RouterView and RouterLink](/router-link.gif)
+
+## Adding Params
+
+In this clip, we're going to add a param to our `settings.keys` route. Notice how as soon as I save the route with `[id]` in the path, I get a Typescript error where I have a router-link in the `SettingsView` tab. Which I can satisfy by providing a value for that required param. Alternatively I can add (?) to the param, making it optional `[?id]`. Finally, I'll pull in the `path` utility to change the expected type for `id` from string to number. Notice how this too creates Typescript errors until I provide a valid value of the correct type. Read more about [route params](/core-concepts/route-params#route-params).
```
pleek91, line 29:
```suggestion
Let's add a param to our `settings.keys` route. Notice how as soon as I save the route with `[id]` in the path, I get a Typescript error where I have a router-link in the `SettingsView` tab. Which I can satisfy by providing a value for that required param. Alternatively I can add (?) to the param, making it optional `[?id]`. Finally, I'll pull in the `path` utility to change the expected type for `id` from string to number. Notice how this too creates Typescript errors until I provide a valid value of the correct type. Read more about [route params](/core-concepts/route-params#route-params).
```

pleek91 (approved):
Nit suggestions around voice and trying to simplify a bit. Approved as is.

PR #216, opened 2024-07-07
createRouter accepts array of route arrays

stackoverfloweth:
adds overload to createRouter that supports passing an array of route arrays

```ts
const aRoutes = createRoutes([
  { name: 'a-route-1', path: '/a-route-1', component },
  { name: 'a-route-2', path: '/a-route-2', component },
])
const bRoutes = createRoutes([
  { name: 'b-route-1', path: '/b-route-1', component },
  { name: 'b-route-2', path: '/b-route-2', component },
])

createRouter([aRoutes, bRoutes])
```

This will be extremely useful for devs now that they have access to `createExternalRoutes` as well as `createRoutes`.


src/services/createRouter.ts
```diff
@@ -47,7 +48,10 @@ type RouterUpdateOptions = {
  * const router = createRouter(routes)
  * ```
  */
-export function createRouter<const T extends Routes>(routes: T, options: RouterOptions = {}): Router<T> {
+export function createRouter<const T extends Routes>(routes: T, options?: RouterOptions): Router<T>
+export function createRouter<const T extends Routes>(arrayOfRoutes: T[], options?: RouterOptions): Router<T>
```
pleek91, line 52: I'm surprised we don't need to flatten this. But I suppose T is a union so we get a union of routes on the router?

src/utilities/guards.ts
```diff
@@ -19,3 +19,7 @@ export function hasProperty<
 
   return true
 }
+
+export function isNestedArray<T extends any[] | readonly any[]>(value: T | T[]): value is T[] {
+  return value.length > 0 && Array.isArray(value[0])
```
pleek91, line 24: Should this maybe be `value.every(item => Array.isArray(item))`

stackoverfloweth: I suppose that's perhaps technically more accurate given that this is a utility, but in the context of createRouter, if the 1st item in the array is an array, they are all going to be. I'd prefer not to iterate through potentially a large set to prove what Typescript types should have already enforced?

src/services/createRouter.ts
stackoverfloweth: I agree, I was pleasantly surprised but it definitely does get correctly merged types

image

src/utilities/guards.ts
pleek91: I don't feel strongly. But it's probably a small array and it reads better IMO. Merge away as is 👍

src/services/createRouter.ts
pleek91: Nice, simpler than I imagined.

pleek91 (approved):

PR #222, opened 2024-07-13
Route match updates

stackoverfloweth:
this PR makes 2 changes to route matching logic
- when sorting possible matches, route depth matters more than number of optional params satisfied
- removed host match rule, since checking for external routes now happens before route matching rules are executed


src/services/createRouterFind.spec.ts
```diff
@@ -40,10 +40,3 @@ test('when given a url that does not match a route returns undefined', () => {
 
   expect(route).toBeUndefined()
 })
-
-test('when given an external url that does not match a route returns undefined', () => {
-  const find = createRouterFind(routes)
-  const route = find('https://example.com')
-
-  expect(route).toBeUndefined()
-})
```
pleek91, line 49: Is this no longer true?

pleek91 (approved):

src/services/createRouterFind.spec.ts
stackoverfloweth: correct, because the route matching rules no longer care about host so this url looks like just an empty string, which matches like 4 different routes

pleek91: Well that's a problem then right? If a user runs with they'll get a route that doesn't match the url they passed in.

stackoverfloweth: I guess? we have that external check within the router when actually setting the current route, but no such check for `router.resolve` or `router.find`. Should we move/duplicate that check deeper?

src/services/createRouter.ts
```diff
@@ -168,7 +168,25 @@ export function createRouter<const T extends Routes>(routesOrArrayOfRoutes: T |
...
+      return getResolvedRouteForUrl(routes, url)
+    }
+
+    if (isExternal(source)) {
+      return undefined
+    }
+
+    const url = resolve(source)
+
+    return getResolvedRouteForUrl(routes, url)
```
pleek91, line 187: I'm assuming this is how it was already implemented. But isn't `source` already a string at this point? Because the previous if statements.

stackoverfloweth: yeah... you said `string`, but it's actually `Url`. Not sure if sending that through `resolve` does something or not 🤔

pleek91: Seems like we could skip the resolve step here and pass it right into getResolvedRouteForUrl but 🤷‍♂️

stackoverfloweth: removed

PR #223, opened 2024-07-18
added host utility, similar to path and query

stackoverfloweth:
This PR enables the same interpolation pattern with param support the the `host` part of a URL that we already support for `path` and `query`. 

Usage is exactly what you'd expect, you can define params as a plain string or with a new `host` utility.

```ts
const routes = createExternalRoutes([
  name: 'foo',
  host: 'https://[subdomain].kitbag.dev',
  path: '/'
])
```

```ts
const routes = createExternalRoutes([
  name: 'foo',
  host: host('https://[subdomain].kitbag.dev', {subdomain: availableSubdomainsParam}),
  path: '/'
])
```

as with other params
- they can be optional
- they support all the same type for params
- they can be passed when calling `router.resolve`, `router-link`, `router.push`, etc.

### DIFF
--- a/src/services/host.spec.ts
+++ b/src/services/host.spec.ts
@@ -1,0 +1,41 @@
+import { expect, test } from 'vitest'
+import { DuplicateParamsError } from '@/errors'
+import { host } from '@/services/host'
+
+test('given host without params, returns empty object', () => {
+  const response = host('https://kitbag.dev', {})
+
+  expect(response.params).toMatchObject({})
+})
+
+test('given host with simple params, returns each param name as type String', () => {
+  const response = host('https://[subdomain]kitbag.dev', {})
+
+  expect(response.params).toMatchObject({
+    subdomain: String,
+  })
+})
+
+test('given host with optional params, returns each param name as type String', () => {
+  const response = host('https://[?subdomain]kitbag.dev', {})
+
+  expect(JSON.stringify(response.params)).toMatch(JSON.stringify({
+    subdomain: String,
+  }))
+})
+
+test('given host with param types, returns each param with corresponding param', () => {
+  const response = host('https://[subdomain]kitbag.dev', {
+    subdomain: Boolean,
+  })
+
+  expect(response.params).toMatchObject({
+    subdomain: Boolean,
+  })
+})
+
+test('given host with the same param name, throws DuplicateParamsError', () => {
+  const action: () => void = () => host('https://[?subdomain]kitbag.[subdomain]', { })
+
+  expect(action).toThrowError(DuplicateParamsError)
+})
--- a/src/services/host.spec.ts
+++ b/src/services/host.spec.ts
@@ -3,13 +3,13 @@ import { DuplicateParamsError } from '@/errors'
 import { host } from '@/services/host'
 
 test('given host without params, returns empty object', () => {
-  const response = host('https://kitbag.dev', {})
+  const response = host('kitbag.dev', {})
 
   expect(response.params).toMatchObject({})
 })
 
 test('given host with simple params, returns each param name as type String', () => {
-  const response = host('https://[subdomain]kitbag.dev', {})
+  const response = host('[subdomain]kitbag.dev', {})
 
   expect(response.params).toMatchObject({
     subdomain: String,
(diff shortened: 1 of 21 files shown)


src/services/host.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import { host } from '@/services/host'
+
+test('given host without params, returns empty object', () => {
+  const response = host('https://kitbag.dev', {})
+
+  expect(response.params).toMatchObject({})
+})
+
+test('given host with simple params, returns each param name as type String', () => {
+  const response = host('https://[subdomain]kitbag.dev', {})
```
pleek91, line 12:
```suggestion
  const response = host('https://[subdomain].kitbag.dev', {})
```

src/services/urlAssembly.spec.ts
```diff
@@ -293,4 +295,117 @@ describe('static query', () => {
...
+
+    const url = assembleUrl(route)
+
+    expect(url).toBe('https://kitbag.dev/')
+  })
+
+  test.each([
+    ['https://[?subdomain]kitbag.dev'],
+    [host('https://[?subdomain]kitbag.dev', { subdomain: String })],
+    [host('https://[?subdomain]kitbag.dev', { subdomain: withDefault(String, 'abc') })],
```
pleek91, line 322: These having the period missing makes sense. But it also makes me think about devx for this sort of thing. 

I wonder if we could allow users to specify a period or slash is optional without using  param.

src/services/urlAssembly.ts
```diff
@@ -14,16 +15,29 @@ type AssembleUrlOptions = {
...
+  return withQuery(`${hostWithParamsSet}${pathWithParamsSet}`, queryWithParamsSet, queryValues)
+}
+
+function assembleHostParamValues(host: Host, paramValues: Record<string, unknown>): string {
+  const value = host.toString()
 
-  if (route.host) {
-    return `${route.host}${relativeUrl}`
+  if (!value.length) {
+    return value
```
pleek91, line 29: Do we even need this check?

src/types/externalRouteProps.ts
```diff
@@ -1,9 +1,10 @@
+import { Host } from '@/types/host'
 import { Path } from '@/types/path'
 import { Query } from '@/types/query'
 import { Route } from '@/types/route'
 
 // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
-export type ExternalRoutes = Route<string, '', Path, Query, boolean>[]
+export type ExternalRoutes = Route<string, Host, Path, Query, boolean>[]
```
pleek91, line 7: I believe this should be 
```suggestion
export type ExternalRoutes = Route<string, Host<''>, Path, Query, boolean>[]
```

It's restricting children in external routes from having a host. This type should really be called `ExternalChildren` or something.

src/types/host.ts
```diff
@@ -0,0 +1,44 @@
+import { host } from '@/services/host'
+import { ExtractParamName, ExtractPathParamType, MergeParams, ParamEnd, ParamStart } from '@/types/params'
+import { Param } from '@/types/paramTypes'
+import { Identity } from '@/types/utilities'
+import { isRecord } from '@/utilities/guards'
+
+type ExtractParamsFromHostString<
```
pleek91, line 7: These types are the same as the Path types right?

src/types/register.ts
```diff
@@ -38,7 +39,7 @@ export type RegisteredRouterRoute = RegisteredRouter['route']
  */
 export type RegisteredRoutes = Register extends { router: Router<infer TRoutes extends Routes> }
   ? TRoutes
-  : Route<string, '', Path<'', {}>, Query<'', {}>, false>[]
+  : Route<string, Host, Path<'', {}>, Query<'', {}>, false>[]
```
pleek91, line 42: If we don't need the empty string for host why do we need it for Path and Query?

```suggestion
  : Route<string, Host<''>, Path<'', {}>, Query<'', {}>, false>[]
```

pleek91 (commented):
Love this. Just the consistency alone is worth it. Couple questions but overall LGTM

src/types/externalRouteProps.ts
stackoverfloweth: solved this together

src/types/register.ts
stackoverfloweth: I don't think we needed it on Path and Query

pleek91 (approved):

PR #224, opened 2024-07-20
should host params be included here?

stackoverfloweth:
I made this change because it felt obvious, but then after thinking about it some more I'm not sure. 

This utility is only used by `getResolvedRouteForUrl`. The `getResolvedRouteForUrl` utility is used in router for 2 things, to get the `to` for hooks and inside `router.find`. Both of which currently have escapes for situations where it's given an external route. 

I feel like it SHOULD be added, but maybe those places that use `getResolvedRouteForUrl` shouldn't just escape? If a user calls `router.find(https://my.example.com/clearly/external)`, maybe it SHOULD return the matched external route? Same for the `to` on hooks, it would be handy for executing before leave hooks to know you're about to go to your fancy external route?


stackoverfloweth:
for now, we aren't planning on adding matching functionality to external routes

PR #227, opened 2024-07-21
make path optional

stackoverfloweth:
Not sure why it took this long but I think `path` should have always been optional. This felt especially true when defining parent routes, or external routes

```ts
createExternalRoutes([
  {
    host: 'https://kitbag.dev',
    path: '',
    children: createExternalRoutes([
      {
        name: 'foo',
        path: 'child',
      },
    ]),
  },
])
```

After this PR, path behaves like query. It's never required and when omitted is assumed to be an empty string.

```ts
createExternalRoutes([
  {
    host: 'https://kitbag.dev',
    children: createExternalRoutes([
      {
        name: 'foo',
        path: 'child',
      },
    ]),
  },
])
```

### DIFF
--- a/src/services/createExternalRoutes.spec.ts
+++ b/src/services/createExternalRoutes.spec.ts
@@ -19,7 +19,7 @@ test('when given external route props with a host and children return host for e
       children: createExternalRoutes([
         {
           name: 'foo',
-          path: 'child',
+          path: '/foo',
         },
       ]),
     },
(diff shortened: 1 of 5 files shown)


src/services/createExternalRoutes.spec.ts
```diff
@@ -17,7 +16,6 @@ test('when given external route props with a host and children return host for e
   const [routeA, routeB] = createExternalRoutes([
     {
       host: 'https://kitbag.dev',
-      path: '/',
```
pleek91, line 20: Isn't needed for this test. But routeB as a url would end up being https//kitbag.devchild right?

pleek91 (approved):
Nice, LGTM!

src/services/createExternalRoutes.spec.ts
stackoverfloweth: ah yeah, I'll fix

PR #230, opened 2024-07-23
replaced createRoutes with createRoute

stackoverfloweth:
Throughout the life of Kitbag Router, we have evolved our syntax for defining routes many times. We went from something very similar to vue-router to our own flattened structure of `createRoutes` which offered considerable performance improvements. Yet again we think we have found a better syntax. 

By defining the route individually, we capture generics for every piece of the route that we can use to build the best possible dx. This means we can tailor the options for a route more specifically. It also feels like it greatly simplified our internal types.

Defining routes individually means we flip the way you declare relationships of nested routes. Instead of declaring `children`, which is an array of routes, you instead create the parent first and pass `parent` into your child routes.

```txt
└── user
    └── profile
    └── settings
        └── keys
        └── notifications
```

```ts
const user = createRoute({
  name: 'user',
  path: '/user',
  component: ...,
})

const profile = createRoute({
  parent: user,
  name: 'profile',
  path: '/profile',
  component: ...,
})

const settings = createRoute({
  parent: user,
  name: 'settings',
  path: '/settings',
  component: ...,
})

const settingsKeys = createRoute({ 
  parent: settings,
  name: 'keys', 
  path: '/keys', 
  component: ... 
})

const settingsNotifications = createRoute({ 
  parent: settings,
  name: 'notifications', 
  path: '/notifications', 
  component: ... 
})
```


src/components/routerView.browser.spec.ts
```diff
@@ -70,21 +66,23 @@ test('renders components for initial route', async () => {
...
-    }, {
+    }),
+    createRoute({
       name: 'childC',
       path: '/childC',
       component: { template: 'ChildC' },
-    },
-  ])
+    }),
+  ] as const
```
pleek91, line 85: Otherwise its just `Route[]`? 

Random note, the route names don't really make sense in this test.

src/main.ts
```diff
@@ -3,5 +3,10 @@ export * from './compositions'
 export * from './errors'
 export * from './services'
 export * from './types'
+
+export type { ParamStart, ParamEnd, ExtractParamName, ExtractPathParamType, ExtractRouteParamTypes, ExtractParamTypes, ExtractParamType, MergeParams } from './types/params'
+export { paramStart, paramEnd, isParamGetter, isParamGetSet } from './types/params'
+export type { CreateRouteOptions } from './types/createRouteOptions'
```
pleek91, line 9: A lot of these I wouldn't expect to be exported. Probably just the `CreateRouteOptions` type? Curious what your thoughts are here.

src/services/getResolvedRouteForUrl.browser.spec.ts
```diff
@@ -33,98 +33,93 @@ test('given path WITHOUT params, returns match', () => {
...
+  const namedChild = createRoute({
+    parent: namedParent,
+    name: 'namedChild',
+    path: '',
+  })
+
+  const namedGrandchild = createRoute({
+    parent: namedChild,
+    name: 'namedGrandchild',
+    path: '',
```
pleek91, line 93: I noticed theres a few places in tests where we're passing an empty path. Which I think we could just omit now that `path` is optional.

src/services/host.ts
```diff
@@ -15,15 +15,13 @@ import { Identity } from '@/types/utilities'
  *
  * @example
  * ```ts
- * import { createRoutes, host } from '@kitbag/router'
+ * import { createRoute, host } from '@kitbag/router'
```
pleek91, line 18:
```suggestion
 * import { createExternalRoute, host } from '@kitbag/router'
```

src/services/host.ts
```diff
@@ -15,15 +15,13 @@ import { Identity } from '@/types/utilities'
...
- *     name: 'docs',
- *     host: host('[subdomain]', { foo: String }),
- *     component: Docs
- *   },
- * ])
+ * export const routes = createRoute({
+ *   name: 'docs',
+ *   host: host('[subdomain]', { foo: String }),
+ *   component: Docs
+ * })
```
pleek91, line 24:
```suggestion
 * export const routes = createExternalRoute({
 *   name: 'docs',
 *   host: host('[subdomain].example.com', { foo: String }),
 *   component: Docs
 * })
```

src/types/params.ts
```diff
@@ -80,8 +80,9 @@ export type ExtractPathParamType<
 export type ExtractRouteParamTypes<TRoute> = TRoute extends {
   path: { params: infer PathParams extends Record<string, Param> },
   query: { params: infer QueryParams extends Record<string, Param> },
+  host: { params: infer HostParams extends Record<string, Param> },
 }
-  ? ExtractParamTypes<MergeParams<PathParams, QueryParams>>
+  ? ExtractParamTypes<MergeParams<HostParams, MergeParams<PathParams, QueryParams>>>
```
pleek91, line 85: This made me wonder if we even need this `MergeParams` utility anymore. Can't this just be `HostParams & PathParams & QueryParams`?

src/services/createRoute.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Path, ToPath, toPath } from '@/types/path'
+import { Query, ToQuery, toQuery } from '@/types/query'
+import { Route } from '@/types/route'
+import { checkDuplicateKeys } from '@/utilities/checkDuplicateKeys'
+
+export function createRoute<
+  TName extends string | undefined = undefined,
+  TPath extends string | Path | undefined = undefined,
+  TQuery extends string | Query | undefined = undefined
+>(options: CreateRouteOptionsWithoutParent<TName, TPath, TQuery>): Route<TName extends string ? TName : '', Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>>
```
pleek91, line 18: Could we have a `ToName` or `ToKey` type for normalizing the `string | undefined` similar to `ToHost`, `ToPath`, and `ToQuery`?

src/services/createRoute.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function createRoute<
+  TParent extends Route,
+  TName extends string | undefined = undefined,
+  TPath extends string | Path | undefined = undefined,
+  TQuery extends string | Query | undefined = undefined
+>(options: CreateRouteOptionsWithParent<TParent, TName, TPath, TQuery>): Route<CombineName<TParent['key'], TName extends string ? TName : ''>, Host<'', {}>, CombinePath<TParent['path'], ToPath<TPath>>, CombineQuery<TParent['query'], ToQuery<TQuery>>>
```
pleek91, line 25: We should rename `CombineName` to `CombineKey`. We use names to create keys.

src/services/createRoute.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function createRoute(options: CreateRouteOptions | CreateRouteOptionsWithParent<Route>): Route {
+  const routeWithComponent = addRouterViewComponentIfParentWithoutComponent(options)
+  const path = toPath(options.path)
+  const query = toQuery(options.query)
```
pleek91, line 30: This would feel nice
```ts
const key = toKey(options.name)
```

src/services/createRoute.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function createRoute<
+  TParent extends Route,
+  TName extends string | undefined = undefined,
+  TPath extends string | Path | undefined = undefined,
+  TQuery extends string | Query | undefined = undefined
+>(options: CreateRouteOptionsWithParent<TParent, TName, TPath, TQuery>): Route<CombineName<TParent['key'], TName extends string ? TName : ''>, Host<'', {}>, CombinePath<TParent['path'], ToPath<TPath>>, CombineQuery<TParent['query'], ToQuery<TQuery>>>
+
+export function createRoute(options: CreateRouteOptions | CreateRouteOptionsWithParent<Route>): Route {
+  const routeWithComponent = addRouterViewComponentIfParentWithoutComponent(options)
```
pleek91, line 28: So we're just defaulting all routes to `router-view` components yeah?

src/types/createRouteOptions.ts
```diff
@@ -0,0 +1,126 @@
...
+type WithComponents = {
+  /**
+   * Multiple components for named views, which can be either synchronous or asynchronous components.
+   */
+  components: Record<string, Component>,
+}
+
+type WithComponentCallback<Params = unknown> = {
+  loadComponent: (params: Params) => any,
+}
```
pleek91, line 42: Lets chat about this. I'm not quite following the implementation you'd do to use this. But I see the params types 🎉

src/types/createRouteOptions.ts
```diff
@@ -0,0 +1,126 @@
...
+type ParentQuery<TParent extends Route | undefined> = TParent extends Route ? TParent['query'] : Query<'', {}>
+type ComponentCallbackParams<TCombinedPath extends Path, TCombinedQuery extends Query> = ExtractParamTypes<MergeParams<TCombinedPath['params'], TCombinedQuery['params']>>
+
+export type CreateRouteOptions<
+  TName extends string | undefined = string,
+  TPath extends string | Path | undefined = string | Path | undefined,
+  TQuery extends string | Query | undefined = string | Query | undefined,
+  THost extends string | Host | undefined = string | Host | undefined,
+  TParent extends Route | undefined = undefined
+> = Partial<WithComponent | WithComponents | WithComponentCallback<ComponentCallbackParams<CombinePath<ParentPath<TParent>, ToPath<TPath>>, CombineQuery<ParentQuery<TParent>, ToQuery<TQuery>>>>> & WithHooks & {
```
pleek91, line 74: I think we could simplify the component types quite a bit now that a component is always optional. Let's chat about this.

src/types/createRouteOptions.ts
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Query (aka search) part of URL.
+   */
+  query?: TQuery,
+  /**
+   * Represents additional metadata associated with a route, customizable via declaration merging.
+   */
+  meta?: RouteMeta,
+  // will be removed in follow-up PR
+  disabled?: boolean,
```
pleek91, line 96: ❤️

src/types/createRouteOptions.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+export type CreateRouteOptionsWithParent<
+  TParent extends Route,
+  TName extends string | undefined = undefined,
+  TPath extends string | Path | undefined = undefined,
+  TQuery extends string | Query | undefined = undefined
+> = CreateRouteOptions<TName, TPath, TQuery, Host<'', {}>, TParent> & {
+  parent: TParent,
+}
```
pleek91, line 114: nit: I'd probably move these into the `createRoute` service same as you did for the external route types.

pleek91 (commented):
This is awesome. Love how this turned out. So much simpler and more useful. Large PR but honestly wasn't that bad to review.

src/components/routerView.browser.spec.ts
stackoverfloweth: as const wasn't actually needed, at some point I added thinking it was the issue but it wasn't. Removed everywhere.

not sure what you don't like about route names but I changed them, lemme know if that's not better

src/main.ts
stackoverfloweth: I moved the top 2 lines from barrel of @/types without changing what's exported.

That being said I don't disagree. I'll remove lines 7 and 8

src/services/getResolvedRouteForUrl.browser.spec.ts
stackoverfloweth: great call, I removed all instances of `path: ''` from all tests

src/types/params.ts
stackoverfloweth: that does seem to be true. Why did we need it before? 🤔

src/services/createRoute.ts
stackoverfloweth: haha I did when applying your last suggestion, love it ❤️

src/services/createRoute.ts
stackoverfloweth: indeed it does ❤️

src/services/createRoute.ts
stackoverfloweth: correct. Not 100% sure yet if that's okay or if we'll need to move that logic into the RouterView

src/types/params.ts
pleek91: I think that's just a relic of previous types.

src/services/createRoute.ts
pleek91: I think that is fine for now

src/types/createRouteOptions.ts
stackoverfloweth: great catch. I actually don't need both. Keeping a version of this here and removing the duplicate type in external

src/types/createRouteOptions.ts
pleek91: We removed this because we plan on changing it anyway.

docs/api/functions/host.md
```diff
@@ -29,14 +29,12 @@ An object representing the host which includes the host string, its parameters,
...
-  {
-    name: 'home',
-    host: host('https://[subdomain].kitbag.dev', { subdomain: Number })
-  },
-])
+import { createExternalRoute, host } from '@kitbag/router'
+
+export const docs = createExternalRoute({
+  name: 'home',
+  host: host('https://[subdomain].kitbag.dev', { subdomain: Number })
```
pleek91, line 36: using `Number` here doesn't really make sense as an example. Probably fine to just use `String` even though that's the default

docs/core-concepts/defining-routes.md
```diff
@@ -68,17 +86,15 @@ When an individual route is disabled, it will never count as an exact match. Chi
...
-  }
-])
+const user = createRoute({
+  name: 'user',
+  path: '/user',
+  disabled: true, // [!code focus] 
+  component: ...,
+  children: createRouter([
+    ...
+  ])
```
pleek91, line 96: Need to remove `children` from here.

pleek91 (approved):
Nice work man!

PR #232, opened 2024-07-25
update docs gifs for v0.6+

stackoverfloweth:


pleek91 (approved):
The new ones are significantly smaller than the previous ones. Idk if that was intentional or not.

stackoverfloweth:
> The new ones are significantly smaller than the previous ones. Idk if that was intentional or not. 

Yes I sized them down to the size they are rendered as on the actual page in hopes it would decrease file size and improve page load times.

PR #233, opened 2024-07-26
remove disabled concept

stackoverfloweth:
with the latest iteration of defining routes we can now define parent routes and choose to not include them when calling `createRouter`. This effectively does the task of what "disabled" did previously but in a more straight forward and hopefully obvious way to our developers.

Currently, kitbag router lets you set a route as "disabled".  A disabled route still effects the children, passing down params, path, query, host, and key but a disabled route can never be matched. So end users cannot create a `router-link` to a disabled route or navigate with `router.push` to a disabled route.

```ts
const routes = createRoutes([
  { name: 'parent', disabled: true, children: createRoutes([
    { name: 'foo', component: ... },
    { name: 'bar', component: ... },
  ])
])

```

With our new syntax `createRoute`, we can still create the parent route and assign as a parent but omit from our routes when calling `createRouter`.

```ts
const parent = createRoute({ name: 'parent' })

const routes = [
  createRoute({ parent, name: 'foo', component: ... }),
  createRoute({ parent, name: 'bar', component: ... },)
] as const

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ router.push('/user/settings')
 router.push('https://github.com/kitbagjs/router')
 ```
 
-This `source` argument is type safe, expecting either a Url or a valid route "key". Url is any string that starts with "http", "https", or a forward slash "/". Route key is a string of route names joined by a period `.` that lead to a non-disabled route. Additionally if using the route key, push will require params be passed in if there are any.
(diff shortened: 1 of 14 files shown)


pleek91 (approved):
So much simpler ❤️

PR #234, opened 2024-07-26
Refactor types for creating routes and add overloads and generics to `createRoute` for component types

pleek91:
# Description
To make the types for `createRoute` and `createExternalRoute` more intuitive I'm breaking apart `CreateRouteOptions` into more pieces and using those pieces directly within the overload declarations. This contrasts the current implementation where there are two type `CreateRouteOptionsWithoutParent` and `CreateRouteOptionsWithParent`.

We can now express options like `CreateRouteOptions & WithParent & WithComponent & ...` which makes each overload more intuitive and prevents some of the overlap between `createRoute` and `createExternalRoute`. For example, `host` is not an option when using `createRoute` and `component` is not an option when using `createExternalRoute`.

This also adds overloads and generics to `createRoute` specifically for component(s).

### DIFF
--- a/src/services/createExternalRoute.ts
+++ b/src/services/createExternalRoute.ts
@@ -2,25 +2,13 @@ import { markRaw } from 'vue'
 import { CombineKey } from '@/services/combineKey'
 import { CombinePath } from '@/services/combinePath'
 import { CombineQuery } from '@/services/combineQuery'
-import { combineRoutes, CreateRouteOptions, isWithParent, WithoutParent, WithParent } from '@/types/createRouteOptions'
+import { combineRoutes, CreateRouteOptions, isWithHost, isWithParent, WithHost, WithoutHost, WithoutParent, WithParent } from '@/types/createRouteOptions'
 import { Host, toHost, ToHost } from '@/types/host'
 import { ToKey, toKey } from '@/types/key'
 import { Path, toPath, ToPath } from '@/types/path'
 import { Query, toQuery, ToQuery } from '@/types/query'
 import { Route } from '@/types/route'
 import { checkDuplicateKeys } from '@/utilities/checkDuplicateKeys'
-
-type WithHost<THost extends string | Host> = {
-  host: THost,
-}
-
-function isWithHost(options: CreateRouteOptions): options is CreateRouteOptions & WithHost<string | Host> {
-  return 'host' in options && Boolean(options.host)
-}
-
-type WithoutHost = {
-  host?: never,
-}
 
 export function createExternalRoute<
   const THost extends string | Host,
--- a/src/services/createExternalRoute.ts
+++ b/src/services/createExternalRoute.ts
@@ -2,7 +2,7 @@ import { markRaw } from 'vue'
 import { CombineKey } from '@/services/combineKey'
 import { CombinePath } from '@/services/combinePath'
 import { CombineQuery } from '@/services/combineQuery'
-import { combineRoutes, CreateRouteOptions, CreateRouteOptionsWithoutParent, CreateRouteOptionsWithParent, isRouteWithParent } from '@/types/createRouteOptions'
+import { combineRoutes, CreateRouteOptions, isWithParent, WithoutParent, WithParent } from '@/types/createRouteOptions'
 import { Host, toHost, ToHost } from '@/types/host'
 import { ToKey, toKey } from '@/types/key'
 import { Path, toPath, ToPath } from '@/types/path'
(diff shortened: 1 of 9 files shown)


src/services/createExternalRoute.ts
```diff
@@ -11,24 +11,24 @@ import { Route } from '@/types/route'
...
->(options: CreateRouteOptionsWithParent<TParent, TName, TPath, TQuery, Host<'', {}>>): Route<CombineKey<TParent['key'], ToKey<TName>>, ToHost<Host<'', {}>>, CombinePath<TParent['path'], ToPath<TPath>>, CombineQuery<TParent['query'], ToQuery<TQuery>>>
+>(options: CreateRouteOptions<TName, TPath, TQuery> & WithoutHost & WithParent<TParent>): Route<CombineKey<TParent['key'], ToKey<TName>>, Host<'', {}>, CombinePath<TParent['path'], ToPath<TPath>>, CombineQuery<TParent['query'], ToQuery<TQuery>>>
 
-export function createExternalRoute(options: CreateRouteOptions | CreateRouteOptionsWithParent<Route>): Route {
+export function createExternalRoute(options: CreateRouteOptions): Route {
   const key = toKey(options.name)
   const path = toPath(options.path)
   const query = toQuery(options.query)
-  const host = toHost(options.host ?? '')
+  const host = isWithHost(options) ? toHost(options.host) : toHost('')
```
pleek91, line 31: This check is only needed because of types. Functionally its the same.

src/services/paramValidation.spec.ts
```diff
@@ -145,15 +145,14 @@ test.each([
 })
 
 test.each([
-  { path: '/duplicate/[foo]', query: 'params=[?foo]' },
+  { path: '/duplicate/[foo]', host: 'https://router.kitbag.dev', query: 'params=[?foo]' },
```
pleek91, line 148: This test correctly type errored because you cannot create an external route without a host.

src/services/paramValidation.spec.ts
```diff
@@ -145,15 +145,14 @@ test.each([
...
   { path: '/duplicate/[foo]', host: 'https://[foo].kitbag.dev' },
   { path: '/', host: 'https://[?foo].kitbag.dev', query: 'params=[?foo]' },
   { path: '/duplicate/[foo]', host: 'https://[foo].kitbag.dev', query: 'params=[foo]' },
 ])('given route with duplicate param names across path and query, throws DuplicateParamsError', (route) => {
   const action: () => void = () => createExternalRoute({
     name: 'different-cased-params',
     ...route,
-    component,
```
pleek91, line 156: All of these external routes that had a component correctly errored

src/types/createRouteOptions.ts
```diff
@@ -22,63 +22,68 @@ type WithHooks = {
...
+  /**
+   * A Vue component, which can be either synchronous or asynchronous components.
+   */
+  component?: TComponent,
 }
 
-export function isRouteWithComponents(value: CreateRouteOptions): value is CreateRouteOptions & WithComponentRecord {
-  return 'components' in value
+export function isWithComponent(options: CreateRouteOptions): options is CreateRouteOptions & { component: Component } {
+  return 'component' in options && Boolean(options.component)
```
pleek91, line 62: notice I added `Boolean` checks in these guards. Just a little strictor to make sure there's actually a value

src/types/createRouteOptions.ts
```diff
@@ -22,63 +22,68 @@ type WithHooks = {
...
+  parent?: never,
+}
+
+export type WithComponent<
+  TComponent extends Component | undefined = Component | undefined
+> = {
+  /**
+   * A Vue component, which can be either synchronous or asynchronous components.
+   */
+  component?: TComponent,
```
stackoverfloweth, line 58: I'm surprised to see the `?` here. I think I personally preferred the `Partial<...>` route so `isWithComponent` would say `options is `WithComponent` but don't feel strongly if you think this is simpler

src/services/createRoute.ts
```diff
@@ -15,17 +15,34 @@ import { checkDuplicateKeys } from '@/utilities/checkDuplicateKeys'
 export function createRoute<
   const TName extends string | undefined = undefined,
   const TPath extends string | Path | undefined = undefined,
-  const TQuery extends string | Query | undefined = undefined
->(options: CreateRouteOptionsWithoutParent<TName, TPath, TQuery>): Route<ToKey<TName>, Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>>
+  const TQuery extends string | Query | undefined = undefined,
+  TComponent extends Component | undefined = undefined
```
stackoverfloweth, line 19: does this new TComponent generic add something that's useful for us in future PRs? It's not clear to me what retaining the type for component does for our route definition right now

stackoverfloweth (approved):
couple of small comments, otherwise looks like a nice change

src/services/createRoute.ts
pleek91: Correct this isn't really useful in the PR but it will be in the next one

src/types/createRouteOptions.ts
pleek91: I'd like the `With` types to reflect the use case. The `component` property is always optional when defining a route so I think its best to make it optional in the type itself rather than wrap it in `Partial` in the function definition. Plus the function types are a lot as is so reducing `Partial<WithComponent<TComponent>>` down to `WithComponent<TComponent>` makes it a little easier to read.

stackoverfloweth: @pleek91 Sounds good. Couldn't you define the type guard as `options is Required<WithComponent>`?

PR #237, opened 2024-07-28
combine register interface and meta, case insensitive

stackoverfloweth:
with the [recent change to route meta](https://github.com/kitbagjs/router/pull/236) we noticed that we had `RouteMeta` as it's own interface that would require declaration merging in addition to the `Register` interface.

```ts
declare module '@kitbag/router' {
  interface Register {
    router: typeof router
  }

  interface RouteMeta {
    isPublic?: boolean
  }
}
```

This PR combines them and also makes the keys of register case insensitive

```ts
declare module '@kitbag/router' {
  interface Register {
    router: typeof router,
    routeMeta: {
      isPublic?: boolean
    }
  }
}
```


stackoverfloweth:
@pleek91 this PR also deletes `RouteState`. No idea what that was supposed to be used for but seems unused and worth deleting?

pleek91 (approved):
LGTM though I would vote to not accept multiple cases of the properties in the `Register` interface.

stackoverfloweth:
> LGTM though I would vote to not accept multiple cases of the properties in the `Register` interface.

@pleek91 I'm open to reverting that, for context I was questioning whether keys inside an object should be camel or pascal. So far we have consistently done camel, but I could see an argument for pascal so I figured this lets devs use whichever they think is more correct?

PR #238, opened 2024-07-29
Route state

stackoverfloweth:
This PR establishes the ability to fully utilize [history state](https://developer.mozilla.org/en-US/docs/Web/API/History/state). 

Now routes can define the expected format for state with the `state` option, using the same format we already use for route params. This includes the default types like `String`, `Number`, `Boolean`, etc. As well as support for defaults and custom params.

```ts
const route = createRoute({
  name: 'foo',
  state: {
    bar: Number
  }
})
```

State is set with `router.push` and `router.replace` within the options argument.

```ts
router.replace('foo', {}, { state: { bar: 14 } })
```

State is access from a resolved route with the `state` prop.

```ts
const route = useRoute()

route.state.foo // <-- 14
```

## Open Questions

- [ ] currently history.refresh does not do anything with state, which I assume will lose state if there was any. Is that expected?
- [ ] should there be other ways of setting state?
- [ ] currently the params expect to be given a string but I'm not doing anything to explicitly turn values into a string when setting state. State comes in from history as `unknown`, should I expect that to be `Record<string, string>`?


src/types/createRouteOptions.ts
```diff
@@ -97,6 +99,10 @@ export type CreateRouteOptions<
    * Represents additional metadata associated with a route, customizable via declaration merging.
    */
   meta?: TMeta,
+  /**
+   * Type params for additional data intended to be stored in history state, all keys will be optional.
```
pleek91, line 103:
```suggestion
   * Type params for additional data intended to be stored in history state, all keys will be optional unless a default is provided.
```

src/types/route.ts
```diff
@@ -25,17 +21,18 @@ export type Route<
...
   TQuery extends Query = Query,
-  TMeta extends RouteMeta = RouteMeta
+  TMeta extends RouteMeta = RouteMeta,
+  TStateParams extends Record<string, Param> = Record<string, Param>
 > = {
   /**
    * The specific route properties that were matched in the current route.
   */
-  matched: CreateRouteOptionsMatched & { meta: TMeta },
+  matched: CreateRouteOptions & WithHooks & (WithHost | WithoutHost) & (WithComponent | WithComponents) & (WithParent | WithoutParent) & { meta: TMeta } & { state: TStateParams },
```
pleek91, line 30: Can't this just be `CreateRouteOptionsMatched & { state: TStateParams }`? Rather than inlining the entire `CreateRouteOptionsMatched` type.

src/types/routerPush.ts
```diff
@@ -1,21 +1,24 @@
...
 import { Url } from '@/types/url'
 import { AllPropertiesAreOptional } from '@/types/utilities'
 
-export type RouterPushOptions = {
+export type RouterPushOptions<
+  TState = unknown
+> = {
   query?: Record<string, string>,
   replace?: boolean,
+  state?: Partial<TState>,
```
pleek91, line 13: This is interesting. I didn't imaging you'd be able to set the state when navigating. But kinda cool that you can.

src/types/state.ts
```diff
@@ -0,0 +1,25 @@
+import { ExtractParamTypes } from '@/types/params'
+import { Param } from '@/types/paramTypes'
+import { Routes } from '@/types/route'
+import { RouteGetByKey } from '@/types/routeWithParams'
+
+export type RouteState<T extends Record<string, unknown>> = {
+  scrollPosition?: number,
+  pageTitle?: string,
```
pleek91, line 8: Agreed these will likely exist at some point. But we should probably leave them out for now. 
```suggestion
```

pleek91:
@stackoverfloweth let's discuss the possibility of merging state similar to `route.params`. But overall this is looking really good. 

Also curious if you add a default to a state param will the types reflect that the value will always be defined? I saw the test to prove the value will be defined. But not the types. Might have missed that.

pleek91:
> currently history.refresh does not do anything with state, which I assume will lose state if there was any. Is that expected?

This is a good question. I think we always want to update the state when 
- Navigation happens
- a refresh happens
- a fresh page load

A fresh page load won't always have state but it might if the user refreshes the page via the browser.

> should there be other ways of setting state?

I'd expect to be able to set state directly on the route and via route.update

```ts
route.state.foo = 'bar'

router.update({}, { state: { foo: 'bar' })
```

> currently the params expect to be given a string but I'm not doing anything to explicitly turn values into a string when setting state. State comes in from history as `unknown`, should I expect that to be `Record<string, string>`?

We already discussed this but we need to use both the set and get from the param. Set when storing the value and get when retrieving it. That way we can use the user's defined logic. Otherwise we'll end up with weird things like losing class instances or not converting a csv like param back into an array.

pleek91 (approved):

src/services/createRouter.spec.ts
```diff
@@ -7,13 +7,14 @@ import * as resolveUtilities from '@/services/createRouterResolve'
...
-      name: 'root',
-      component,
-      path: '/',
-    }),
-  ], {
+  const foo = createRoute({
+    name: 'root',
+    component,
+    path: '/',
+    state: { something: Boolean },
```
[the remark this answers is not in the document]
stackoverfloweth: this is outdated, if you look again you'll see that state gets removed from this test and is instead used in a separate test created to test initial state specifically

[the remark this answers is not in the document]
pleek91: Sorry I didn't mean to add this. Apparently I had a draft from days ago.

PR #239, opened 2024-07-29
Add `props` to `createRoute` to bind props and attrs to route components

pleek91:
# Description
Add an additional optional `props` property to both the `WithComponent` and `WithComponents` types that allows for binding props and attrs to the route component when it is mounted. The `props` callback is passed all of the route's params and can also return a promise. 

Given the following route
```ts
const route = createRoute({
  path: path('/[foo]/[bar]', { bar: Number }),
  component: echo,
})
```

Adding `props` is typed as

```ts
(params: { bar: number; foo: string; }) => MaybePromise<{ value: string; }>
```

image

If a route uses `components` the `props` are typed as a record keyed by the components. So each component can have a props callback.


src/components/routerView.browser.spec.ts
```diff
@@ -115,14 +116,11 @@ test('updates components when route changes', async () => {
   expect(app.html()).toBe('Foo')
 })
 
-test.each([
-  defineAsyncComponent(() => import('./helloWorld')),
-  () => import('./helloWorld'),
```
pleek91, line 120: This is an important thing to note. Previously we supported any async import as a component by automatically wrapping any function in `defineAsyncComponent`. 

I'm removing that in this PR. Now if you want to use an async component you should declare it as such by using vue's `defineAsyncComponent` utility. 

This will give us more control over components without transforming them (sometimes unnecessarily). Since we cannot detect if a component function is an async import without first calling it (which makes the import call).

src/components/routerView.vue
```diff
@@ -43,21 +46,27 @@
...
-      if (typeof component === 'function') {
-        return defineAsyncComponent(component as AsyncComponentLoader)
-      }
+    if (!component) {
+      return null
+    }
 
-      return component
+    if (props) {
+      return componentUtil(component, () => props(route.params))
```
pleek91, line 57: This is the important piece of implementation. If we detect a route had props for the component we wrap the component in our `component` utility which already handles binding props from a callback (which can return a promise).

src/services/createRoute.ts
```diff
@@ -3,48 +3,74 @@ import { CombineKey } from '@/services/combineKey'
...
+
 export function createRoute<
   const TName extends string | undefined = undefined,
   const TPath extends string | Path | undefined = undefined,
   const TQuery extends string | Query | undefined = undefined,
-  const TMeta extends RouteMeta = RouteMeta,
-  TComponent extends Component | undefined = undefined
->(options: CreateRouteOptions<TName, TPath, TQuery, TMeta> & WithHooks & WithComponent<TComponent> & WithoutParent): Route<ToKey<TName>, Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>, TMeta>
+  const TMeta extends RouteMeta = RouteMeta
+>(options: CreateRouteOptions<TName, TPath, TQuery, TMeta> & WithHooks & WithoutComponents & WithoutParent): Route<ToKey<TName>, Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>>
```
pleek91, line 31: This is a new overload that includes `WithoutComponents`. This helps typescript correctly identify which overload should be used when either `component` or `components` are passed.

src/services/createRoute.ts
```diff
@@ -3,48 +3,74 @@ import { CombineKey } from '@/services/combineKey'
...
 export function createRoute<
+  const TParent extends Route,
   const TName extends string | undefined = undefined,
   const TPath extends string | Path | undefined = undefined,
   const TQuery extends string | Query | undefined = undefined,
-  const TMeta extends RouteMeta = RouteMeta,
-  TComponents extends Record<string, Component> | undefined = undefined
->(options: CreateRouteOptions<TName, TPath, TQuery, TMeta> & WithHooks & WithComponents<TComponents> & WithoutParent): Route<ToKey<TName>, Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>, TMeta>
+  const TMeta extends RouteMeta = RouteMeta
+>(options: CreateRouteOptions<TName, TPath, TQuery, TMeta> & WithHooks & WithoutComponents & WithParent<TParent>): Route<CombineKey<TParent['key'], ToKey<TName>>, Host<'', {}>, CombinePath<TParent['path'], ToPath<TPath>>, CombineQuery<TParent['query'], ToQuery<TQuery>>>
```
pleek91, line 39: This is a second new overload which is `WithoutComponets & WithParent`.

src/services/createRoute.ts
```diff
@@ -3,48 +3,74 @@ import { CombineKey } from '@/services/combineKey'
...
 
 export function createRoute<
+  TComponent extends Component,
+  const TName extends string | undefined = undefined,
+  const TPath extends string | Path | undefined = undefined,
+  const TQuery extends string | Query | undefined = undefined,
+  const TMeta extends RouteMeta = RouteMeta
+>(options: CreateRouteOptions<TName, TPath, TQuery, TMeta> & WithHooks & WithComponent<TComponent, RouteParams<TPath, TQuery>> & WithoutParent): Route<ToKey<TName>, Host<'', {}>, ToPath<TPath>, ToQuery<TQuery>>
```
pleek91, line 47: The rest of the changes to overloads are adding params to the component types. What use to be `WithComponent<TComponent>` is now `WithComponent<TComponent, RouteParams<TPath, TQuery>>`

src/types/createRouteOptions.ts
```diff
@@ -50,28 +51,40 @@ export type WithoutParent = {
...
 export type WithComponent<
-  TComponent extends Component | undefined = Component | undefined
+  TComponent extends Component = Component,
+  TParams extends Record<string, unknown> = Record<string, unknown>
 > = {
   /**
    * A Vue component, which can be either synchronous or asynchronous components.
    */
-  component?: TComponent,
+  component: TComponent,
```
pleek91, line 60: Adding the `WithoutComponents` overloads and making `WithComponent` and `WithComponents` have required properties fixed some issue with correctly identifying overloads. And means that the `isWithComponent` and `isWithComponents` utilities now just assert `is WithComponent(s)`

src/types/createRouteOptions.ts
```diff
@@ -50,28 +51,40 @@ export type WithoutParent = {
...
+  TComponents extends Record<string, Component> = Record<string, Component>,
+  TParams extends Record<string, unknown> = Record<string, unknown>
 > = {
   /**
    * Multiple components for named views, which can be either synchronous or asynchronous components.
    */
-  components?: TComponents,
+  components: TComponents,
+  props?: {
+    [TKey in keyof TComponents]?: (params: TParams) => TComponents[TKey] extends Component ? MaybePromise<ComponentProps<TComponents[TKey]>> : {}
```
pleek91, line 77: You might think that the props type here and in `WithComponent` could be abstracted into a type that accepts the component and the params. And it totally can, but inlining these produces much better type hints. Otherwise typescript doesn't expand the type and if you get an error its more difficult to decipher. Choosing to inline this type in both so that the devx is better.

PR #242, opened 2024-08-14
merge parent route meta with children route meta

stackoverfloweth:
This PR updates `createRoute` utility to use a new `combineMeta` to merge child meta with parent (if parent exists). This follows the pattern used for state (and others).  Merging meta is just a bit unique in that duplicate keys are allowed as long as the child's property of the same name matches `typeof` for the parent. The thinking here being that the child can override the meta property of a parent.

This PR also adds a new test suite for `createRoute`, with tests for new meta functionality as well as combine logic for other parts of the route.

### DIFF
--- a/src/services/createRoute.spec.ts
+++ b/src/services/createRoute.spec.ts
@@ -1,0 +1,125 @@
+import { expect, test } from 'vitest'
+import { createRoute } from '@/services/createRoute'
+import { path } from '@/services/path'
+import { query } from '@/services/query'
+
+test('given parent, key is combined', () => {
+  const parent = createRoute({
+    name: 'parent',
+  })
+
+  const child = createRoute({
+    parent: parent,
+    name: 'child',
+  })
+
+  expect(child.key).toBe('parent.child')
+})
+
+test('given parent, path is combined', () => {
+  const parent = createRoute({
+    path: '/parent',
+  })
+
+  const child = createRoute({
+    parent: parent,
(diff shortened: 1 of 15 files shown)


src/services/createRoute.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  expect(child.query).toMatchObject({
+    query: 'static=123&sort=[sort]',
+    params: {
+      sort: Boolean,
+    },
+  })
+})
+
+test('given parent, state is combined into stateParams', () => {
```
pleek91, line 55: every time I read `stateParams` I wonder why its not just `state` like `path` and `query` and `host`.

pleek91 (approved):
Nice, this looks great

src/services/createRoute.spec.ts
stackoverfloweth: well the issue stems from `ResolveRoute` needing to know both what the schema (params) for the state is, as well as having the ACTUAL runtime values for the state. This is unlike path/query/etc where once we go from Route -> Resolved the router no longer needs to know what the schema (params) is and if it did, it could use the `route.matched`. We might be able to store state on `route.matched`, but I think that would require a generic which I know you were opposed to.

honestly though, I prefer this delineation between the params and the runtime values. I would like to see us update `path` -> `pathParams`, `query` -> `queryParams`, etc

stackoverfloweth: resolved sync

PR #248, opened 2024-08-22
New router "base" option

### LINKED ISSUE #245: Serve app at subfolder?
For example, we want our Vue app to run at `/p/` URL. In `vue-router`, we achieve that by [`createWebHistory`](https://router.vuejs.org/api/#createWebHistory), but in `@kitbag/router`, I haven't found an equivalent.

stackoverfloweth:
closes #245 

With this change, developers can assign a "base" when creating a router

```ts
  base: '/foo',
})
```

This will have the same effect of creating a root non-named route with the path `/foo`. All routes will have their paths prefixed with "/foo".


src/services/insertBaseRoute.ts
```diff
@@ -0,0 +1,19 @@
+import { combinePath } from '@/services/combinePath'
+import { path } from '@/services/path'
+import { Routes } from '@/types'
+import { isNestedArray, stringHasValue } from '@/utilities/guards'
+
+export function insertBaseRoute(routesOrArrayOfRoutes: Routes | Routes[], base?: string): Routes {
+  const routes = isNestedArray(routesOrArrayOfRoutes) ? routesOrArrayOfRoutes.flat() : routesOrArrayOfRoutes
```
pleek91, line 7: I'd probably pull this out so that this method isn't also responsible for flattening the routes.

src/services/insertBaseRoute.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (!stringHasValue(base)) {
+    return routes
+  }
+
+  const basePath = path(base, {})
+
+  return routes.map(route => ({
+    ...route,
+    path: combinePath(basePath, route.path),
+  }))
```
pleek91, line 18: You could also use createRoute here right? Idk that it makes a difference, just what I assumed to find here. 

```
const baseRoute = createRoute({ path: base })

return routes.map(route => createRoute({
  parent: baseRoute,
  ...route
})
```

pleek91 (commented):
Overall this looks like I'd expect. But I'm curious if users who use `base` will expect to not see the base show up in their routes when working with middleware. Like base doesn't really exist except for when converting strings to routes and routes to strings. 

Could base just be used in `resolve` and `find`?

pleek91 (approved):

PR #251, opened 2024-08-31
Using Route Name > Key

stackoverfloweth:
This PR removes the concept of a route `key`. 

We think routes named with dot notation for context is a good pattern, but we realized that we don't actually need this to be enforced by the router. As long as each route has a unique name, you should be free to use that routes name to navigate.

Previously, route `name` was combined from parent to child similar to `path`, `query`, `state`, and `meta` to create the `key`. 

For example

```ts
const settings = createRoute({
  // has key `settings`
  name: 'settings',
  path: '/settings',
  component: ...,
})

const settingsKeys = createRoute({ 
  // has key `settings.keys`
  parent: settings,
  name: 'keys', 
  path: '/keys', 
  component: ... 
})

const settingsNotifications = createRoute({ 
  // has key `settings.notifications`
  parent: settings,
  name: 'notifications', 
  path: '/notifications', 
  component: ... 
})
```

After this PR, using this dot notation will be totally optional. Developers can give each route a unique name and navigate to that route by it's name only.

```ts
const settings = createRoute({
  name: 'settings',
  path: '/settings',
  component: ...,
})

const settingsKeys = createRoute({ 
  parent: settings,
  name: 'keys', 
  path: '/keys', 
  component: ... 
})

const settingsNotifications = createRoute({ 
  parent: settings,
  name: 'notifications', 
  path: '/notifications', 
  component: ... 
})
```

In this example, we can navigate with 

```ts
router.push('notifications')
```

The router now has a check for duplicate route names very similar to how it checks for duplicate param names. If a developer tries to create routes with duplicate names, they'll get a runtime error `DuplicateNamesError`. 

Docs should be updated accordingly


docs/advanced-concepts/route-narrowing.md
```diff
@@ -19,29 +19,29 @@ const user = createRoute({
...
-  name: 'settings',
+  name: 'user.settings',
   path: '/settings',
   component: ...,
 })
 
 const router = createRouter([home, user, profile, settings])
 
-router.route.key // "home" | "user" | "user.profile" | "user.settings"
+router.route.name // "home" | "user" | "profile" | "settings"
```
pleek91, line 37: I think this contradicts the change above where you made the names dot notation user.settings and user.profile

docs/advanced-concepts/route-narrowing.md
```diff
@@ -50,7 +50,7 @@ You can also use the `isRoute` type guard. You could write the same logic as abo
 
 ```ts
 if(router.route, 'user', { exact: true }) {
-  router.route.key // "user"
+  router.route.name // "user"
```
pleek91, line 53: I like that there's one less concept ❤️ route.name reads so nicely

pleek91 (approved):
Nice, I think this is a good change. Puts the naming convention in userland and makes the router simpler by removing the "key" concept altogether.

docs/advanced-concepts/route-narrowing.md
stackoverfloweth: oops, your right

docs/advanced-concepts/route-narrowing.md
stackoverfloweth: agreed 👍

PR #255, opened 2024-09-03
Prefetch async route components

pleek91:
# Description
Adds a new feature where if a route is defined with an async component (by using vue's `defineAsyncComponent` utility) that component will be prefetched into the app when a `router-link` to that route is mounted.


src/components/routerLink.browser.spec.ts
```diff
@@ -230,4 +231,139 @@ test.each([
...
+  let loaded = undefined
+
+  const route = createRoute({
+    name: 'route',
+    path: '/route',
+    component: defineAsyncComponent(() => {
+      return new Promise(resolve => {
+        loaded = true
+        resolve({ default: { template: '' } })
+      })
```
stackoverfloweth, line 252: clever!

src/components/routerLink.vue
```diff
@@ -7,13 +7,29 @@
...
+    /**
+     * Determines what assets are prefetched when router-link is rendered for this route. Overrides route level prefetch.
+     */
+    prefetch?: PrefetchConfig,
+  }
+
+  const props = withDefaults(defineProps<RouterLinkProps & RouterPushOptions>(), {
+    // because prefetch can be a boolean vue automatically sets the default to false.
+    // Specifically setting the default to undefined
+    prefetch: undefined,
```
stackoverfloweth, line 31: boo macros 👎

stackoverfloweth (approved):
very clean solution, great feature 👏

PR #256, opened 2024-09-05
store rejections type in router

stackoverfloweth:
This PR should have no impact on the router at runtime. All it does is move the responsibility of defining rejection types from being part of the register interface

```ts
declare module '@kitbag/router' {
  interface Register {
    // router: typeof router
    rejections: ['AuthNeeded']
  }
}
```

into a generic captured by `createRouter`

```ts
  rejections: {
    AuthNeeded: MyLoginComponent
  }
})
```

If approved, will need a docs change before merging

### DIFF
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -53,9 +53,9 @@ type RouterUpdateOptions = {
  * const router = createRouter(routes)
  * ```
  */
-export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(routes: TRoutes, options?: TOptions): Router<TRoutes, TOptions>
-export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(arrayOfRoutes: TRoutes[], options?: TOptions): Router<TRoutes, TOptions>
-export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(routesOrArrayOfRoutes: TRoutes | TRoutes[], options: TOptions): Router<TRoutes, TOptions> {
+export function createRouter<const TRoutes extends Routes, const TOptions extends RouterOptions>(routes: TRoutes, options?: TOptions): Router<TRoutes, TOptions>
+export function createRouter<const TRoutes extends Routes, const TOptions extends RouterOptions>(arrayOfRoutes: TRoutes[], options?: TOptions): Router<TRoutes, TOptions>
+export function createRouter<const TRoutes extends Routes, const TOptions extends RouterOptions>(routesOrArrayOfRoutes: TRoutes | TRoutes[], options: TOptions): Router<TRoutes, TOptions> {
   const flattenedRoutes = isNestedArray(routesOrArrayOfRoutes) ? routesOrArrayOfRoutes.flat() : routesOrArrayOfRoutes
   const routes = insertBaseRoute(flattenedRoutes, options.base)
 
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -53,9 +53,9 @@ type RouterUpdateOptions = {
  * const router = createRouter(routes)
  * ```
  */
-export function createRouter<const T extends Routes>(routes: T, options?: RouterOptions): Router<T>
-export function createRouter<const T extends Routes>(arrayOfRoutes: T[], options?: RouterOptions): Router<T>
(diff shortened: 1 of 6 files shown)


src/services/createRouter.ts
```diff
@@ -227,7 +227,7 @@ export function createRouter<const T extends Routes>(routesOrArrayOfRoutes: T |
     app.provide(routerInjectionKey, router as any)
   }
 
-  const router: Router<T> = {
+  const router: Router<TRoutes> = {
```
stackoverfloweth, line 230: this is worth noting, the router we create doesn't actually use the `TOptions` argument. In fact if I add it, I'd have to change more about hooks to possibly deviate from just `RegisteredRejectionType`.

src/services/createRouter.ts
```diff
@@ -53,9 +53,9 @@ type RouterUpdateOptions = {
  * const router = createRouter(routes)
  * ```
  */
-export function createRouter<const T extends Routes>(routes: T, options?: RouterOptions): Router<T>
-export function createRouter<const T extends Routes>(arrayOfRoutes: T[], options?: RouterOptions): Router<T>
-export function createRouter<const T extends Routes>(routesOrArrayOfRoutes: T | T[], options: RouterOptions = {}): Router<T> {
+export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(routes: TRoutes, options?: TOptions): Router<TRoutes, TOptions>
+export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(arrayOfRoutes: TRoutes[], options?: TOptions): Router<TRoutes, TOptions>
+export function createRouter<const TRoutes extends Routes, TOptions extends RouterOptions>(routesOrArrayOfRoutes: TRoutes | TRoutes[], options: TOptions): Router<TRoutes, TOptions> {
```
pleek91, line 58: Might not be necessary but I'd add const to the new TOptions generic

src/services/createRouterReject.ts
```diff
@@ -3,23 +3,22 @@ import { genericRejection } from '@/components/rejection'
...
-type GetRejectionRoute = (type: RouterRejectionType) => ResolvedRoute
+type GetRejectionRoute = (type: RegisteredRejectionType) => ResolvedRoute
 type IsRejectionRoute = (route: ResolvedRoute) => boolean
 
-export type RouterRejection = Ref<null | { type: RouterRejectionType, component: Component }>
+export type RouterRejection = Ref<null | { type: RegisteredRejectionType, component: Component }>
 
 type CreateRouterRejectContext = {
-  rejections?: RouterRejectionComponents['rejections'],
+  rejections?: Partial<Record<string, Component>>,
```
pleek91, line 21: Is this partial needed?

stackoverfloweth (commented):
technically a lot of the changes made in hooks.ts don't HAVE to be done. This technically works even when types/Router doesn't even use the TOptions argument in the type at all. I added the generic to hooks/rejections because it meant that those options actually served a purpose.

src/types/router.ts
```diff
@@ -86,27 +92,27 @@ export type Router<
   /**
    * Registers a hook to be called before a route is entered.
    */
-  onBeforeRouteEnter: AddBeforeRouteHook,
+  onBeforeRouteEnter: AddBeforeRouteHook<TOptions>,
```
pleek91, line 95: Why do these need generics now?

pleek91 (commented):
Overall looks good. But we're adding a lot of generic passing to hooks which previously were pretty simple. Any concerns there?

src/services/createRouterReject.ts
stackoverfloweth: yes and no. It adds `undefined` as a possible value which is what we want. I could make it `Record<string, Component | undefined>` if you like that more

pleek91: Makes sense, fine as is

src/types/router.ts
stackoverfloweth: we don't need the generics, technically I could remove the generics from hooks/rejections and just have the `Router<TRoutes, TOptions>` just eslint ignore the fact that `TOptions` is unused.

pleek91: Seems like if we don't need them we shouldn't use them. Curious what this looks like without them

pleek91 (approved):

Issue #244, opened 2024-08-15
onBeforeRouteEnter not triggered if coming from reject route


stackoverfloweth:
I added a `onBeforeRouteEnter` hook to my route

```ts
const profile = createRoute({
	name: 'profile',
	path: '/profile',
	component: defineAsyncComponent(() => import('../components/ProfilePage.vue')),
	onBeforeRouteEnter: (to, { reject }) => {
		if (!activeUser.value) {
			reject('AuthNeeded')
		}
	},
})
```

works totally as expected the first time, if I try navigating to `/profile` without an active user the rejection happens, is handled by sending me to rejection route (doesn't matter if I use built in rejection component or custom). However, if I click a router-link again to go to `/profile` it seems to totally bypass this `onBeforeRouteEnter` hook.

pleek91:
@stackoverfloweth is this when navigating from the rejection back to profile? Or just anytime you try and go to profile after hitting a rejection?

stackoverfloweth:
@pleek91 an update on this
I traced the issue to `getBeforeRouteHooksFromRoutes`, which takes `to` and `from` and returns all the hooks.

The first time I enter my route with this hook I see a hook in `onBeforeRouteEnter ` where I'd expect
<img width="307" alt="image" src="https://github.com/user-attachments/assets/b046fbca-4ed9-4bce-89af-cc5efb86c715">

The second time I enter my route the `onBeforeRouteEnter` is empty
<img width="291" alt="image" src="https://github.com/user-attachments/assets/435e8cb4-660c-4da8-a0fb-78e62f958d85">

[This line](https://github.com/kitbagjs/router/blob/main/src/services/getRouteHooks.ts#L10) has a condition `isRouteEnter`, that is almost certainly the issue but I don't understand what it's trying to do assert.

I made a branch in our router-preview repo setup to reproduce
https://github.com/kitbagjs/router-preview/compare/issue-244-repro?expand=1

pleek91:
I wrote a test that I think reproduces what you're talking about there. Responding here mostly so I can remember what I found.

The issue is you didn't actually change routes. You navigated to `/profile` which triggered a rejection. But you're still on that same route. So when you push to `/profile` again its a route update not a route enter. You can see in this test that the update hook is called.

So I think its working as designed. But maybe not working as it needs to? I think the logic for enter vs update is correct. But it doesn't maybe match up with developers expectations when combined with rejections. 

```ts
test('test hooks, async () => {
[...]

Issue #193, opened 2024-06-08
Add support for customizable scroll behavior


stackoverfloweth:
- likely an option set when developer calls `createRouter`
- given context of `to` and `from`, allows user to set scroll position after route is loaded
- scroll position can be specific px value
- possibly set scroll position based on element
- will need to consider how to await elements that are loading from async component
- will need to update docs https://router.kitbag.dev/migrating-vue-router.html

https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll

stackoverfloweth:
I'm closing this as an issue that is mostly already handled by the browser and users should be able to solve more complex use cases with existing hooks if

PR #261, opened 2024-09-08
Minor fixes or rejection route type guard and types in hook runner methods

stackoverfloweth:
while exploring the idea of "system" or "default" hooks to [handle scroll position](https://github.com/kitbagjs/router/issues/193) I found a couple bugs. 

- move `isRejectionRoute` out of createRouterReject, not used anyways
- ensure that `isRejectionRouteSymbol` defaults to `false` when updating current route. We use `Object.assign` to update current route. All non-rejection routes do not have this property. The initial current route is a rejection route (not found). So the property is there when we create the router and never gets unset. So every route appeared to be a rejection route
- we were taking `T extends Routes` on the hook runner methods but T was available from the parent function and wasn't being overridden.

### DIFF
--- a/src/services/createCurrentRoute.ts
+++ b/src/services/createCurrentRoute.ts
@@ -19,7 +19,10 @@ export function createCurrentRoute(fallbackRoute: ResolvedRoute, push: RouterPus
   const route = reactive({ ...fallbackRoute })
 
   const updateRoute: ResolvedRouteUpdate = (newRoute) => {
-    Object.assign(route, { [isRejectionRouteSymbol]: false, ...newRoute })
+    Object.assign(route, { 
+      [isRejectionRouteSymbol]: false,
+      ...newRoute 
+    })
   }
 
   const currentRoute = route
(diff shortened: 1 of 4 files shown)


src/services/createCurrentRoute.ts
```diff
@@ -18,7 +19,7 @@ export function createCurrentRoute(fallbackRoute: ResolvedRoute, push: RouterPus
   const route = reactive({ ...fallbackRoute })
 
   const updateRoute: ResolvedRouteUpdate = (newRoute) => {
-    Object.assign(route, { ...newRoute })
+    Object.assign(route, { [isRejectionRouteSymbol]: false, ...newRoute })
```
pleek91, line 22: nit
```suggestion
    Object.assign(route, { 
      [isRejectionRouteSymbol]: false,
      ...newRoute 
    })
```

pleek91: Any chance this fixes https://github.com/kitbagjs/router/issues/244?

pleek91 (approved):

src/services/createCurrentRoute.ts
stackoverfloweth: @pleek91 just updated my test branch in router-preview to use this version of router and the issue seems to still exist unfortunately 😞

pleek91 (approved):

PR #264, opened 2024-09-09
fix: missing import isRoute in the docs

hi-reeve:
this PR fix a missing import and function calling in the docs route narrowing


pleek91 (approved):
Thank you!

PR #265, opened 2024-09-22
push/replace accepts option for hash, assembles on url

stackoverfloweth:
This PR adds a missing core piece of the URL, the hash.

## Writing the value

Previously, users could only set the hash by pushing with relative URL syntax with hash included

```ts
router.push('relative/foo#bar')
```

With this change, users can push (and replace) named routes with hash as part of the options

```ts
router.push('foo', {}, { hash: 'bar' })
```

This also works for `RouterLink`

```html
<router-link :to="resolve => resolve('foo', {}, {hash: 'bar'})">
```

And `useLink`

```ts
const { href } = useLink('foo', {}, { hash: 'bar' })
```

## Reading the value

For reading the current hash value, this PR adds `{ hash?: string }` property to `ResolvedRoute`. Which means `route.hash` is available everywhere resolved routes are used.

```ts
const route = useRoute()

route.hash // string | undefined
```

## Future enchancement

In a future PR, I'd like to add support for defining `hash` as part of your route definition. More to come on that.

### DIFF
--- a/src/services/urlAssembly.ts
+++ b/src/services/urlAssembly.ts
@@ -11,6 +11,7 @@ import { Route } from '@/types/route'
 type AssembleUrlOptions = {
   params?: Record<string, unknown>,
   query?: Record<string, string>,
+  hash?: string,
 }
 
 export function assembleUrl(route: Route, options: AssembleUrlOptions = {}): string {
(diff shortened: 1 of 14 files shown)


src/services/urlAssembly.ts
```diff
@@ -20,7 +21,10 @@ export function assembleUrl(route: Route, options: AssembleUrlOptions = {}): str
   const pathWithParamsSet = assemblePathParamValues(route.path, paramValues)
   const queryWithParamsSet = assembleQueryParamValues(route.query, paramValues)
 
-  return withQuery(`${hostWithParamsSet}${pathWithParamsSet}`, queryWithParamsSet, queryValues)
+  const url = withQuery(`${hostWithParamsSet}${pathWithParamsSet}`, queryWithParamsSet, queryValues)
+  const hash = options.hash ? `#${options.hash.replace(/^#/, '')}` : ''
```
pleek91, line 25: A `withHash` utility would be nice here

pleek91 (approved):
LGTM!

src/services/urlAssembly.ts
stackoverfloweth: I considered this but ultimately felt like this was simple enough to leave in place. `withQuery` actually covers quite a bit of complexity which I think justifies it's utility file (and handful of tests)

pleek91: I don't feel strongly. But a ternary and a regex in one line is something I'd hide behind a utility. Plus it reads nice to see withQuery then withHash. 

Up to you though, approved as is

PR #271, opened 2024-10-04
add `id` field to route

stackoverfloweth:
This PR adds a new `id: string` field to `Route` type. This field will be useful internally, but is also part of the public API so hopefully useful to end users as well. The `id` should be unique among all routes created, regardless of if/which routers get created with the route.

Currently the value of the id is created with an auto-incrementing closure function so each `id` is just a number. I chose to use `string` for the type in case we decide to pivot to something more complex in the future.


src/services/createRouteId.ts
pleek91, line 1: ❤️

pleek91 (approved):
Looks great!

PR #275, opened 2024-10-05
establish `PropStore`, move responsibility of executing `route.prop` functions

stackoverfloweth:
This PR establishes a centralized prop store, created (and provided) inside of `createRouter`. The PropStore exposes `setProps(route: ResolvedRoute)` and `getProps(route: ResolvedRoute, name?: string)`. This effectively shifts responsibility for finding props method on routes and executing from inside RouterView and into the store.

This theoretically improves performances slightly since props for descendants are all executed at the same time as the parents, whereas before they had to wait to be mounted. So if parents had async props, children prop functions had to wait for that to resolve first.

More importantly, this should benefit our broader goal of enabling prefetching props. With this layer, we should be able to have RouterLink (where prefetching happens) simply kick off prop functions in the prop store when a route has a link on the page and not waiting until the user actually navigates to it. 

In the future, for sure when we add prefetching, we will want a more elegant cache busting technic. Currently every time the store execute `setProps` for a route it assumes whatever else is cached was for unrelated routes and is cleared out. I thought it was better to start simple and make the cache layer more sophisticated as needed.


src/components/routerView.vue
```diff
@@ -24,6 +25,8 @@
   const route = useRoute()
   const rejection = useRejection()
   const depth = useRouterDepth()
+
+  const propStore = inject(propStoreKey)
```
pleek91, line 29: Nit: be nice to have a composition for this. usePropsStore or something.

pleek91: And it can throw a router not installed error rather than being possibly undefined.

src/components/routerView.vue
```diff
@@ -59,7 +62,7 @@
       // eslint-disable-next-line @typescript-eslint/no-unused-expressions, no-unused-expressions
       route.params
 
-      return componentUtil(component, () => props(route.params))
+      return componentUtil(component, () => props)
```
pleek91, line 65: My guess is since the props are called within the store this is no longer reactive to params changing. Is that true?

src/models/PropStore.ts
pleek91, line 1: I'll look at this move closely later. But looks good at first glance 👍

src/components/routerView.vue
stackoverfloweth: we have [this unit test](https://github.com/kitbagjs/router/blob/main/src/components/routerView.browser.spec.ts#L335) that I think is asserting that props are still reactive

pleek91: Hmm right, not sure how though lol.

src/components/routerView.vue
stackoverfloweth: 👍

src/compositions/usePropStore.ts
```diff
@@ -0,0 +1,12 @@
+import { inject } from 'vue'
+import { PropStore, propStoreKey } from '@/models/PropStore'
+
+export function usePropStore(): PropStore {
+  const store = inject(propStoreKey)
+
+  if (!store) {
+    throw 'PropStore not provided!'
```
pleek91, line 8: I think most of our compositions throw a `RouterNotInstalledError` if it cannot find something the router is supposed to provide when it is installed as a plugin.

src/models/PropStore.ts
```diff
@@ -0,0 +1,49 @@
...
+import { CreateRouteOptions, isWithComponent, isWithComponents } from '@/types/createRouteOptions'
+import { ResolvedRoute } from '@/types/resolved'
+import { MaybePromise } from '@/types/utilities'
+
+export const propStoreKey: InjectionKey<PropStore> = Symbol()
+
+type NamedComponentProps = { name: string, props?: (params: Record<string, unknown>) => MaybePromise<Record<string, unknown>> }
+
+export class PropStore {
+  private readonly store = new Map<string, MaybePromise<Record<string, unknown>>>()
```
pleek91, line 11: I think if we make this a `reactive` then the props in router view will automatically be reactive without the hack of accessing `route.params`.

pleek91: Separately, I'm wondering if we even need this specific of a type. Could it just be `store = new Map<string, unknown>()`?

src/models/PropStore.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (props) {
+          const key = this.getPropKey(route, name)
+          const value = props(route.params)
+
+          this.store.set(key, value)
+        }
+      })
+  }
+
+  public getProps(route: ResolvedRoute, name: string = 'default'): MaybePromise<Record<string, unknown>> | undefined {
```
pleek91, line 28: I think name should probably just be required and we can specifically pass in `default` if needed.

src/models/PropStore.ts
```diff
@@ -0,0 +1,49 @@
...
+  private readonly store = new Map<string, MaybePromise<Record<string, unknown>>>()
+
+  public setProps(route: ResolvedRoute): void {
+    this.store.clear()
+
+    route.matches
+      .flatMap(match => this.getComponentProps(match))
+      .forEach(({ name, props }) => {
+        if (props) {
+          const key = this.getPropKey(route, name)
```
pleek91, line 20: I think there's a bug hidden here with unnamed routes. Since `route` here is the resolved route, every key will have the same route id and props. Which means `name` is the only distinction here. So won't any matches that are unnamed share the same key? Which means the props will get overwritten. 

Doesn't `match` have an id as well? I think we want to use that id a not the name. 

Right? Or am I missing something.

src/components/routerView.vue
```diff
@@ -48,7 +51,7 @@
     }
 
     const component = getComponent(match)
-    const props = getProps(match)
+    const props = propStore.getProps(route, name)
```
pleek91, line 54: Coming back here after looking at `PropStore. Since this uses the `route` and not the `match` depth isn't factored in here right? Its going to get the props for the first component by the matching name right? 

We might not have good tests to accurately reflect this logic.

src/models/PropStore.ts
stackoverfloweth: yeah, this is a good catch

src/models/PropStore.ts
```diff
@@ -0,0 +1,56 @@
+import { InjectionKey, reactive } from 'vue'
+import { isWithComponent, isWithComponents } from '@/types/createRouteOptions'
+import { ResolvedRoute } from '@/types/resolved'
+import { Route } from '@/types/route'
+import { MaybePromise } from '@/types/utilities'
+
+export const propStoreKey: InjectionKey<PropStore> = Symbol()
+
+type ComponentProps = { id: string, name: string, props?: (params: Record<string, unknown>) => MaybePromise<Record<string, unknown>> }
```
pleek91, line 9: I think this is true
```suggestion
type ComponentProps = { id: string, name: string, props?: (params: Record<string, unknown>) => unknown }
```

src/models/PropStore.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class PropStore {
```
pleek91, line 11: nit/thought

We've limited the use of classes. Could we rewrite this as a factory? `createPropStore` that returns `{ setProps, getProps }`. That way we can destructure at implementation.

src/components/routerView.vue
pleek91: Wanna try removing line 63 and see if the tests still pass now that the prop store is using `reactive`?

src/models/PropStore.ts
stackoverfloweth: the only reason I opted for the class was to match your implementation of `RouteHookStore`, do you feel like the factory pattern is more valuable than being consistent?

pleek91: Uh the route hook store should probably be a factory as well... lol

stackoverfloweth: done! I actually didn't end up using destructuring since the whole "store" gets provided/injected but I still felt like the value of factory > classes is worthwhile even without the destructuring?

vite.config.js.timestamp-1728242264822-7ef6de8540d42.mjs
```diff
@@ -0,0 +1,51 @@
+// vite.config.js
```
pleek91, line 1: lol guessing this was an accident?

pleek91 (approved):
Other than the extra vite config LGTM!

vite.config.js.timestamp-1728242264822-7ef6de8540d42.mjs
stackoverfloweth: 🤷

pleek91 (approved):

PR #276, opened 2024-10-07
Add `context` to props callback, similar to hooks

stackoverfloweth:
This PR extends the useful functionality of triggering `push`, `replace`, and `reject` from `hooks` into the `props` callback.

Fundamentally the approach is the same, if a callback wants to use these methods it will actually throw a specific type of exception. So I wrapped the `setProps` call in `createRouter` in a try catch so that if the developers props callback throws one of these errors the router can perform the expected task.

Since the approach was the same as hooks I decided to extract the logic from hooks into a `createCallbackContext()` service used by both.

As a side quest this PR also moves and renames `RouterReject` from `Router.ts` to `RegisteredRouterReject` in `Register.ts` since it actually assumes the types for rejection are the registered rejections.

### DIFF
--- a/src/services/createCallbackContext.ts
+++ b/src/services/createCallbackContext.ts
@@ -1,0 +1,34 @@
+import { RouterPushError } from '@/errors/routerPushError'
+import { RouterRejectionError } from '@/errors/routerRejectionError'
+import { RegisteredRouterPush, RegisteredRouterReject, RegisteredRouterReplace } from '@/types/register'
+import { RouterPushOptions } from '@/types/routerPush'
+import { isUrl } from '@/types/url'
+
+export type CallbackContext = {
+  reject: RegisteredRouterReject,
+  push: RegisteredRouterPush,
+  replace: RegisteredRouterReplace,
+}
+
+export function createCallbackContext(): CallbackContext {
+  const reject: RegisteredRouterReject = (type) => {
+    throw new RouterRejectionError(type)
+  }
+
+  const push: RegisteredRouterPush = (...parameters: any[]) => {
+    throw new RouterPushError(parameters)
+  }
+
+  const replace: RegisteredRouterPush = (source: any, paramsOrOptions?: any, maybeOptions?: any) => {
+    if (isUrl(source)) {
+      const options: RouterPushOptions = paramsOrOptions ?? {}
+      throw new RouterPushError([source, { ...options, replace: true }])
+    }
+
+    const params = paramsOrOptions
+    const options: RouterPushOptions = maybeOptions ?? {}
+    throw new RouterPushError([source, params, { ...options, replace: true }])
+  }
+
+  return { reject, push, replace }
+}
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -3,6 +3,8 @@ import { App } from 'vue'
 import { RouterLink, RouterView } from '@/components'
 import { routerRejectionKey } from '@/compositions/useRejection'
 import { routerInjectionKey } from '@/compositions/useRouter'
+import { RouterPushError } from '@/errors/routerPushError'
+import { RouterRejectionError } from '@/errors/routerRejectionError'
 import { createCurrentRoute } from '@/services/createCurrentRoute'
 import { createIsExternal } from '@/services/createIsExternal'
 import { createMaybeRelativeUrl } from '@/services/createMaybeRelativeUrl'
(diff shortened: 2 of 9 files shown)


src/services/createCallbackContext.ts
stackoverfloweth, line 1: this is basically copy-pasta from hooks

src/services/createRouter.ts
```diff
@@ -123,7 +126,18 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
...
+    } catch (error) {
+      if (error instanceof RouterPushError) {
+        await push(...error.to as Parameters<RouterPush>)
+        return
+      }
+
+      if (error instanceof RouterRejectionError) {
+        setRejection(error.type)
+      }
+    }
```
stackoverfloweth, line 140: I don't love having all this logic inserted in an already big file and a big function. Open to suggestions

src/services/hooks.ts
```diff
@@ -76,8 +57,8 @@ export function createRouteHookRunners<const T extends Routes>(): RouteHookRunne
       const results = allHooks.map(callback => callback(to, {
         from,
         reject,
-        push: push as RegisteredRouterPush,
-        replace: replace as RegisteredRouterPush,
+        push,
+        replace,
```
stackoverfloweth, line 61: casting happens inside `createCallbackContext()`

src/services/createRouter.ts
```diff
@@ -183,7 +197,7 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
     return set(url, { ...options, state })
   }
 
-  const reject: RouterReject = (type) => {
+  const reject: RegisteredRouterReject = (type) => {
```
stackoverfloweth, line 200: this is just that rename/moving of the rejection type

src/services/createCallbackContext.ts
```diff
@@ -0,0 +1,34 @@
+import { RouterPushError } from '@/errors/routerPushError'
+import { RouterRejectionError } from '@/errors/routerRejectionError'
+import { RegisteredRouterPush, RegisteredRouterReject, RegisteredRouterReplace } from '@/types/register'
+import { RouterPushOptions } from '@/types/routerPush'
+import { isUrl } from '@/types/url'
+
+export type CallbackContext = {
+  reject: RegisteredRouterReject,
+  push: RegisteredRouterPush,
+  replace: RegisteredRouterReplace,
```
pleek91, line 10: We should update the return types for these to `never` that way typescript will correctly infer that they always throw an error. Then the `throw` in implementation won't be necessary for type guarding or highlighting unreachable code.

src/services/createRouter.ts
```diff
@@ -123,7 +126,18 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
         throw new Error(`Switch is not exhaustive for before hook response status: ${JSON.stringify(beforeResponse satisfies never)}`)
     }
 
-    propStore.setProps(to)
+    try {
+      propStore.setProps(to)
+    } catch (error) {
+      if (error instanceof RouterPushError) {
+        await push(...error.to as Parameters<RouterPush>)
```
pleek91, line 133: Curious why we need the `as` here. We have something similar for hooks but no `as`.

src/types/hooks.ts
```diff
@@ -29,10 +29,7 @@ export type RouteHookAbort = () => void
  */
 type RouteHookContext = {
   from: ResolvedRoute | null,
-  reject: RouterReject,
-  push: RegisteredRouterPush,
-  replace: RegisteredRouterReplace,
-}
+} & CallbackContext
```
pleek91, line 32: super nit: sorry it wouldn't let me suggest this. Just easier to read

```
type RouteHookContext = CallbackContext & {
  from: ResolvedRoute | null,
}
```

pleek91 (approved):
Couple suggestions but overall LGTM!

pleek91:
Actually, does this work with async props? My guess is no since the error won't be thrown until the props are awaited in RouterView (in the async props wrapper)

pleek91:
Thinking about this more last night in the perspective of prefetching. Because prefetched props will need to error without actually doing the rejection or push. 

What if we had the props store return the props or the error for a specific component? Then the current component utility could be responsible for doing the actual rejection or push.

src/services/createRouter.ts
stackoverfloweth: we do the casting [here](https://github.com/kitbagjs/router/blob/main/src/services/hooks.ts#L90)

pleek91:
Replaced by https://github.com/kitbagjs/router/pull/329

PR #277, opened 2024-10-09
Add prefetching for props

pleek91:
# Description
Adds the ability to prefetch props for routes. This follows the same config as component prefetching but is disabled by default.

This allows for props to be loaded prior to a route being visited so that when a router-link is clicked the props are already loaded (or have at least already started loading). 

```ts
const route = createRoute({
  name: 'user',
  path: '/users/[userId]',
  prefetch: { props: true },
  component: UserView,
  props: async ({ userId }) => {
    const response = await fetch(`/api/users/${userId}`)
    const user = await response.json()

    return { user }
  },
})
```

You can also set prefetch at the router or the router-link level same as for components.


src/components/routerView.vue
```diff
@@ -44,14 +45,18 @@
       return rejection.value.component
     }
 
-    const match = route.matches.at(depth)
+    const match: Route['matched'] | undefined = route.matches.at(depth)
```
pleek91, line 48: This type isn't totally necessary but since there's no registered router by default `route.matches` ends up being `any[]`. So `match` is also `any`. Adding this type inline gives us better type hints and checking.

src/compositions/useLink.ts
```diff
@@ -30,6 +32,10 @@ export type UseLink = {
    * True if route matches current URL. Route is the same as what's currently stored at `router.route`.
    */
   isExactMatch: Ref<boolean>,
+  /**
+  * Returns true if host does not match host stored on router instance
+  */
+  isExternal: Ref<boolean>,
```
pleek91, line 38: This is unrelated to this PR but I moved this here to useLink from router link. Keeping the link logic in the composition.

pleek91:
@stackoverfloweth this is going to conflict with https://github.com/kitbagjs/router/pull/276 because I reworked the prop store in this PR. But should be easy to work out the conflicts.

src/compositions/useLink.ts
```diff
@@ -106,24 +119,43 @@ export function useLink(
       routerPrefetch,
       linkPrefetch,
     })
+
+    props = prefetchPropsForRoute(route, {
+      getRouteProps,
```
stackoverfloweth, line 124: why pass this through? can't the `prefetchPropsForRoute` below just call it directly from the import?

src/compositions/useLink.ts
```diff
@@ -149,4 +181,29 @@ function prefetchComponentsForRoute(route: ResolvedRoute, { routerPrefetch, link
     }
   })
 
+}
+
+type PropsPrefetchContext = {
+  routerPrefetch: PrefetchConfig | undefined,
+  linkPrefetch: PrefetchConfig | undefined,
+  getRouteProps: PropStore['getRouteProps'],
+}
```
stackoverfloweth, line 190: I'm hoping we can take `getRouteProps` off this type, and then this type is almost identical to `PrefetchConfigs` in types/prefetch.ts. Do we need both types? Even if we need it, we should move this to prefetch.ts and have the types build off each other

src/compositions/useLink.ts
```diff
@@ -106,24 +119,43 @@ export function useLink(
...
-    push: (options?: RouterPushOptions) => router.push(href.value, {}, { ...optionsRef.value, ...options }),
-    replace: (options?: RouterReplaceOptions) => router.replace(href.value, {}, { ...optionsRef.value, ...options }),
+    isExternal,
+    push,
+    replace,
   }
 }
 
-type ComponentPrefect = {
+type ComponentPrefetchContext = {
```
stackoverfloweth, line 153: this type is almost identical to PrefetchConfigs in types/prefetch.ts. Do we need both types? Even if we need it, we should move this to prefetch.ts and have the types build off each other

src/services/createPropStore.ts
```diff
@@ -8,37 +8,109 @@ export const propStoreKey: InjectionKey<PropStore> = Symbol()
...
+   * The specific params for the actual route
+   */
+  params: unknown,
+}
+
+/**
+ * A unique identifier for a component's props based on the route id, view name, and stringified params
+ * Format: `${id}-${name}-${params}
+ */
+type RoutePropsKey = `${string}-${string}-${string}`
```
stackoverfloweth, line 30: I considered this as well, but felt like it was over engineering since we never later infer out any of the individual pieces

stackoverfloweth (commented):
feedback on the prop store sent via slack

PR #278, opened 2024-10-09
Fix some vue warnings appearing in test runs

pleek91:
# Description
Fixes some noisy warnings when running tests.


stackoverfloweth (approved):
❤️

PR #279, opened 2024-10-09
Remove requirement on `<Suspense>`  for async component props

pleek91:
# Description
When working on https://github.com/kitbagjs/router/pull/278 I noticed some `Suspense` warnings.

```
<Suspense> is an experimental feature and its API will likely change.
```

I started out attempting to just silence these. But researching more I'm not sure we even need to use suspense. You do not need to use suspense for async components. Its optional and only needs to be used for loading/error states. 

You do however seem to need `Suspense` when using `async setup`. Which we were using in the `AsyncPropsWrapper`. But we don't have to. Rewriting that component to not need top level await removes the need for `Suspense` altogether. 

I'm not 100% sure if this removes the possibility for leveraging `Suspense` for loading states. But we haven't really proved out some of those ideas around `Suspense` anyway. So removing the requirement of using `Suspense` seems like a good thing at least for now.

### DIFF
--- a/src/services/component.ts
+++ b/src/services/component.ts
@@ -1,5 +1,5 @@
 /* eslint-disable vue/one-component-per-file */
-import { AsyncComponentLoader, Component, FunctionalComponent, defineComponent, h } from 'vue'
+import { AsyncComponentLoader, Component, FunctionalComponent, defineComponent, h, ref } from 'vue'
 import { MaybePromise } from '@/types/utilities'
 
 type Constructor = new (...args: any) => any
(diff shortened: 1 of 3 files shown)


src/services/component.ts
```diff
@@ -52,10 +52,21 @@ function asyncPropsWrapper<TComponent extends Component>(component: TComponent,
...
-    async setup() {
-      const values = await props
+    setup() {
+      const values = ref()
 
-      return () => h(component, values)
+      // eslint-disable-next-line semi-style
+      ;(async () => {
+        values.value = await props
+      })()
```
pleek91, line 61: Iffy!

stackoverfloweth: self invoking functions brings me back to jquery days lol

stackoverfloweth (approved):
I loved how comparatively simple component.ts setup was when it was async but if this works and removes a bunch of warnings I'm okay with it.  I'm also okay with just having suspense warnings, since they really only show up when we're running our test suite, right?

pleek91:
@stackoverfloweth correct. But it also means that any user that adds async to their props now has a broken route until they add suspense. And then they get the suspense warnings. Figured for now it's better to make it "just work" until we have a better idea of how we want to leverage suspense.

Issue #269, opened 2024-09-29
Router instance has to be explicitly created after the store gets created


adaviloper:
## Problem
When following the docs, the router functionality is all contained within the `src/router/index.ts` file while the store is created in the `main.ts`.
In order to ensure the router has access to the store, we need to call `createRouter` in the `main.ts` file.

The problem here is that this then pollutes the contents of the `main.ts` file to also need to be where any router hooks would need to be defined.

PR #280, opened 2024-10-09
Delay initializing the router until it is installed or manually started

pleek91:
# Description
Because global hooks need to be defined on the router AFTER it is created it is problematic that the initial route is set immediately on creation. This can cause race conditions and import issues with other modules. 

By replacing the `initialized` property of the router with a `start` method and then automatically calling `start` when the router is installed we can make sure that the user is able to define global hooks and have them run on that initial route while avoiding module errors. 

Closes https://github.com/kitbagjs/router/issues/269

### DIFF
--- a/docs/api/types/Router.md
+++ b/docs/api/types/Router.md
@@ -38,13 +38,13 @@ go: (delta: number) => void;
 
 Moves the current history entry to a specific point in the history stack.
 
-### initialized
+### start
 
 ```ts
-initialized: Promise<void>;
+start: () => Promise<void>;
 ```
 
-A promise that resolves when the router is fully initialized.
+Initializes the router based on the initial route. Automatically called when the router is installed. Calling this more than once has no effect.
 
 ### isExternal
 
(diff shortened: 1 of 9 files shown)


stackoverfloweth (approved):
This is great and all, but don't we want to add an option to `createRouter` that prevents automatic calling of start? Seems like that defeats the purpose if dev wants to control when it starts?

stackoverfloweth:
Also, does this close https://github.com/kitbagjs/router/issues/269?

pleek91:
> This is great and all, but don't we want to add an option to `createRouter` that prevents automatic calling of start? Seems like that defeats the purpose if dev wants to control when it starts?

My thought is once it's installed it can start, we could add an option to prevent auto starting. But the main issue was starting immediately before it was even being used.

stackoverfloweth:
> > This is great and all, but don't we want to add an option to `createRouter` that prevents automatic calling of start? Seems like that defeats the purpose if dev wants to control when it starts?
> 
> My thought is once it's installed it can start, we could add an option to prevent auto starting. But the main issue was starting immediately before it was even being used. 

Oh that makes sense. I missed the part where the `start()` call was inside `install`. I thought it was just inside of `createRouter`

PR #281, opened 2024-10-10
Update the props store to support prefetching

pleek91:
# Description
WIP attempt at updating the props store to allow for prefetching. Updates the `setProps` method to accept a `prefetch` boolean and attempts to accurately clear stale props from the store. 

When `prefetch` is `true` props will be put into the store and then we exit. If `prefetch` is `false` then after putting props into the store we then go delete any props from the store that do not match the route that was passed in. 

In both cases if a key already exists in the store we do not execute the prop callback for that match. If a key already exists then it was prefetched and shouldn't be fetched again. 

@stackoverfloweth hopefully this aligns more with the mental model you had in mind for the props store and prefetching in general. 

Note: This doesn't fix the issue with a single boolean not accurately reflecting the fact that a route has multiple matches and each match might have a different prefetch value for props. The `prefetch` option might need to include a callback to determine if the current match should prefetch props based on the `PrefetchConfigs` for that match.

### DIFF
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -8,14 +8,14 @@ export const propStoreKey: InjectionKey<PropStore> = Symbol()
 type ComponentProps = { id: string, name: string, props?: (params: Record<string, unknown>) => unknown }
 
 export type PropStore = {
-  setProps: (route: ResolvedRoute, options: { prefetched: boolean }) => void,
+  setProps: (route: ResolvedRoute, options: { prefetch: boolean }) => void,
   getProps: (id: string, name: string, params: Record<string, unknown>) => unknown,
 }
 
 export function createPropStore(): PropStore {
   const store: Map<string, unknown> = reactive(new Map())
 
-  const setProps: PropStore['setProps'] = (route, { prefetched }) => {
+  const setProps: PropStore['setProps'] = (route, { prefetch }) => {
     const routeKeys: string[] = []
     const componentProps = route.matches.flatMap(match => getComponentProps(match))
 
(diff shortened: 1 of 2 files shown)


src/services/createPropStore.ts
```diff
@@ -2,43 +2,64 @@ import { InjectionKey, reactive } from 'vue'
...
 
 export const propStoreKey: InjectionKey<PropStore> = Symbol()
 
 type ComponentProps = { id: string, name: string, props?: (params: Record<string, unknown>) => unknown }
 
 export type PropStore = {
-  setProps: (route: ResolvedRoute) => void,
-  getProps: (id: string, name: string, params: unknown) => MaybePromise<unknown> | undefined,
+  setProps: (route: ResolvedRoute, options: { prefetched: boolean }) => void,
+  getProps: (id: string, name: string, params: Record<string, unknown>) => unknown,
```
pleek91, line 12: `MaybePromise<unknown> | undefined` is widened to `unknown` anyway.

src/services/createPropStore.ts
```diff
@@ -2,43 +2,64 @@ import { InjectionKey, reactive } from 'vue'
...
 }
 
 export function createPropStore(): PropStore {
-  const store = reactive(new Map<string, unknown>())
+  const store: Map<string, unknown> = reactive(new Map())
```
pleek91, line 16: `reactive` produces a mapped type because of ref unwrapping. Writing this way produces a simpler type for the `store`.

src/services/createPropStore.ts
```diff
@@ -2,43 +2,61 @@ import { InjectionKey, reactive } from 'vue'
...
 
-          store.set(key, value)
-        }
-      })
+      const key = getPropKey(id, name, route.params)
+      routeKeys.push(key)
+
+      if (store.has(key)) {
+        continue
+      }
```
pleek91, line 32: I'm not 100% sure skipping every time is the right thing to do here. But I think it makes sense?

PR #282, opened 2024-10-10
Update prop store to support prefetching

stackoverfloweth:
this PR is my pass at solving the same problems as [this PR](https://github.com/kitbagjs/router/pull/281).

most notably
- moves from an option `prefetch: true` to separate methods because the implementations are considerably different
- updates the store type to put `prefetched: true` on each store entry
- ensures that matches that have `prefetch: false` in their route definition will not be prefetched even when the parent tries to

### DIFF
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -8,14 +8,14 @@ export const propStoreKey: InjectionKey<PropStore> = Symbol()
 type ComponentProps = { id: string, name: string, props?: (params: Record<string, unknown>) => unknown }
 
 export type PropStore = {
-  setProps: (route: ResolvedRoute, options: { prefetched: boolean }) => void,
+  setProps: (route: ResolvedRoute, options: { prefetch: boolean }) => void,
   getProps: (id: string, name: string, params: Record<string, unknown>) => unknown,
 }
 
 export function createPropStore(): PropStore {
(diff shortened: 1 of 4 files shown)


src/services/createPropStore.ts
```diff
@@ -1,44 +1,63 @@
...
+      .filter(match => getPrefetchOption({ ...prefetch, routePrefetch: match.prefetch }, 'props'))
+      .flatMap(getComponentProps)
       .forEach(({ id, name, props }) => {
         if (props) {
-          const key = getPropKey(id, name, route.params)
+          const key = getPropKey(id, name, route)
           const value = props(route.params)
 
-          store.set(key, value)
+          store.set(key, { prefetched: true, value })
```
pleek91, line 29:
```suggestion
          store.set(key, value)
```

src/services/createPropStore.ts
```diff
@@ -1,44 +1,63 @@
...
+
+      if (!props || store.has(key)) {
+        return routeKeys
+      }
+
+      const value = props(route.params)
+
+      store.set(key, value)
+
+      return [...routeKeys, key]
```
pleek91, line 47: super nit: I'd just push the new key and return the same array. Rather than creating a new array every time.

pleek91 (approved):
We probably want to add some tests but this looks good. Nice job

PR #287, opened 2024-10-20
Prefetching route component props

pleek91:
# Description
Adds the ability to prefetch props for routes. This follows the same config as component prefetching but is disabled by default.

This allows for props to be loaded prior to a route being visited so that when a router-link is clicked the props are already loaded (or have at least already started loading). 

```ts
const route = createRoute({
  name: 'user',
  path: '/users/[userId]',
  prefetch: { props: true },
  component: UserView,
  props: async ({ userId }) => {
    const response = await fetch(`/api/users/${userId}`)
    const user = await response.json()

    return { user }
  },
})
```

You can also set prefetch at the router or the router-link level same as for components.


stackoverfloweth (approved):
🚀 🚀 🚀 🚀 🚀

PR #288, opened 2024-10-20
move prefetch logic out of useLink and into composables

stackoverfloweth:
this PR moves logic out of useLink.vue and into 2 new composables `usePrefetchedComponent` and `usePrefetchedProps`. The main goal here is reducing complexity of useLink but could also be useful in the future if additional places might use prefetching logic.

### DIFF
--- a/src/compositions/usePrefetchedComponent.ts
+++ b/src/compositions/usePrefetchedComponent.ts
@@ -1,0 +1,35 @@
+import { computed, MaybeRef, ref, watch } from 'vue'
+import { isWithComponent, isWithComponents } from '@/types/createRouteOptions'
+import { getPrefetchOption, PrefetchConfigs } from '@/types/prefetch'
+import { ResolvedRoute } from '@/types/resolved'
+import { isAsyncComponent } from '@/utilities/components'
+
+export function usePrefetchedComponent(route: MaybeRef<ResolvedRoute | undefined>, prefetch: MaybeRef<PrefetchConfigs | undefined>): void {
+  const routeRef = ref(route)
+  const prefetchRef = ref(prefetch)
+
+  const matchesToPrefetchComponents = computed(() => {
+    const matches = routeRef.value?.matches ?? []
+
+    return matches.filter(route => getPrefetchOption({
+      ...prefetchRef.value,
+      routePrefetch: route.prefetch,
+    }, 'components'))
+  })
+
+  watch(matchesToPrefetchComponents, routes => {
+    routes.forEach(route => {
+      if (isWithComponent(route) && isAsyncComponent(route.component)) {
+        route.component.setup()
+      }
+
+      if (isWithComponents(route)) {
+        Object.values(route.components).forEach(component => {
+          if (isAsyncComponent(component)) {
+            component.setup()
+          }
+        })
+      }
+    })
+  }, { immediate: true })
+}
--- a/src/compositions/usePrefetchedProps.ts
+++ b/src/compositions/usePrefetchedProps.ts
@@ -1,0 +1,30 @@
+import { MaybeRef, Ref, ref, watch } from 'vue'
+import { usePropStore } from '@/compositions/usePropStore'
+import { PrefetchConfigs } from '@/types/prefetch'
+import { ResolvedRoute } from '@/types/resolved'
+
+export type UsePrefetchedProps = {
+  prefetchProps: Ref<Record<string, unknown>>,
+  commitPrefetchedProps: () => void,
+}
+
+export function usePrefetchedProps(route: MaybeRef<ResolvedRoute | undefined>, prefetch: MaybeRef<PrefetchConfigs | undefined>): UsePrefetchedProps {
+  const routeRef = ref(route)
+  const prefetchRef = ref(prefetch)
+  const store = usePropStore()
+  const prefetchProps = ref<Record<string, unknown>>({})
+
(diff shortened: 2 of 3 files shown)


src/compositions/usePrefetchedComponent.ts
```diff
@@ -0,0 +1,35 @@
+import { computed, MaybeRef, ref, watch } from 'vue'
+import { isWithComponent, isWithComponents } from '@/types/createRouteOptions'
+import { getPrefetchOption, PrefetchConfigs } from '@/types/prefetch'
+import { ResolvedRoute } from '@/types/resolved'
+import { isAsyncComponent } from '@/utilities/components'
+
+export function usePrefetchedComponent(route: MaybeRef<ResolvedRoute | undefined>, prefetch: MaybeRef<PrefetchConfigs | undefined>): void {
+  const routeRef = ref(route)
+  const prefetchRef = ref(prefetch)
```
pleek91, line 9: nit: using `toValue` inside the `matchesToPrefetchComponents` is preferred

src/compositions/usePrefetchedComponent.ts
```diff
@@ -0,0 +1,35 @@
+import { computed, MaybeRef, ref, watch } from 'vue'
+import { isWithComponent, isWithComponents } from '@/types/createRouteOptions'
+import { getPrefetchOption, PrefetchConfigs } from '@/types/prefetch'
+import { ResolvedRoute } from '@/types/resolved'
+import { isAsyncComponent } from '@/utilities/components'
+
+export function usePrefetchedComponent(route: MaybeRef<ResolvedRoute | undefined>, prefetch: MaybeRef<PrefetchConfigs | undefined>): void {
```
pleek91, line 7: `MaybeRefOrGetter` is preferred over `MaybeRef` just because its more flexible.

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
+import { MaybeRef, Ref, ref, watch } from 'vue'
+import { usePropStore } from '@/compositions/usePropStore'
+import { PrefetchConfigs } from '@/types/prefetch'
+import { ResolvedRoute } from '@/types/resolved'
+
+export type UsePrefetchedProps = {
+  prefetchProps: Ref<Record<string, unknown>>,
```
pleek91, line 7: We probably don't need to expose this right?

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+  commitPrefetchedProps: () => void,
```
pleek91, line 8: nit: kinda like how simple "commit" would be here. 
```suggestion
  commit: () => void,
```

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function usePrefetchedProps(route: MaybeRef<ResolvedRoute | undefined>, prefetch: MaybeRef<PrefetchConfigs | undefined>): UsePrefetchedProps {
```
pleek91, line 11: Another spot to use `MaybeRefOrGetter`

One thing I did in my version of this was may this a single object so you could do this

```
usePrefetchedProps(() => ({
  route: route.value,
  routerPrefetch: router.prefetch,
  linkPrefetch: options.prefetch
}))
```

Don't feel strongly, just felt nice when I was playing with it.

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+  const routeRef = ref(route)
+  const prefetchRef = ref(prefetch)
+  const store = usePropStore()
+  const prefetchProps = ref<Record<string, unknown>>({})
```
pleek91, line 15: this isn't reactive so we should just use a `let` here.  Also, we're in a prefetch props composition. We could probably simplify some of these names. 
```suggestion
  const props: Record<string, unknown> = {}
```

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  watch([routeRef, prefetchRef], ([route, prefetch]) => {
```
pleek91, line 17: The single object works nicely here just because you end up with this

```
watch(option, ({ route, ...configs }) => {
  ...
})
```

src/compositions/usePrefetchedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!route) {
+      return
+    }
+
+    prefetchProps.value = store.getPrefetchProps(route, prefetch ?? {})
+  }, { immediate: true })
+
+  function commitPrefetchedProps(): void {
+    store.setPrefetchProps(prefetchProps.value)
+  }
```
pleek91, line 27: I really like the "commit" idea ❤️  I struggled with what to call this.

pleek91 (approved):
Overall LGTM. I'd be fine with merging this as is. Comments are mostly nits, thoughts, suggestions. Everything minor. 

I was approaching this as a single prefetching composition. Which might be preferred in the future once we get to prefetch strategies. But I like the separation because its easier to grok.

src/compositions/usePrefetchedProps.ts
pleek91: I see you are exposing it. I guess I wouldn't expose it and I wouldn't make it reactive. That would be my suggestion.

PR #305, opened 2024-11-09
fix bug with router base

### LINKED ISSUE #302: Component name not shown when mounting app under subfolder
This can be reproduced in https://github.com/kitbagjs/router-preview by modifying as following:

In _vite.config.ts_:

```ts
import { defineConfig } from 'vite'
import vue from '@vitejs/plugin-vue'
import vueDevTools from 'vite-plugin-vue-devtools'

const BASE_PATH = '/f'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [vue(), vueDevTools()],
  base: BASE_PATH,
})
```

In _router/index.ts_, change the `createRouter` line to:

```ts
export const router = createRouter([home, settings, profile, keys], {
  base: import.meta.env.BASE_URL
})
```

Now, all pages are shown as `routerView` when seeing in VueDevTools:

image

stackoverfloweth:
closes #302 

when `base` is supplied, the router acts as though the user created a parent above all other routes. When it's re-creating the rest of the routes, it was incorrectly passing in the `Route` type (from router) and not the `route.matched`, which is the options the user would have supplied when calling `createRoute`.  

Note, because there are so many overloads, the type is unhappy and I didn't want to mess with it right now 😄

### DIFF
--- a/src/services/insertBaseRoute.spec.ts
+++ b/src/services/insertBaseRoute.spec.ts
@@ -23,6 +23,7 @@ test.each([
 
 test('given value for base, returns routes with base prefixed', () => {
   const base = '/kitbag'
+
   const parent = createRoute({ name: 'c', path: '/c' })
   const routes = [
     createRoute({ name: 'a', path: '/a' }),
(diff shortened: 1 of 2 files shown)


pleek91 (approved):

pleek91 (approved):
Nice, I like this better.

Issue #299, opened 2024-11-08
How to build breadcrumb?


hongquan:
I am trying to build a breadcrumb for deep nested pages, to let user easily go to ancestor pages.

In `vue-router`, I use [`route.matched`](https://router.vuejs.org/api/interfaces/RouteLocationNormalizedTyped.html#matched) to get the URLs of the ancestor pages. But in `@kitbag/router`, there is no way. In `@kitbag/router`, the `route.matches` just return the data used for defining routes (in `routes.ts`), where `name` is of `string | undefined` type, and `path` is just a portion, not a full path.

stackoverfloweth:
@hongquan do your breadcrumbs use `router-link`? You should be able to use the `name` property but I can see how that's not a perfect solution. You'll need to use a type guard or casting to convince TS your name is valid. It also wouldn't work for any routes that have params. 

```ts
<template v-for="match in route.matches" :key="match.id">
  <router-link :to="resolve => resolve(match.name as typeof route['name'])">{{  match.name }}</router-link>
</template>
```

Note if a match has an undefined name, it's not a navigable route. 

We will see if we can find a better solution for you

stackoverfloweth:
@hongquan you can indeed use `router-link`, we just need a type guard to satisfy Typescript for the known name piece and we can pass `params` as well. Here's a complete working example.

```html
<template>
  <template v-for="match in matches" :key="match.id">
    <router-link :to="match">{{  match }}</router-link>
  </template>
  
  <router-view />
</template>
```

```ts
import { Url, isUrl, useRoute, useRouter } from '@kitbag/router';
import { computed } from 'vue';

const router = useRouter()
const route = useRoute()

function isRouteName(name: string | undefined): name is typeof route['name'] {
  return !!name && name !== 'NotFound'
}

const matches = computed(() => {
  return route.matches.reduce<Url[]>((matches, match) => {
    if(isRouteName(match.name)){
      const url = router.resolve(match.name, route.params)

      if(isUrl(url)){
        matches.push(url)
      }
    }

    return matches
  }, [])
})
```

Note, the check for `name !== "NotFound"` is checking that the current route is not a [rejection route](https://router.kitbag.dev/advanced-concepts/rejections.html).  Again, we have a type guard for that which we can include in the public API. For now checking the name will suffice since rejection routes use the reject type for a name.

hongquan:
@stackoverfloweth Thank you.

[...]

PR #308, opened 2024-11-10
simplified approach to ensuring Url type is returned from urlAssembly

stackoverfloweth:
In an attempt to simplify [common use cases](https://github.com/kitbagjs/router/issues/299) of `route.matches`, we discovered some functions that were returning `string` when they should be returning `Url`. 

In a [previous approach](https://github.com/kitbagjs/router/pull/303), we put it on the developer to make sure the routes they built satisfy the Url type when assembled.  However, we couldn't _really_ test routes since routes might have dynamic params. Plus developers probably aren't super familiar with what it means for a route to be "assembled". 

This PR takes a much simpler approach by just assembling the url, and then if the assembled url is NOT a valid Url type, it assumes the route must be relative and adds the forward slash.

This is no different from the query, how we automatically add the `?`, or hash how we always normalize the value to starting with `#`. "Paths" (when assembled) are expected the start with `/`, so if the developer omits it, we simply add it for them.

as a side-quest this PR removes complexity from url parts (path, host, query, hash). 
- these objects are simple containers now, no `toString` and no `hasValue` based on potentially stale values
- logic for cleaning and joining url parts is moved to "withPath", "withHost", etc.


src/services/urlAssembly.ts
```diff
@@ -19,23 +22,25 @@ type AssembleUrlOptions = {
...
+    (url) => withHash(url, hash),
+  ]
 
-  return `/${url}`
+  return assembly.reduce<Url>((url, join) => join(url), '/')
 }
 
 function assembleHostParamValues(host: Host, paramValues: Record<string, unknown>): string {
-  const value = host.toString()
+  const { value } = host
```
pleek91, line 43: nit: should probably just use `host.value` rather than destructuring. Same for path and query.

src/services/withPath.ts
```diff
@@ -0,0 +1,13 @@
...
+export function withPath(url: Url, path: string): Url
+export function withPath(url: string, path: string): Url
+export function withPath(url: string, path: string): Url {
+  const cleanPath = path.replace(/^\/*/, '')
+
+  if (isUrl(url)) {
+    return `${url}${cleanPath}`
+  }
+
+  return `/${url}${cleanPath}`
```
pleek91, line 12: Doesn't clean path never have a slash at the beginning so 8-10 won't ever do anything right?

src/services/withPath.ts
```diff
@@ -0,0 +1,13 @@
+import { isUrl, Url } from '@/types/url'
+
+export function withPath(url: Url, path: string): Url
+export function withPath(url: string, path: string): Url
+export function withPath(url: string, path: string): Url {
```
pleek91, line 5: This accepts a `url` but it assumes that the url doesn't have a query or hash. Should these utilities account for urls of different shapes? Right now they're order dependent, which is probably fine as long as they're only internal utilities.

src/services/withHost.ts
```diff
@@ -0,0 +1,19 @@
+import { isUrl, Url } from '@/types/url'
+import { stringHasValue } from '@/utilities'
+
+export function withHost(url: Url, host?: string): Url
```
pleek91, line 4: Interesting that this is the only "with" utility where the second argument is optional. Why is that?

pleek91 (commented):
Overall this looks good and I like the direction.

src/services/withPath.ts
stackoverfloweth: cleaning the path doesn't have any effect on the `url` property check on line 8

pleek91: Yeah I totally read that wrong. Carry on

src/services/withHost.ts
stackoverfloweth: agreed. I went through all the "with" utilities and made them much more similar. The only one that's different now is `withQuery` because it's also used in `createRouterResolve`. 

All of them now always return `Url`, host is sent as `string`, they all have an optional 2nd argument

src/services/withHash.ts
```diff
@@ -0,0 +1,11 @@
+import { Url } from '@/types/url'
+import { createMaybeRelativeUrl, maybeRelativeUrlToString } from '@/services/maybeRelativeUrl'
+
+export function withHash(url: Url, hash?: string): Url
+export function withHash(url: string, hash?: string): Url
+export function withHash(url: string, hash?: string): Url {
+  const { hash: previousHash, ...parts } = createMaybeRelativeUrl(url)
+  const cleanHash = hash ? `#${hash}` : ''
+
+  return maybeRelativeUrlToString({ hash: cleanHash, ...parts })
```
pleek91, line 10: This makes me feel like the url to string utility should be responsible for inserting the tokens like `#` and `?` and `://`.

PR #309, opened 2024-11-13
Update route.query to support all members of `URLSearchParams` and make the query writable on `RouterRoute`

pleek91:
# Description
`URLSearchParams` is a really nice interface for working with the query. But it cannot be accurately proxied which is how we trigger updates to the route in `RouterRoute`. But we can get 99% of the nice devX by wrapping `URLSearchParams` in a normal object that just aliases all the members. The only things that don't work with this are
- You cannot pass `ResolvedRouteQuery` into the constructor of `URLSearchParams`. But you can easily work around with by passing `query.toString()`
- You cannot directly iterate over the query like you could with `URLSearchParams`. But iterating over `query.values()` provides the same functionality. 

This PR adds all of the missing members to the `ResolvedRouteQuery` type and updates the `RouterRoute` to trigger route updates when `set`, `append`, or `delete` are called.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,20 +12,8 @@ jobs:
         id: setup
         uses: kitbagjs/actions-setup-project@main
 
-      - name: Run unit tests
+      - name: Run tests
         run: npm run test
-
-  type-test:
-    name: Type Tests
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Setup
-        id: setup
-        uses: kitbagjs/actions-setup-project@main
-
-      - name: Run unit tests
-        run: npm run test:types
 
   types:
     name: Types
--- a/src/services/createResolvedRouteQuery.ts
+++ b/src/services/createResolvedRouteQuery.ts
@@ -1,10 +1,30 @@
 import { ResolvedRouteQuery } from '@/types/resolvedQuery'
 
+/**
+ * Creates a dumb wrapper around URLSearchParams because URLSearchParams cannot be correctly be proxied to support writing params to the RouterRoute
+ */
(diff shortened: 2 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -12,21 +12,9 @@ jobs:
         id: setup
         uses: kitbagjs/actions-setup-project@main
 
-      - name: Run unit tests
+      - name: Run tests
```
stackoverfloweth, line 15: ❤️

src/services/createResolvedRouteQuery.ts
```diff
@@ -1,10 +1,30 @@
...
+      params.delete(...args)
+    },
+    toString: (...args) => params.toString(...args),
+    forEach: (...args) => {
+      params.forEach(...args)
+    },
+    entries: (...args) => params.entries(...args),
+    keys: (...args) => params.keys(...args),
+    values: (...args) => params.values(...args),
+    has: (...args) => params.has(...args),
```
stackoverfloweth, line 28: wouldn't this also work?

```suggestion
    get: params.get,
    getAll: params.getAll,
    set: params.set,
    append: params.append,
    delete: params.delete,
    toString: params.toString,
    forEach: params.forEach,
    entries: params.entries,
    keys: params.keys,
    values: params.values,
    has: params.has,
```

stackoverfloweth (approved):

src/services/createResolvedRouteQuery.ts
pleek91: We tried this but the unit tests failed. Pretty sure its because this basically destructures the URLSearchParams and loses the context.

PR #310, opened 2024-11-15
Extend query support to match URLSeachParams

stackoverfloweth:
We discovered [in a recent PR](https://github.com/kitbagjs/router/pull/308) that the type we're using to pass query values (`Record<string, string>`) was limiting our ability to mirror native behavior of built in `URLSeachParams` type. Specifically, the object syntax doesn't allow for multiple entries with the same key.

This PR updates our `QueryRecord` type to match any value accepted by the constructor of `URLSeachParams`.

### DIFF
--- a/docs/api/components/RouterLink.md
+++ b/docs/api/components/RouterLink.md
@@ -7,10 +7,12 @@ RouterLink component renders anchor tag (`<a>`) for routing both within the SPA 
 | Parameter | Type |
 | :---- | :---- |
 | to | [`Url`](/api/types/Url) \| `(resolve: RouterResolve) => Url` |
-| query | `ConstructorParameters<typeof URLSearchParams>[0]` \| `undefined` |
+| query | `QuerySource` \| `undefined` |
 | hash | `string` \| `undefined` |
 | replace | `boolean` \| `undefined` |
 | prefetch | `boolean` \| `PrefetchConfigOptions` \| `undefined` |
+
+[QuerySource](/api/types/QuerySource)
 
 [RouterResolve](/api/types/RouterResolve)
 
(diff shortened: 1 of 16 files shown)


docs/api/components/RouterLink.md
```diff
@@ -7,7 +7,7 @@ RouterLink component renders anchor tag (`<a>`) for routing both within the SPA
 | Parameter | Type |
 | :---- | :---- |
 | to | [`Url`](/api/types/Url) \| `(resolve: RouterResolve) => Url` |
-| query | `Record<string, string>` \| `undefined` |
+| query | `ConstructorParameters<typeof URLSearchParams>[0]` \| `undefined` |
```
pleek91, line 10: We should be using the `QueryRecord` (which I'm suggesting we rename to `QuerySource` everywhere and not this constructor type.

src/types/query.ts
```diff
@@ -4,6 +4,8 @@ import { Param } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { isRecord } from '@/utilities/guards'
 
+export type QueryRecord = ConstructorParameters<typeof URLSearchParams>[0]
```
pleek91, line 7: I tested this out and I think we need to just copy the type from the constructor rather than trying to reference it. This type here gets widened by typescript and ends up accepting some invalid input. For example it thinks this value is fine when it should error (and does error for URLSearchParams)
image

Lets do something like this
```ts
type QuerySource =
  | URLSearchParams
  | string
  | Record<string, string | readonly string[]>
  | Iterable<[string, string]>
  | readonly [string, string][]
```

Suggesting `QuerySource` because `QueryRecord` implies its only a record.

pleek91 (approved):
❤️

PR #312, opened 2024-11-17
Make `route.query` assignable so the entire query can be replaced

pleek91:
# Description
https://github.com/kitbagjs/router/pull/309 Make it possible to modify the query directly via the route. But making multiple operations makes multiple `push` calls. You could call `route.update({}, { query: newQuery })` but since we already have the mechanisms to make parts of the route writable we can allow users to assign the query as a nice syntax that is a better devx. 

This PR makes this possible
```ts
const route = useRoute()

route.query = 'foo=bar'

// or any other QuerySource such as URLSearchParams
route.query = new URLSearchParams({ foo: 'bar' })
```

### DIFF
--- a/src/services/createRouter.spec.ts
+++ b/src/services/createRouter.spec.ts
@@ -247,7 +247,7 @@ test('query is writable', async () => {
 
   await start()
 
-  route.query = new URLSearchParams('foo=bar&foo=baz')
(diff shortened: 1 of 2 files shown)


stackoverfloweth (approved):
nice! Good to cover all these possible use cases

PR #314, opened 2024-11-17
`useQueryValue` composition to make working with query values that are not params just as easy

pleek91:
# Description
Query params are useful for route matching and declaring known route state. But sometimes the query state cannot be known ahead of time, is dynamic, or for other reasons cannot be declared on the route. 

`useQueryValue` aims to make working with query values that are not params as feature rich as params themselves. 

Working with simple values
```ts
const { value } = useQueryValue('foo')
//           ?^ Ref<string | null>
```

Working with arrays
```ts
const { values } = useQueryValue('foo')
//           ?^ Ref<string[]>
```

Using params
```ts
const { value, values } = useQueryValue('foo', Number)

value // Ref<number | null>

values // Ref<number[]>
```

A `remove` callback is also returned to make deleting query keys simple
```ts
const { value, remove } = useQueryValue('foo')

value.value = 'bar'

// ?foo=bar

remove()

// query is empty
```

### DIFF
--- a/src/compositions/useQueryValue.browser.spec.ts
+++ b/src/compositions/useQueryValue.browser.spec.ts
@@ -1,0 +1,184 @@
+import { createRouter } from '@/services'
+import { createRoute } from '@/services/createRoute'
+import { expect, test } from 'vitest'
+import { useQueryValue } from './useQueryValue'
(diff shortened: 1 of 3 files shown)


stackoverfloweth (approved):
love this! 😍

PR #315, opened 2024-11-17
Make route.params assignable so all params can be replaced

pleek91:
# Description
Sugar around `route.update({ foo: 'bar', fuz: 'buz' })` to make it possible to just write

```ts
route.params = { foo: 'bar', fuz: 'buz' }
```


stackoverfloweth (approved):
🎉

PR #319, opened 2024-11-19
Create dependabot.yml

pleek91:
Lets github auto update our dev dependencies


stackoverfloweth (approved):
Will this force our users to be in latest versions as well? I'm not sure we want to enforce that without a good reason?

PR #329, opened 2024-11-26
Add the ability to push, replace, and reject within prop callback via `CallbackContext

pleek91:
# Description
Builds on https://github.com/kitbagjs/router/pull/328 to allow passing `CallbackContext` into route prop callbacks. 

Props are different from hooks in that they are not a blocking part of the navigation lifecycle. And they can also be prefetched. This means that context errors can be thrown
- Before navigation has even started
- During navigation
- After navigation has finished

Props can also be async. So we cannot know if a context error was thrown until the promise resolves. So to account for that we'll convert all props to async just for the response of `setProps` so we can then handle any context errors asynchronously.

For prefetched props, we can store any sync errors as the props value in the props store. Then when `setProps` is called we can rethrow the original error.  For async props that are prefetched the value is awaited (same as any props that were not prefetched).


pleek91:
@stackoverfloweth got this figured out I think. I think vitest was correctly calling out the unhandled rejections. Even though the code was working I believe an error would still end up in the console. So what I ended up doing is when calling prop callback we'll always make it "successful". 

For sync props
- Success: just store the props
- Error: Store the error

For async props
Store the promise with a `.catch(error => error)` so that rather than the promise rejecting it'll be fulfilled but the value will be the error instance that was thrown.

stackoverfloweth (approved):
Nice solution to that exception being thrown. Might be a good candidate for some sort of logging for our end users if we ever embark on our "add debugging" task

PR #330, opened 2024-11-28
Props callback context review

stackoverfloweth:
further implements the suggestion of a shared parent class so checking instances is as simple as

```ts
} catch (error) {
  if (error instanceof AfterCallbackContextError) {
    return error.response
  }

  throw error
}
```


stackoverfloweth:
created off wrong-branch

PR #331, opened 2024-11-28
shared instanceof suggestion

stackoverfloweth:
further implements the suggestion of a shared parent class so checking instances is as simple as

```ts
} catch (error) {
  if (error instanceof AfterCallbackContextError) {
    return error.response
  }

  throw error
}
```


pleek91:
@stackoverfloweth my opinion is to close this. I'd rather keep the callback context errors agnostic of where in the lifecycle they are used. But if you feel strongly let's discuss 👍

Issue #340, opened 2024-12-03
Build component props from query/state?


hagabaka:
When defining a route, currently you can [provide component props based on the route params](https://kitbag-router.netlify.app/core-concepts/component-props.html):
<blockquote>

```ts
const user = createRoute({
  props: (params) => {
    return { foo: params.bar }
  }
})
```

</blockquote>

But the props getter doesn't have access to route query and state, even though as they can be passed when navigating to a route like params. One reason why I would like to to use state instead of a param is to pass a complex value for a component prop, without having a long URL.

pleek91:
@hagabaka great suggestion. Let me look into adding some more information about the route itself into the [callback's context](https://kitbag-router.netlify.app/core-concepts/component-props.html#context)

pleek91:
@hagabaka take a look at the linked PR and let me know what you think. I believe that should handle your use case and its much more flexible.

hagabaka:
> @hagabaka take a look at the linked PR and let me know what you think. I believe that should handle your use case and its much more flexible.

The pull request looks good, however it seems like only String states are passed to the props function, and not other types. Here are two test cases showing this (could be added to src/components/routerView.browser.spec.ts):
```ts
test('Binds props and attrs from route state (string)', async () => {
  const route = createRoute({
    name: 'propsFromStringState',
    path: '/propsFromStringState',
    component: echo,
    state: { value: String },
    props: ({ state: { value = 'missing state' } }) => ({ value }),
  })

  const router = createRouter([route], {
    initialUrl: '/',
  })

  await router.start()

  const root = {
    template: '<RouterView/>',
  }

  const app = mount(root, {
    global: {
      plugins: [router],
    },
  })

  await router.push('/propsFromStringState', { state: { value: 'hello' } })
  expect(app.html()).toBe('hello')
})

test('Binds props and attrs from route state (number)', async () => {
  const route = createRoute({
    name: 'propsFromNumberState',
    path: '/propsFromNumberState',
    component: echo,
    state: { value: String },
    props: ({ state: { value = 'missing state' } }) => ({ value }),
  })

  const router = createRouter([route], {
    initialUrl: '/',
  })

  await router.start()

  const root = {
    template: '<RouterView/>',
  }

  const app = mount(root, {
    global: {
      plugins: [router],
    },
  })

  await router.push('/propsFromNumberState', { state: { value: 42 } })
  expect(app.html()).toBe('42')
})
```
```
   ✓ Binds props and attrs from route state (string)
[...]

PR #341, opened 2024-12-04
Pass the entire resolved route rather than just the params

pleek91:
# Description
Previously the props callback when creating a route passed just the params for the route in as the first argument. But there are lots of useful pieces on the route itself such as the query, hash, state, etc. Passing in the entire route offers much more functionality. 

**This is a breaking change**
Any uses of the props callback will need to change from 
```ts
createRoute({
   ...,
   props: (params) => {
    return { value: params.foo }
  }
})
```
to
```ts
createRoute({
   ...,
   props: (route) => {
    return { value: route.params.foo }
  }
})
```

Closes https://github.com/kitbagjs/router/issues/340


stackoverfloweth (approved):
🚢

PR #342, opened 2024-12-04
Enhance prefetching with prefetch strategies

pleek91:
# Description
Prefetching is configured with a boolean. Either something is prefetched or its not prefetched. But there's no control over when something is prefetched. Prefetch strategies allow for configuring when something should be prefetched. In this PR I'm introducing two strategies
- eager - prefetch immediately when link is mounted
- lazy - prefetch once link is in view

When configuring prefetching you can now use `boolean | PrefetchStrategy | PrefetchConfig`
```ts
createRoute({
   ...,
   prefetch: true,
})

createRoute({
   ...,
   prefetch: 'lazy',
})

createRoute({
   ...,
   prefetch: {
    components: 'eager',
    props: 'lazy'
  },
})
```


src/compositions/usePrefetching.ts
```diff
@@ -1,51 +1,97 @@
...
 export function usePrefetching(config: MaybeRefOrGetter<UsePrefetchingConfig>): UsePrefetching {
-  let props: Record<string, unknown> = {}
+  const prefetchedProps = new Map<PrefetchStrategy, Record<string, unknown>>()
+  const element = ref<HTMLElement>()
 
   const { getPrefetchProps, setPrefetchProps } = usePropStore()
+  const { observe, unobserve, isElementVisible } = useVisibilityObserver()
+
+  const commit: UsePrefetching['commit'] = () => {
+    const props = Array.from(prefetchedProps.values()).reduce((acc, value) => {
```
pleek91, line 26: Use a full word for `acc`
```suggestion
    const props = Array.from(prefetchedProps.values()).reduce((acc, value) => {
```

src/compositions/usePrefetching.ts
```diff
@@ -1,51 +1,97 @@
...
+    setPrefetchProps(props)
+  }
+
+  onMounted(() => {
+    if (!element.value) {
+      console.warn('The usePrefetching composition will not work correctly if the element ref is not bound.')
+      return
+    }
+
+    observe(element.value)
```
pleek91, line 41: worth noting that currently useVisibilityObserver isn't reactive to a template ref changing.

stackoverfloweth (approved):
🤘

PR #356, opened 2024-12-09
Automatic api docs generataion

pleek91:
# Description
The original api docs were generated using typedoc and then heavily modified manually. These docs have been hard to maintain and have gotten out of date over time as we've changed internals but haven't updated api docs consistently. 

This PR updates the typedoc's configs, plugins, and introduces a custom npm script that automates all of the polish and organization we want that typedoc doesn't support out of the box. 

I've then regenerated the api docs and updated the regular docs to fix any broken links. 

Notes: 
We use to have api docs for `router-view` and `router-link`. However I wasn't able to get typedoc to generate any useful docs for these. I've remove them from the api docs but will add new dedicated doc pages for them in the regular docs in a separate PR.


scripts/api.js
pleek91, line 1: This file is 99% written by ai. Don't judge. Works like a charm lol.

PR #359, opened 2024-12-10
Add support for `suspense` when mounting  components

pleek91:
# Description
When defining routes you can define a props function that can be async. By default this does not require [suspense](https://vuejs.org/guide/built-ins/suspense.html) to be used. But with this update to the component utils suspense will be used if it is detected. 

This makes it possible to use suspense's fallback for loading states. 
```html
<Suspense>
  <template #fallback>
    <span>Loading</span>
  </template>

  <RouterView/>
</Suspense>
```


pleek91:
Hmm not sure why the test runner doesn't like `Promise.withResolvers`. Added `ESNext` to the tsconfig and it runs locally 🤔

src/services/component.ts
```diff
@@ -21,31 +20,27 @@ type ComponentPropsGetter<TComponent extends Component> = () => MaybePromise<Com
...
         if (values instanceof Error) {
           return '' 
         }
 
         if (isPromise(values)) {
+          // there isn't a way to check if suspense is used in the component without accessing a private property
+          // @ts-expect-error
+          if(instance?.suspense) {
+            return h(suspenseAsyncPropsWrapper(component, values))
+          }
```
pleek91, line 42: @stackoverfloweth this is the best way I could figure out to check if the component is within a suspense boundary. This is a private part of the vue api. But if `suspense` is `null` that means there's no suspense boundary.

stackoverfloweth (approved):
nice 👏

PR #360, opened 2024-12-10
Make sure docs use `as const` when necessary

pleek91:
# Description
When defining an array of routes its important to use `as const` otherwise the type returned by `createRouter` won't be as useful


stackoverfloweth (approved):
good call, trips us up, certainly will trip up others

PR #363, opened 2024-12-13
Fix prefetching components with strategy `lazy` throwing component context error

pleek91:
# Description
Prefetching components using a `lazy` strategy was throwing this error
```
TypeError: Cannot read properties of null (reading 'ids'
```
This was because to prefetch components we had been calling `setup` manually on vue's AsyncComponentWrapper component. However the setup for AsyncComponentWrapper requires a component context to hook up suspense functionality. So calling setup outside the router link's component setup (when the lazy strategy is executed) there was no instance. 

Fixing this by calling the [loader function](https://github.com/vuejs/core/blob/201936f9a3909ae9dca4e131e1f5ad5a1c0feb77/packages/runtime-core/src/apiAsyncComponent.ts#L121) directly.

### DIFF
--- a/src/components/routerLink.browser.spec.ts
+++ b/src/components/routerLink.browser.spec.ts
@@ -647,4 +647,62 @@ describe('prefetch props', () => {
 
     expect(callback).toHaveBeenCalled()
   })
+
+  test('components are not prefetched until link is visible when prefetch is lazy', async () => {
+    let loaded = false
+
+    const routeA = createRoute({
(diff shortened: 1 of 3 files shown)


stackoverfloweth (approved):
Cool! Is async loader a private property from Vue?

pleek91:
> Cool! Is async loader a private property from Vue?

Yes, but I think for detecting some built in vue behaviors we're going to have to access the private api on occasion.

PR #364, opened 2024-12-14
add "href" property to ResolvedRoute

stackoverfloweth:
this is step 1 of a couple different initiatives.

1.) we need a key that's unique to handle transitions when params change, currently using `route.id` but that won't change when params change. With this new `href` property, we will have a better candidate. (this is really just a 1 line docs change, so I decided to commit it here)

2.) we want to avoid "widening" the route inside of RouterLink, currently users might use an explicit route name as part of the `ToCallback` which uses RouterResolve and returns a Url. Once we "widen" from an explicit route name to a Url, we've found that will sometimes actually point to a different route now, causing a bad DX and confusion.

### DIFF
--- a/docs/advanced-concepts/transitions.md
+++ b/docs/advanced-concepts/transitions.md
@@ -7,7 +7,7 @@ The `RouterView` component provides a default slot to support layering in a [Vue
   <template #default={ component }>
     <transition name="fade">
       <component :is="Component" />
-    </transition>G
+    </transition>
   </template>
 </router-view>
 ```G
--- a/docs/advanced-concepts/transitions.md
+++ b/docs/advanced-concepts/transitions.md
@@ -7,19 +7,19 @@ The `RouterView` component provides a default slot to support layering in a [Vue
   <template #default={ component }>
     <transition name="fade">
       <component :is="Component" />
-    </transition>
+    </transition>G
   </template>
 </router-view>
-```
+```G
 
 ## Avoid Reusing Components
 
-Vue will occasionally reuse components if a route change ends up rendering the same underlying component. You avoid this by [using the `key` attribute](https://vuejs.org/api/built-in-special-attributes.html#key).
+Vue will occasionally reuse components if a route change ends up rendering the same underlying component. You can avoid this by [using the `key` attribute](https://vuejs.org/api/built-in-special-attributes.html#key).
 
 ```html{3}
 <router-view>
   <template #default={ component, route }>
-    <transition name="fade" :key="route.key">
+    <transition name="fade" :key="route.href">
       <component :is="Component" />
     </transition>
   </template>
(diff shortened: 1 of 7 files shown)


docs/advanced-concepts/transitions.md
```diff
@@ -7,19 +7,19 @@ The `RouterView` component provides a default slot to support layering in a [Vue
   <template #default={ component }>
     <transition name="fade">
       <component :is="Component" />
-    </transition>
+    </transition>G
```
pleek91, line 10:
```suggestion
    </transition>
```

src/services/getResolvedRouteForUrl.ts
```diff
@@ -39,5 +40,6 @@ export function getResolvedRouteForUrl(routes: Routes, url: string, state?: unkn
     params: getRouteParamValues(route, url),
     state: getStateValues(route.state, state),
     hash,
+    href: asUrl(url)
```
pleek91, line 43: Seems a little odd we need asUrl. But I get why.

docs/advanced-concepts/transitions.md
```diff
@@ -7,19 +7,19 @@ The `RouterView` component provides a default slot to support layering in a [Vue
...
 ## Avoid Reusing Components
 
-Vue will occasionally reuse components if a route change ends up rendering the same underlying component. You avoid this by [using the `key` attribute](https://vuejs.org/api/built-in-special-attributes.html#key).
+Vue will occasionally reuse components if a route change ends up rendering the same underlying component. You can avoid this by [using the `key` attribute](https://vuejs.org/api/built-in-special-attributes.html#key).
 
 ```html{3}
 <router-view>
   <template #default={ component, route }>
-    <transition name="fade" :key="route.key">
+    <transition name="fade" :key="route.href">
```
pleek91, line 22: Did you add href to RouterRoute?

src/services/createRouterReject.ts
```diff
@@ -46,18 +46,17 @@ export function createRouterReject({
...
-    const resolved = {
+    return {
       id: route.id,
       matched: route,
       matches: [route],
       name: type,
       query: createResolvedRouteQuery(''),
       params: {},
       state: {},
+      href: '/',
```
pleek91, line 57: Hmm hard coding this to '/' is interesting. This is really only used for NotFound when the initialUrl doesn't match anything right?

docs/advanced-concepts/transitions.md
stackoverfloweth: yep

src/services/createRouterReject.ts
stackoverfloweth: yes that's where we use this reject route. The actual `href` of the rejection route is probably never actually used anywhere but required to satisfy the type

pleek91 (approved):

PR #365, opened 2024-12-14
combined router resolve and router find

stackoverfloweth:
follow-up from https://github.com/kitbagjs/router/pull/364

We have accumulated a handful of utilities that seemed to serve similar purposes. Those utilities also had a tendency to use the `Url` as the source of truth, so even if it was given an explicit route `name`, it would often "widen" to a `Url`. I say "widen" here because sometimes users will provide multiple routes that assemble to the same `Url`.

This PR 
- removes `router.find`. This was used to translate a `Url` into a `ResolvedRoute | undefined` from inside `useLink`.  That responsibility is now part of `router.resolve`.
- updates contract for `router.resolve` to return a `ResolvedRoute` instead of a `Url`.


src/components/routerLink.vue
```diff
@@ -58,13 +72,25 @@
...
+    if (route.value) {
+      return useLink(route.value, options)
+    }
+
+    if (isUrl(props.to)) {
+      return useLink(props.to, options)
+    }
+
+    throw new Error('Failed to resolve route in RouterLink.')
+  }
```
stackoverfloweth, line 92: is it okay to move composable calls into a function like this?

src/services/createResolvedRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+    id: route.id,
+    matched: route.matched,
+    matches: route.matches,
+    name: route.name,
+    query: createResolvedRouteQuery(search),
+    params: getRouteParamValues(route, href),
+    state: getStateValues(route.state, options.state),
+    hash,
+    href
+  }
```
stackoverfloweth, line 32: this is shared logic for creating ResolvedRoute from a `url` or from a `route` and `params`. Previously inside `getResolvedRouteForUrl`

src/services/createRouter.spec.ts
```diff
@@ -462,3 +441,171 @@ test('initial route is not set until the router is started', async () => {
 
   expect(router.route.name).toBe('root')
 })
+
+describe('router.resolve', () => {
```
stackoverfloweth, line 445: this tests come from the now defunct `createRouterFind` and `createRouteResolve` spec files

src/utilities/urlSearchParams.ts
```diff
@@ -10,3 +13,9 @@ export function combineUrlSearchParams(aParams: URLSearchParams | QuerySource, b
 
   return combinedParams
 }
+
+export function appendQuery(url: Url, query: QuerySource): Url {
+  const { searchParams, ...parts } = parseUrl(url)
+
+  return createUrl({ ...parts, searchParams: combineUrlSearchParams(searchParams, query) })
+}
```
stackoverfloweth, line 21: when we're using `router.push` (or replace) and given a `url` we no longer call `resolve`. The only thing we really needed to do was combine the query that might have been provided in the resolve options arg. Now we use this function instead.

This is also used inside of `resolve` when given a non-external `url`. That way the `href` property of the ResolvedRoute includes the query from the options.

src/services/createRouter.ts
```diff
@@ -180,61 +181,79 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
...
+    maybeOptions?: RouterResolveOptions,
+  ) => {
+    if (!isUrl(source)) {
+      const params = paramsOrOptions ?? {}
+      const options: RouterResolveOptions = maybeOptions ?? {}
+      const match = routes.find((route) => route.name === source)
+
+      if (!match) {
+        throw new RouteNotFoundError(String(source))
+      }
```
pleek91, line 196: I don't think we need to throw an error do we? RouterResolve's return type is `ResolvedRoute | undefined`. So I think we can just return `undefined` here.

src/services/createRouterResolve.ts
pleek91, line 1: Did it just not make sense to abstract this anymore?

src/services/createResolvedRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Route } from "@/types/route"
+import { parseUrl } from "@/services/urlParser"
+import { assembleUrl } from "@/services/urlAssembly"
+import { RouterResolveOptions } from "@/types/RouterResolve"
+import { createResolvedRouteQuery } from "@/services/createResolvedRouteQuery"
+import { getRouteParamValues } from "@/services/paramValidation"
+import { getStateValues } from "@/services/state"
+import { asUrl } from "@/types/url"
+
+export function createResolvedRoute(route: Route, url: string, options?: RouterResolveOptions): ResolvedRoute | undefined
```
pleek91, line 11: Why does this accept both a route and a url? I assumed we'd create the url within this method.

src/components/routerLink.vue
```diff
@@ -12,13 +12,15 @@
...
   import { PrefetchConfig } from '@/types/prefetch'
   import { RegisteredRouter } from '@/types/register'
   import { RouterPushOptions } from '@/types/routerPush'
   import { Url, isUrl } from '@/types/url'
+  import { ResolvedRoute } from '@/types/resolved'
+  import { useRouter } from '@/compositions/useRouter'
+  import { UseLink, useLink } from '@/compositions/useLink'
 
-  type ToCallback = (resolve: RegisteredRouter['resolve']) => Url
+  type ToCallback = (resolve: RegisteredRouter['resolve']) => ResolvedRoute | undefined
```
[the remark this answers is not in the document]
stackoverfloweth: love that 👍

pleek91 (approved):

src/services/createResolvedRoute.ts
stackoverfloweth: we decided to duplicate the few lines of createResolvedRoute into getResolvedRouteForUrl so that this doesn't have the overloads

src/services/createRouterResolve.ts
stackoverfloweth: we moved away from createRouteThing with createRouterPush, replace, find, etc. Now everything is inside createRouter. Mostly so that we have the context of things like `set` method and `isExternal`

src/services/createRouter.spec.ts
```diff
@@ -406,28 +407,6 @@ test('query.values is reactive', async () => {
   expect(values.value).toMatchObject(['foo1', 'bar1', 'foo2'])
 })
 
-test('given an array of Routes, combines into single routes collection', () => {
```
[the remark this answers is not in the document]
stackoverfloweth: still supported, just a harder test to write without being able to spy on `createRouterResolve`

PR #366, opened 2024-12-15
Update `useVisibilityObserver` to be reactive to ref changes

pleek91:
# Description
Since `useLink` can be used to create custom link elements, its possible the element template ref won't be available on mount or that it might be dynamic. Updating `useVisibilityObserver` to be more resilient to changes to the element template ref


stackoverfloweth (approved):
super clean

PR #367, opened 2024-12-15
Add a test to ensure the exact resolved route is used for `router-link`

pleek91:
# Description
Adds a test to make sure that even when there are two routes with the same url and a link is created that the route specified by the resolve is used.


stackoverfloweth (approved):
⬆️

PR #374, opened 2024-12-18
add ResolvedRouteQuery to QuerySource union

### LINKED ISSUE #372: Can't reuse query from original route when replacing route
In a `onBeforeRouteUpdate` I would like to replace the current route to a child route and therefore I need to pass on the full state of the route in the replace. I am trying it to do it that way:
```
onBeforeRouteUpdate: (to, { replace }) => {
  replace(to, route.params, {
      query: route.query,
      hash: route.hash,
      state: route.state,
    })
}
```

but there is an typescript error: 
`
Type 'ResolvedRouteQuery' is not assignable to type 'string | string[][] | Record<string, string> | URLSearchParams | undefined'.
 Type 'ResolvedRouteQuery' is missing the following properties from type 'URLSearchParams': size, sort, [Symbol.iterator]
`

stackoverfloweth:
closes #372

This approach might be too heavy handed as it adds a private symbol to our `ResolvedQuery` type to so that we can discriminate it from the `URLSearchParams` it's intended to mirror.

This means we can continue to pass `QuerySource` args into `new URLSearchParams()` constructor and only call `toString()` if we know it's our `ResolvedQuery`

### DIFF
--- a/src/services/createResolvedRouteQuery.ts
+++ b/src/services/createResolvedRouteQuery.ts
@@ -1,4 +1,4 @@
-import { ResolvedRouteQuery } from '@/types/resolvedQuery'
+import { isResolvedRouteQuerySymbol, ResolvedRouteQuery } from '@/types/resolvedQuery'
 
 /**
  * Creates a dumb wrapper around URLSearchParams because URLSearchParams cannot be correctly be proxied to support writing params to the RouterRoute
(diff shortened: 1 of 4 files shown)


pleek91:
@stackoverfloweth i want to see if we can update the ResolvedRouteQuery to satisfy URLSearchParams and then we wouldn't need to do any special handling.

stackoverfloweth:
closing in favor of simpler https://github.com/kitbagjs/router/pull/375

PR #376, opened 2024-12-19
coalescing around "wrapper" for mounted const in tests

stackoverfloweth:
normalize to `wrapper` > `app`


pleek91 (approved):
This is the sort of thing we probably don't need to over index on. But LGTM

stackoverfloweth:
> This is the sort of thing we probably don't need to over index on. But LGTM

Totally. But also my propensity to over index for consistency sometimes yields good things. You take the good with the bad lol

PR #378, opened 2024-12-19
hash should always be populated in ResolvedRoute

stackoverfloweth:
this actually created an issue with reactivity inside `createCurrentRoute`, or at least that's my theory

we create our reactive with

```ts
  const route = reactive({ ...fallbackRoute })
```

which is later updated with

```ts
  const updateRoute: ResolvedRouteUpdate = (newRoute) => {
    Object.assign(route, {
      [isRejectionRouteSymbol]: false,
      ...newRoute,
    })
  }
```

when fallbackRoute doesn't have `hash`, calling updateRoute with a newRoute with a hash was NOT assigning hash to `router.route.hash`.

### DIFF
--- a/src/services/createRouterReject.ts
+++ b/src/services/createRouterReject.ts
@@ -55,6 +55,7 @@ export function createRouterReject({
       params: {},
       state: {},
       href: '/',
+      hash: '',
       [isRejectionRouteSymbol]: true,
     }
   }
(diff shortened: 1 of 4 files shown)


pleek91 (approved):
Interesting. Maybe objects.assign doesn't work for adding new properties. Which kinda makes sense. Either way making has always a string makes sense from a devx perspective.

stackoverfloweth:
> Interesting. Maybe objects.assign doesn't work for adding new properties. Which kinda makes sense. Either way making has always a string makes sense from a devx perspective. 

Whats interesting is I tested`object.assign`and it DOES add new properties. My guess it's something inside reactive.

PR #380, opened 2024-12-19
URLSearchParams constructor does not require toString be called

stackoverfloweth:
insignificant change pulled off of https://github.com/kitbagjs/router/pull/371


pleek91 (approved):
❤️

PR #384, opened 2024-12-20
remove plugin from Router, add Register tests

stackoverfloweth:
We had reports of registered rejections having the wrong type in [Kitbag's Discord](https://discord.gg/zw7dpcc5HV). 

We've traced the type issue down to using Vue's `Plugin` type as part of the `Router` definition. This PR removes that (while still satisfying the ObjectPlugin syntax). This PR also adds some type tests around some of the common Register types used often.


pleek91 (approved):
LGTM!

PR #391, opened 2024-12-23
Add new "intent" prefetch strategy

pleek91:
# Description
The new "intent" prefetch strategy triggers prefetching when a router link emits a [focusin](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event) or [mouseover](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event) event. This allows for prefetching to happen right before a user navigates to the route.


src/components/routerLink.browser.spec.ts
```diff
@@ -329,7 +329,7 @@ test('isMatch correctly matches parent when sibling has the same url', async ()
     parent: parentRoute,
     name: 'child',
     path: '/child',
-    component: () => h(routerLink, { to: (resolve) => resolve('parent') }, 'parent'),
+    component: () => h(routerLink, { to: (resolve) => resolve('parent') }, () => 'parent'),
```
pleek91, line 332: Unrelated fix for a vue warning when running tests.

src/compositions/useVisibilityObserver.ts
```diff
@@ -25,6 +25,12 @@ export function useVisibilityObserver(element: Ref<Element | undefined>): UseVis
     }
   }, { immediate: true })
 
+  onUnmounted(() => {
+    if (element.value) {
+      observer.unobserve(element.value)
+    }
+  })
+
```
pleek91, line 33: unrelated fix to make sure elements are no longer observed after they are unmounted.

pleek91: Actually I'm going to make a separate PR for this so it will show up as a fix in the change log.

stackoverfloweth (approved):
mad kudos dude, this is a complicated feature implemented with +136-3 🏆

PR #393, opened 2024-12-23
Make sure start cannot be called multiple times

pleek91:
# Description
This was how I originally intended to write this but at the time `Promise.withResolvers` wasn't working because of a missing ts lib. That has since been fixed so fixing this up so even if you call `start(); start()` it won't actually execute twice.


stackoverfloweth (approved):
love to see `withResolvers` is available 🎉

PR #394, opened 2024-12-28
Add route change hooks

pleek91:
# Description
There has been a bit of confusion about the `onBeforeRouteEnter` hook when it pertains to parent routes. Specifically when going from a rejection back to a route (the route hasn't actually changed, so an update hook is called). And when navigating from siblingA to siblingB (again, an update hook is called). 

This PR adds `onBeforeRouteChange` and `onAfterRouteChange` hooks which are called on both enter and update. This allows a single hook to be used whenever a navigation happens. 

Putting this up as a draft because I'm not 100% sold on this. I THINK it makes sense. But the terminology of "change" vs "update" isn't very clear. They seem synonymous at first glance. I also considered something like "onBeforeLocationUpdate" to distinguish between a hook based on what was happening to the route vs what was happening to the location. But considering these new hooks work exactly the same as the other hooks, introduction the "location" keyword seemed heavy handed. 

I think this closes https://github.com/kitbagjs/router/issues/244

### DIFF
--- a/src/models/RouteHooks.ts
+++ b/src/models/RouteHooks.ts
@@ -4,7 +4,9 @@ export class RouteHooks {
   public onBeforeRouteEnter = new Set<BeforeRouteHook>()
   public onBeforeRouteUpdate = new Set<BeforeRouteHook>()
   public onBeforeRouteLeave = new Set<BeforeRouteHook>()
+  public onBeforeRouteChange = new Set<BeforeRouteHook>()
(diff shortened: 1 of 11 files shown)


stackoverfloweth (approved):
After thinking more about the naming convention, I'm comfortable with `update` being the more granular and `change` encompassing both enter + update. Feels like a nice change

pleek91:
Going to close this since its out of date and I'm not sure its the correct approach anyway.

PR #395, opened 2024-12-28
Rewrite docs for core concepts

pleek91:
# Description
Updates the docs for core concepts by breaking them out into the following pages and writing more concise but hopefully more complete docs about each concept. Also fills in other various gaps in the api docs and composables. 
- Routes
- External Routes
- Params
- Router
- Router Route
- Navigation


docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+So far we have only looked at params in the `path` property. But params can also be used in the `query` property. When using params in the `query` the param goes where you expect the value to be in the url's search string.
+
+```ts {3}
+const events = createRoute({
+  name: 'events',
+  query: 'highlight=[?highlight]',
+})
+```
+
+The param name and the search key do not have to be the same. In this example the url search stringwould like like `?highlight=hello` and the param would have the value `hello`. But the param's name is `term`.
```
pleek91, line 162:
```suggestion
The param name and the search key do not have to be the same. In this example the url search string would like like `?highlight=hello` and the param would have the value `hello`. But the param's name is `term`. 
```

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+
+const months = ['january', 'february', 'march', 'april', 'may', 'june', 'july', 'august', 'september', 'october', 'november', 'december'] as const
+
+const monthParam = createParam((value, { invalid }) => {
+  if (months.includes(value as any)) {
+    return value
+  }
+
+  invalid(`Invalid month: ${value}`)
+})
```
stackoverfloweth, line 77: the way this is defined makes sense if you want your value to be `january | february | ...` but as its written you'd still have `string` since you're `if(months.includes...)` is not a type guard.

I prefer either 1. take this suggestion and let the param type be `string`. or 2 maybe just "import" a type guard and assume devs know what type guards do

```suggestion
const months = ['january', 'february', 'march', 'april', 'may', 'june', 'july', 'august', 'september', 'october', 'november', 'december']

const monthParam = createParam((value, { invalid }) => {
  if (months.includes(value)) {
    return value
  }

  invalid(`Invalid month: ${value}`)
})
```

stackoverfloweth: also, don't we want to `throw invalid(...)`? otherwise our param also appears to return possibly `undefined`

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+  }),
+})
+```
+
+By default custom params will use `toString` when serializing. But you can use a Get/Set param to define how the param should be serialized and deserialized. For example we can make the month param capitalized when parsing and lowercase when serializing.
+
+```ts
+import { createParam } from '@kitbag/router'
+
+const months = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'] as const
```
stackoverfloweth, line 93: same "issue" here as above, so whatever you decide there should be applied here as well

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+
+const monthParam = createParam({
+  get: (value, { invalid }) => {
+    const month = value.charAt(0).toUpperCase() + value.slice(1)
```
stackoverfloweth, line 97: you might consider just using

`import { capitalize } from '@/utilities'` and keep the example as concise as possible

stackoverfloweth: might also be worth adding the little "highlighting" format thing to this code block like you do elsewhere on this page

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+    invalid(`Invalid month: ${month}`)
+  },
+  set: (value) => {
+    return value.toLowerCase()
+  },
+})
+```
+
+## Default Values
+You can define a default value for a param type when creating it. This value will be used when the param is optional and not provided in the url. An optional param of type `String` will have the type `string | undefined` when being accessed. But when a default value is provided it will have the type `string`, even if the value is missing from the url.
```
stackoverfloweth, line 112:
```suggestion
You can define a default value for a param type when defining  it. This default value will be used when the param is optional and not provided in the url. An optional param of type `String` will have the type `string | undefined` when being accessed. But when a default value is provided it will have the type `string`, even if the value is missing from the url.
```

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+const monthParam = createParam({
+  get: (value, { invalid }) => {
+    ...
+  },
+  set: (value) => {
+    ...
+  },
+  defaultValue: 'january',
+})
+```
```
stackoverfloweth, line 124: personally I'd prefer to just skip sharing this option or bury it somewhere inside of "deep dive" section. For 99% of users I just point them towards `withDefault`.

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+
+```ts {7}
+import { createParam, withDefault } from '@kitbag/router'
+
+const monthParam = createParam((value, { invalid }) => {
+  ...
+})
+
+const monthParamWithDefault = withDefault(monthParam, 'january')
+```
```
stackoverfloweth, line 150: I'd probably have skipped this too. It's neat and hopefully will bring delight to someone who tries it but probably doesn't need to be explicitly mentioned as a core-concept

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Query Params
+So far we have only looked at params in the `path` property. But params can also be used in the `query` property. When using params in the `query` the param goes where you expect the value to be in the url's search string.
```
stackoverfloweth, line 153:
```suggestion
So far the examples have only used params in the `path` property.  When using params in the `query`, the param goes where you expect the value to be in the url's search string.
```

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+
+```ts {3}
+const events = createRoute({
+  name: 'events',
+  query: 'highlight=[?highlight]',
+})
+```
+
+The param name and the search key do not have to be the same. In this example the url search string would like like `?highlight=hello` and the param would have the value `hello`. But the param's name is `term`.
```
stackoverfloweth, line 162:
```suggestion
The param name and the search key do not have to be the same. In this example the url search string might be `?highlight=hello`, which has a `term` param with the value `hello`.
```

stackoverfloweth: you mention the param name is `term`, but I don't think that's true based on the example provided

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
...
+
+const routes = [
+  createRoute(...),
+  createRoute(...),
+] as const
+
+const router = createRouter(routes)
+```
+
+## Installing
```
stackoverfloweth, line 18: is it worth repeating this from quick start?

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
...
+import App from './App.vue'
+
+const router = createRouter(...)
+const app = createApp(App)
+
+app.use(router)
+app.mount('#app')
+```
+
+## Type Safety
```
stackoverfloweth, line 33: is it worth repeating this from quick start?

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
+# Router
+The router is responsible for matching routes to urls and updating the browser history. It also provides a way to navigate to routes, access the current route, and provides some routing utilities. See [Router](/api/types/Router.md) type for more information.
+
+## Creating a Router
```
stackoverfloweth, line 4: is it worth repeating this from quick start?

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+```
+
+## Router Options
+There are several options you can pass to the router to customize its behavior. See the [RouterOptions](/api/types/RouterOptions.md) type for more information and a list of options.
+
+```ts
+const router = createRouter(routes, {
+  base: '/app',
```
stackoverfloweth, line 49: thoughts on expanding this example?

```suggestion
  base: '/app',
  initialUrl: '/home',
  historyMode: 'memory',
  prefetch: { components: true, props: false },
  rejections: { authNeeded: ... },
```

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
...
+})
+```
+
+## Router Instance
+The router created by `createRouter` can be used directly in your application. You can also access the router instance using the [useRouter](/composables/useRouter.md) composable within your components.
```
stackoverfloweth, line 54: as you might have picked up on by now, I'd prefer our core-concepts to be as concise as possible. I personally wouldn't even mention the fact that you _could_ import your router, let's just tell them about `useRouter` as a good pattern for accessing the installed router

docs/core-concepts/router.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Install
+Installs the router into a Vue application instance.
+
+```ts
+router.install(app)
+```
+
+### Start
+Initializes the router based on the initial route. Automatically called when the router is installed. Calling this more than once has no effect.
```
stackoverfloweth, line 142:
```suggestion
Idempotent function that initializes the router based on the initial route. Automatically called when the router is installed.
```

docs/core-concepts/navigation.md
```diff
@@ -0,0 +1,103 @@
...
+
+::: code-group
+```ts [Router]
+// type safe
+router.push('blogPost', {
+  blogPostId: 1,
+})
+
+// not type safe
+router.push('/blogPost/1')
```
stackoverfloweth, line 31:
```suggestion
// type safe ✅
router.push('blogPost', {
  blogPostId: 1,
})

// not type safe ⚠️
router.push('/blogPost/1')
```

docs/core-concepts/navigation.md
```diff
@@ -0,0 +1,103 @@
...
+<router-link to="(resolve) => resolve('blog')">Blog</router-link>
+
+<!-- replace -->
+<router-link to="(resolve) => resolve('blog')" replace>Blog</router-link>
+```
+:::
+
+
+## Resolved Routes
+A [ResolvedRoute](/api/types/ResolvedRoute) is the base of what makes up the [Router Route](/core-concepts/router-route). It represents a [route](/core-concepts/routes) that has been matched to a specific location. It includes any params, state, query, and hash values for that location. Resolved routes are how Kitbag Router ensures type safety when navigating. There are a few ways to get a resolved route.
```
stackoverfloweth, line 71: this last sentence says 
> There are a few ways to get a resolved route.

but then goes on to show 1 way to get a resolved route and 2 other tabs that reference the 1st tab and also somewhat repeat the code-group above

docs/core-concepts/router-route.md
```diff
@@ -0,0 +1,137 @@
+# Router Route
+The current route is represented by the "Router Route". See the [RouterRoute](/api/types/RouterRoute.md) type for more information.
```
stackoverfloweth, line 2:
```suggestion
The current route is represented by the [RouterRoute](/api/types/RouterRoute.md).
```

docs/core-concepts/router-route.md
```diff
@@ -0,0 +1,137 @@
...
+
+The route can be accessed using the [useRoute](/composables/useRoute.md) composable within your components. It is also available on the router instance as the `route` property.
```
stackoverfloweth, line 4:
```suggestion
This current route is accessed using the [useRoute](/composables/useRoute.md) composable within your components. It is also available on the router instance as the `route` property.
```

docs/core-concepts/navigation.md
stackoverfloweth: I'm struggling to see the value in conveying the "resolved route" concept to users at all

stackoverfloweth (commented):
Appreciate you going through docs, I like a lot of these changes

docs/core-concepts/routes.md
```diff
@@ -0,0 +1,177 @@
...
+
+const blogPost = createRoute({
+  parent: blog,
+  name: 'blogPost',
+  path: '/:postId',
+})
+```
+
+## Component
+The `component` property is used to define the component that will be rendered when the route is active. Component is optional and a [RouterView](/components/router-view) will be rendered if no component is provided.
```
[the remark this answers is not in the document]
pleek91: Isn't that suggestion a comma splice?

docs/core-concepts/routes.md
```diff
@@ -0,0 +1,177 @@
...
+
+const home = createRoute({
+  name: 'home',
+  path: '/',
+  component: HomeView,
+})
+```
+
+## Components
+The `components` property is used to define multiple components for named views.
```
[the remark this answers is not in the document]
pleek91: Clarified in https://github.com/kitbagjs/router/pull/395/commits/ce42573e539d3ba05f9067b6370581c29ad4cd35

docs/core-concepts/routes.md
```diff
@@ -0,0 +1,177 @@
...
+  path: '/',
+  components: {
+    default: HomeView,
+    sidebar: HomeSidebar,
+  },
+})
+```
+
+## Props
+The `props` property is used to define props for route components. It must be a callback function that returns an object. Everything returned from the callback will be bound to the component.
```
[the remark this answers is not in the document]
pleek91: Yep forgot to mention those somewhere. Updated in https://github.com/kitbagjs/router/pull/395/commits/87e9ad58fe96ea33f8b78a734305202c129a71f4

docs/core-concepts/routes.md
```diff
@@ -0,0 +1,177 @@
...
+  path: '/',
+  component: HomeView,
+  props: () => ({
+    userId: 1,
+  }),
+})
+```
+
+## Meta
+The `meta` property is used to define metadata for the route. Meta is optional and can be used to define static metadata for the route.
```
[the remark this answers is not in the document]
pleek91:
```suggestion
The `meta` property is used to define metadata for the route. Meta is optional and can be used to define static metadata for the route to reference in the [router route](/core-concepts/router-route) or in [hooks](/advance-concepts/hooks)
```

docs/core-concepts/routes.md
```diff
@@ -0,0 +1,177 @@
...
+  path: '/',
+  component: HomeView,
+  meta: {
+    title: 'Home',
+  },
+})
+```
+
+## State
+The `state` property is used to define optional data that can stored on the route in the browser's history. State is always optional, but it can be used to pass data to the route when navigating or to preserve state when navigating away from the route.
```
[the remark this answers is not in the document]
pleek91: I definitely wouldn't use the term "metadata" for state. Metadata is static. State is dynamic. Personally I like it as written as it describes the use cases of passing state or preserving state.

docs/core-concepts/external-routes.md
```diff
@@ -0,0 +1,11 @@
+# External Routes
```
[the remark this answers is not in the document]
pleek91: Do you feel strongly its not a core concept? It reads very nicely to go from routes to external routes. 

As far as meta vs state. I'd rather describe what they are rather than be descriptive on how to use them. Which I think we can do in the route docs.

docs/core-concepts/external-routes.md
```diff
@@ -0,0 +1,11 @@
...
+You can also define external routes. External routes are just like normal routes except they must be provided a `host` and they can only have the `name`, `host`, `path`, `query`, and `hash` properties. An external route's host can be provided by a parent. When navigating to an external route, the router will simply push the url to the browser's history.
```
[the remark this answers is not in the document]
pleek91: "With Kitbag Router" is pretty redundant since we're on the kitbag router docs. The "greatest parent" piece is also a little confusing IMO. How about this version? 

```suggestion
External routes allow you to define routes for things outside of your application. External routes have all the same functionality of normal routes, including [params](/core-concepts/params). When navigating to an external route, the router will simply push the url to the browsers history. 

When defining an external route a `host` must be provided, or it can be inherited from a `parent`.
```

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+const events = createRoute({
+  name: 'events',
+  path: path('/events/[year]/[?month]', {
+    year: Number,
+  }),
+})
+```
+
+:::info Route Matching
+When using a param type, if the url cannot be parsed as the specified type it will not be matched. For example if the url was `/events/two-thousand-and-twenty-four/september` it would not be matched because `two-thousand-and-twenty-four` is not a number.
```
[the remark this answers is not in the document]
pleek91:
```suggestion
When using a param type, if the url cannot be parsed as the specified type it will not be matched. For example if the url was `/events/two-thousand-and-twenty-four/september` it would not be matched because `two-thousand-and-twenty-four` is not a number. See [route matching](/advanced-concepts/route-matching) for more information.
```

docs/core-concepts/params.md
```diff
@@ -0,0 +1,168 @@
...
+| -------- | ------- | --------------------------------------------------------------------------------------------- |
+| `String` | ✓       | Any string value                                                                              |
+| `Number` |         | Any value that can be parsed with the `Number` constructor                                    |
+| `Boolean`|         | The literal values `"true"` or `"false"`                                                      |
+| `Date`   |         | Any value that can be parsed with the `Date` constructor. Uses `toISOString` when serializing |
+| `RegExp` |         | A literal regex expression                                                                    |
+| `JSON`   |         | Any value that can be parsed with the `JSON.parse` method                                     |
+
+## Custom Param Types
+You can define your own param types using the `createParam` utility. This serves three purposes:
```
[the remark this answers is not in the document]
pleek91:
```suggestion
You can define your own param types using the `createParam` utility. 

This serves three purposes:
```

docs/core-concepts/params.md
pleek91:
```suggestion
const months = ['january', 'february', 'march', 'april', 'may', 'june', 'july', 'august', 'september', 'october', 'november', 'december']

const monthParam = createParam((value, { invalid }) => {
  if (months.includes(value)) {
    return value
  }

  throw invalid(`Invalid month: ${value}`)
})
```

docs/core-concepts/params.md
pleek91: This is splitting hairs a bit, I'd take your version but there's an extra space in it lol

docs/core-concepts/params.md
pleek91: Updated in https://github.com/kitbagjs/router/pull/395/commits/cb79ed88c5cd4659a79d30ada19f49dba618d4c3

docs/core-concepts/params.md
pleek91: Updated in https://github.com/kitbagjs/router/pull/395/commits/cb79ed88c5cd4659a79d30ada19f49dba618d4c3

docs/core-concepts/params.md
pleek91: This section pretty concisely explains both ways of defining a default value. And I think defining it with `createParam` is just as valid as using `withDefault`. So I'd prefer to keep this.

docs/core-concepts/params.md
pleek91: I think it helps explain how the `withDefault` utility works by creating a new param. I'd like to keep it unless you feel strongly.

docs/core-concepts/params.md
pleek91:
```suggestion
So far the examples have only used params in the `path` property. When using params in the `query`, the param goes where you expect the value to be in the url's search string.
```

docs/core-concepts/router.md
pleek91: Idempotent is one of those words that always makes me feel like people are trying to sound smart and the average person doesn't actually know what it means lol. Do you feel strongly about not just explaining how it works?

docs/core-concepts/router.md
pleek91: I think the quick start is always going to be duplicate of other docs. This page will likely grow to have more information and feel less repetitive. But I think these docs aren't repeated from quick start, quick start repeats them from here. If that makes sense.

docs/core-concepts/router.md
pleek91: Lets discuss this in general on this comment https://github.com/kitbagjs/router/pull/395#discussion_r1899187620

docs/core-concepts/router.md
pleek91: I don't feel strongly, but I thought linking to the api docs for this would mean it would always be up to date and one less piece we have to manually manage.

docs/core-concepts/router.md
pleek91: I agree on being concise but also complete. I don't want to have a core concept and a deep dive page about the same topic. 

Also agree that `useRouter` is the preferred way of getting the router instance. Which is why that's the example here. I didn't intent to imply you could or should export/import it. Just that the return value of `createRouter` IS the router instance.

docs/core-concepts/navigation.md
pleek91: Resolved routes are pretty critical to how the router works and how navigation works. So I think looking under the hood a bit here is nice. The idea behind the code group is showing (through repetition) that these all work the same way. Switching between tabs its easy to identify the pieces that are the same and have that "ahh" moment. 

I think its specifically useful considering router link provides a callback called `resolve` and we have other router link examples that show it. Without this section IMO there's a bit of confusion. "what is resolve",  "why do I need to use this function", etc. 

I thought you'd be proud how concise this was 😂 let's discuss

docs/core-concepts/routes.md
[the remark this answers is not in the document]
stackoverfloweth: I could see the argument for that. I just thought the "and" part sounded awkward.

docs/core-concepts/external-routes.md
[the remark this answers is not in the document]
stackoverfloweth: I understand not wanting to bloat docs with how we think you should use things. 

I'm in favor of getting users to basic understanding as quickly and simply as possible. Majority of users will probably never use external routes.

docs/core-concepts/external-routes.md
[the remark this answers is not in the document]
stackoverfloweth: I'm good with your version. I mostly didn't like a sentence that starts with "You" but I do kinda like using "Kitbag Router" throughout. Yes they're in our docs but the name might be new to them. The more we can drill it the better maybe.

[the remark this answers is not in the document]
pleek91: Agreed that starting with "you" is something we should avoid.

docs/core-concepts/router.md
stackoverfloweth: lol fair enough. It's a good word that devs should know but I like simple

docs/core-concepts/router.md
pleek91: We're going to continue to work on this page. We both agree its needed but not great in its current state.

docs/core-concepts/navigation.md
pleek91: We decided to move the description of what a resolved route is to `/core-concepts/router#resolve` and link to that section.

stackoverfloweth (approved):

PR #397, opened 2024-12-29
New Feature: Router Plugins

pleek91:
# Description
Plugins add the ability to predefine router functionality to be used when creating a router. Plugins can add routes, rejections, and global hooks to a router. 

## Creating a plugin
Any object that satisfies the `RouterPlugin` type can be used. But the new `createRouterPlugin` utility can be used to easily create plugins in a type safe manor. 

```ts
import { createRouterPlugin, createRoute } from '@kitbag/router'

const plugin = creatRouterPlugin({
  routes: [
    createRoute(...),
    createRoute(...),
  ],
  rejections: {
    PluginRejection: { ... }
  },
  onBeforeRouteEnter: () => { ... }
})
```

## Installing plugins
Multiple plugins can be installed when creating a router using the 3rd argument of `createRouter`

```ts
const routes = [...] as const

const router = createRouter(routes, {}, [plugin1, plugin2]) 
```

## Type Safety
Any routes or rejections added by plugins will be available when navigating. Use `router.push`, `router.reject`, and components like `RouterLink` like normal.


pleek91:
@stackoverfloweth couple tests are failing around global hooks. I realized when writing tests for plugins that global hooks didn't work like I expected. I could update the tests to reflect the current functionality, but thought it was worth discussing first.

src/services/createRouter.ts
```diff
@@ -187,8 +202,8 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
     history.startListening()
   }
 
-  const resolve: RouterResolve<TRoutes> = (
-    source: RoutesName<TRoutes>,
+  const resolve: RouterResolve<TRoutes | TPlugin['routes']> = (
```
stackoverfloweth, line 205: I think I'd be tempted to define a type for this as like the first line of createRouter implementation

```ts
type PossibleRoutes = TRoutes | TPlugin['routes']
```

src/services/createRouterPlugin.ts
```diff
@@ -0,0 +1,42 @@
...
+  TRejections extends Record<string, Component> = {}
+>(plugin: Partial<RouterPlugin<TRoutes, TRejections>>): RouterPlugin<TRoutes, TRejections> {
+  return {
+    routes: plugin.routes ?? [] as unknown as TRoutes,
+    rejections: plugin.rejections ?? {} as TRejections,
+    ...plugin,
+  }
+}
+
+export function addRouterPluginHooks(router: Router, plugin: RouterPlugin) {
```
stackoverfloweth, line 18: looks like maybe linting is broken in your vscode, there should be a return type on this function

src/services/createRouter.ts
```diff
@@ -321,7 +348,10 @@ export function createRouter<const TRoutes extends Routes, const TOptions extend
     onAfterRouteLeave,
     prefetch: options?.prefetch,
     start,
+    stop,
   }
 
+  plugins.forEach(plugin => addRouterPluginHooks(router, plugin))
+
```
stackoverfloweth, line 355: for other plugin things like routes, rejections we just kind of spread the plugin with the user args when creating the thing. Could we could do that for hooks as well? Instead of this, maybe pass the plugin to [createRouterHooks](https://github.com/kitbagjs/router/blob/plugins/src/services/createRouterHooks.ts#L17)?

src/services/createRouterPlugin.ts
stackoverfloweth: I have a open question about why hooks are treated slightly differently in createRouter. If we do in fact need this, I don't think it belongs in this file.

src/services/createRouter.spec.ts
```diff
@@ -671,3 +671,42 @@ describe('router.push', () => {
...
+  router.onAfterRouteLeave(onAfterRouteLeave)
+
+  await router.start()
+
+  expect(onBeforeRouteEnter).toHaveBeenCalledTimes(1)
+  expect(onAfterRouteEnter).toHaveBeenCalledTimes(1)
+  
+  await router.push('parentA.childA', { paramA: 'valueA', paramB: 'valueB' })
+  
+  expect(onBeforeRouteEnter).toHaveBeenCalledTimes(2)
```
stackoverfloweth, line 699: digging into this with you, let me know if you have a breakthrough.

stackoverfloweth: also, I prefer to write this with the same expectations after each

```ts
  expect(onBeforeRouteEnter).toHaveBeenCalledTimes(1)
  expect(onBeforeRouteUpdate).toHaveBeenCalledTimes(0)
  expect(onBeforeRouteLeave).toHaveBeenCalledTimes(0)
  expect(onAfterRouteLeave).toHaveBeenCalledTimes(0)
  expect(onAfterRouteUpdate).toHaveBeenCalledTimes(0)
  expect(onAfterRouteEnter).toHaveBeenCalledTimes(1)

  await router.push('parentA.childA', { paramA: 'valueA', paramB: 'valueB' })

  expect(onBeforeRouteEnter).toHaveBeenCalledTimes(2)
  expect(onBeforeRouteUpdate).toHaveBeenCalledTimes(1)
  expect(onBeforeRouteLeave).toHaveBeenCalledTimes(0)
  expect(onAfterRouteLeave).toHaveBeenCalledTimes(0)
  expect(onAfterRouteUpdate).toHaveBeenCalledTimes(1)
  expect(onAfterRouteEnter).toHaveBeenCalledTimes(2)

  await router.push('parentA.childB', { paramA: 'valueB', paramD: 'valueD' })

  expect(onBeforeRouteEnter).toHaveBeenCalledTimes(3)
  expect(onBeforeRouteUpdate).toHaveBeenCalledTimes(2)
  expect(onBeforeRouteLeave).toHaveBeenCalledTimes(1)
  expect(onAfterRouteLeave).toHaveBeenCalledTimes(1)
  expect(onAfterRouteUpdate).toHaveBeenCalledTimes(2)
  expect(onAfterRouteEnter).toHaveBeenCalledTimes(3)
```

stackoverfloweth (commented):
I have been wanting this for so long and assumed it would be incredibly difficult. I'm blown away by how simple this appears to be, so kudos on that. SUPER excited about what this will unlock for us ❤️

src/services/createRouter.spec.ts
pleek91: Good suggestion. I updated both the router test and the plugins test to match this pattern and I updated them to match the current implementation. Updated in https://github.com/kitbagjs/router/pull/397/commits/2bed4556f9ba0b2be33cebca27f9c8851044a8cc

Issue for revisiting hooks https://github.com/kitbagjs/router/issues/402

src/services/createRouterPlugin.ts
pleek91: Hooks are always added to the hooks store. Global hooks are added by calling the routers hook methods like `router.onBeforeRouteEnter(() => {...})`. So somewhere when we're "installing" a plugin we need to add those hooks to the store. 

Where do you think this sort of logic belongs?

src/services/createRouter.ts
pleek91: Yeah tempting but I think this reads nicely still. If it gets more complex I'm down to alias it.

stackoverfloweth (approved):

src/services/createRouter.ts
pleek91: Discussed and we are going to follow this PR up with some refactoring around global hooks so that we can accept hooks as part of create router.

src/services/createRouterPlugin.ts
pleek91: We're going to follow up on this once we do some investigation into global hooks.

PR #405, opened 2025-01-02
Add global hooks to `RouterOptions` and simplify how hooks are managed

pleek91:
# Description
Adds hooks to `RouterOptions` to allow passing in global hooks when creating a router. 

Simplifies how hooks are managed and run by combining the concepts of the hooks store, hook runners, and hook registration methods into a single service. That way we don't have to create things and pass them around. 

This was motivated by https://github.com/kitbagjs/router/pull/397 where we didn't like how global hooks from plugins were registered. This also cleans that up by adding some utilities for getting all the global hooks from the router options and from plugins, and then passing them to a new `addRouteHooks` method that accepts an object of multiple types of hooks.


src/services/createRoute.ts
```diff
@@ -16,6 +16,7 @@ import { Query, ToQuery, toQuery } from '@/types/query'
 import { RouteMeta } from '@/types/register'
 import { Route } from '@/types/route'
 import { checkDuplicateParams } from '@/utilities/checkDuplicateKeys'
+import { WithHooks } from '@/types/hooks'
```
pleek91, line 19: I don't really like that this type is moved and used in two places (here and in `RouterOptions`). Maybe this should remain here and just be duplicated in `RouterOptions`?

src/services/createRouter.ts
```diff
@@ -72,17 +70,17 @@ export function createRouter<
...
   const TRoutes extends Routes,
   const TOptions extends RouterOptions,
   const TPlugin extends RouterPlugin = EmptyRouterPlugin
 >(routesOrArrayOfRoutes: TRoutes | TRoutes[], options?: TOptions, plugins: TPlugin[] = []): Router<TRoutes, TOptions, TPlugin> {
-  const pluginRoutes = plugins.map((plugin) => plugin.routes)
-  const routes = insertBaseRoute([...routesOrArrayOfRoutes, ...pluginRoutes].flat(), options?.base)
+  const routes = getRoutesForRouter(routesOrArrayOfRoutes, plugins, options?.base)
+  const hooks = createRouterHooks()
 
-  checkDuplicateNames(routes)
```
pleek91, line 85: This is unrelated. But part of the pattern of "pass the router and plugin arguments into a utility to combine them and return a normalized response".

src/services/createRouter.ts
```diff
@@ -72,17 +70,17 @@ export function createRouter<
...
+  hooks.addRouteHooks(getGlobalHooksForRouter(options, plugins))
```
pleek91, line 83: I know you wanted something besides the router to add the global hooks. But it didn't really feel good passing the global routes into `createRouterHooks` (even though that's what I started with). I think the router IS responsible for registering its global hooks. But we can abstract away the logic for collecting them all (in the `getGlobalHooksForRouter` utility)

src/services/createRouter.ts
```diff
@@ -124,7 +111,7 @@ export function createRouter<
     const to = find(url, options) ?? getRejectionRoute('NotFound')
     const from = { ...currentRoute }
 
-    const beforeResponse = await runBeforeRouteHooks({ to, from, hooks })
+    const beforeResponse = await hooks.runBeforeRouteHooks({ to, from })
```
pleek91, line 114: I really like that since the `createRouterHooks` service has the hooks store internally, we don't have to pass it around.

src/services/createRouterHooks.ts
pleek91, line 1: This file is quite a bit larger now. But now its self contained, so we don't have to do so much returning of internal state just to pass it into something else.

src/services/createRouterHooks.ts
```diff
@@ -1,53 +1,203 @@
...
+    hooks.onAfterRouteEnter.forEach((hook) => onAfterRouteEnter(hook))
+    hooks.onAfterRouteUpdate.forEach((hook) => onAfterRouteUpdate(hook))
+    hooks.onAfterRouteLeave.forEach((hook) => onAfterRouteLeave(hook))
   }
 
   return {
+    runBeforeRouteHooks,
+    runAfterRouteHooks,
+    addBeforeRouteHook,
+    addAfterRouteHook,
```
stackoverfloweth, line 194: do these need to be part of the API? Seems more like an internal utility to make the public methods DRYer

src/services/getGlobalHooksForRouter.ts
```diff
@@ -0,0 +1,19 @@
...
+    onBeforeRouteUpdate: getHooksForLifecycle('onBeforeRouteUpdate', options, plugins),
+    onBeforeRouteLeave: getHooksForLifecycle('onBeforeRouteLeave', options, plugins),
+    onAfterRouteEnter: getHooksForLifecycle('onAfterRouteEnter', options, plugins),
+    onAfterRouteUpdate: getHooksForLifecycle('onAfterRouteUpdate', options, plugins),
+    onAfterRouteLeave: getHooksForLifecycle('onAfterRouteLeave', options, plugins),
+  }
+}
+
+function getHooksForLifecycle<T extends BeforeRouteHookLifecycle | AfterRouteHookLifecycle>(lifecycle: T, options: RouterOptions, plugins: RouterPlugin[]): AsArray<Hooks[T]> {
+  return [options[lifecycle], ...plugins.map((plugin) => plugin[lifecycle])].flat().filter((hook) => hook !== undefined) as AsArray<Hooks[T]>
```
stackoverfloweth, line 18: eww dude. lol

```suggestion
  const optionsHooks = options[lifecycle] ?? []
  const pluginHooks = plugins.map((plugin) => plugin[lifecycle])
  const allHooks = [optionsHooks, ...pluginHooks]
    .flat()
    .filter((hook) => hook !== undefined)

  return allHooks as AsArray<Hooks[T]>
```

src/services/getRoutesForRouter.ts
stackoverfloweth, line 1: nice. like this encapsulation

src/services/getRoutesForRouter.ts
```diff
@@ -0,0 +1,20 @@
+import { RouterPlugin, Routes } from '@/types'
+import { insertBaseRoute } from './insertBaseRoute'
```
stackoverfloweth, line 2: more import nits

```suggestion
import { Routes } from '@/types/route'
import { RouterPlugin } from '@/types/routerPlugin'
import { insertBaseRoute } from '@/services/insertBaseRoute'
```

stackoverfloweth (approved):
Loved this PR ❤️

src/services/createRouterHooks.ts
pleek91: The component level hooks use these methods. We don't expose them on the router itself.

src/services/getGlobalHooksForRouter.ts
pleek91: I think yours is eww lol so many consts 😂 How about this
```suggestion
  return [
    options[lifecycle],
    ...plugins.map((plugin) => plugin[lifecycle]),
  ].flat().filter((hook) => hook !== undefined) as AsArray<Hooks[T]>
```

pleek91: Also a small private utility, I don't care to much about this especially since there's already an `as` 😂

pleek91: Or maybe this
```suggestion
  const hooks = [
    options[lifecycle],
    ...plugins.map((plugin) => plugin[lifecycle]),
  ].flat().filter((hook) => hook !== undefined)

  return hooks as AsArray<Hooks[T]>
```

Issue #402, opened 2024-12-31
Revisit global hooks and logic around when and what hooks run at startup


pleek91:
# Description
When working on the plugins api we noticed some oddities with hooks. Specifically
1. Global hooks are applied to all root routes (routes with no parents) rather than all routes.
2. When the router is started the "onBeforeRouteLeave" and "onAfterRouteLeave" hooks are executed. 

Let's discuss hooks as a whole and revisit some previous decisions and make sure hooks are working as they should and also satisfying intended use cases. 

Related to https://github.com/kitbagjs/router/pull/394

stackoverfloweth:
we should also consider a section like this on vue-router
<img width="765" alt="image" src="https://github.com/user-attachments/assets/8212a201-7a28-4b02-8543-7adb814d4fc4" />

PR #406, opened 2025-01-03
Update global hooks to not be based on root routes and fix `from` on router start

pleek91:
# Description
This PR fixes both hooks related issues outlined in https://github.com/kitbagjs/router/issues/402

**Global Hooks**
Now when global hooks are registered they are registered without conditional wrappers based on depth. That way when hooks are gathered they can be gathered based on what route navigation is actually happening (enter, update, or leave) no matter at what level. 

This means that if any of the matches are an enter then each global route enter hook will be executed once. Same for update and leave. 

**Update hooks on first navigation**
Since the `NotFound` rejection route is always the starting point for the route, when hooks were executed for the first navigation (when the router was started) the `from` context would have a route. This would cause global `leave` hooks to be executed. 

The router now checks the navigation id when getting the `from` route. And if it is the first navigation it passes `null` instead of the `NotFound` rejection route.

### DIFF
--- a/src/compositions/hooks.ts
+++ b/src/compositions/hooks.ts
@@ -2,7 +2,7 @@ import { inject, onUnmounted } from 'vue'
 import { useRouterDepth } from '@/compositions/useRouterDepth'
 import { RouterNotInstalledError } from '@/errors'
 import { routerHooksKey, RouterHooks } from '@/services/createRouterHooks'
-import { RegisterAfterRouteHook, RegisterBeforeRouteHook, AfterRouteHook, AfterRouteHookLifecycle, BeforeRouteHook, BeforeRouteHookLifecycle } from '@/types/hooks'
+import { AddAfterRouteHook, AddBeforeRouteHook, AfterRouteHook, AfterRouteHookLifecycle, BeforeRouteHook, BeforeRouteHookLifecycle } from '@/types/hooks'
 
 function useRouterHooks(): RouterHooks {
   const hooks = inject(routerHooksKey)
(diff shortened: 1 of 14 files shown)


src/compositions/hooks.ts
```diff
@@ -48,7 +48,7 @@ function afterComponentHookFactory(lifecycle: AfterRouteHookLifecycle) {
  * @returns {RouteHookRemove} A function that removes the added hook.
  * @group Hooks
  */
-export const onBeforeRouteLeave: AddBeforeRouteHook = beforeComponentHookFactory('onBeforeRouteUpdate')
+export const onBeforeRouteLeave: RegisterBeforeRouteHook = beforeComponentHookFactory('onBeforeRouteUpdate')
```
pleek91, line 51: There was a mix of `Register` and `Add` keywords used in hooks. Tried to unify the types around `Register`.

src/compositions/useRoute.browser.spec.ts
pleek91, line 1: I updated the `routes` in `testUtils` and updated these to reflect. Nothing functionally changed here.

src/compositions/hooks.ts
pleek91: Actually the types were register and the methods were add. So I renamed all the types to use "Add" to match the methods.

stackoverfloweth (approved):

src/types/hooks.ts
```diff
@@ -25,39 +17,37 @@ export type WithHooks = {
 
 export type BeforeHookContext = {
   to: ResolvedRoute,
-  from: ResolvedRoute,
+  from: ResolvedRoute | null,
```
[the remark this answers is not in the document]
pleek91: Correct, we even had `| null` in some existing types because that was always the intent. But it got lost along the way.

PR #407, opened 2025-01-05
Matched strongly typed

stackoverfloweth:
[not too long ago](https://github.com/kitbagjs/router/pull/325) we updated the types for `route.matched` and `route.matches` to preserve the name property. This PR takes that further and also preserves everything we can.

- path
- query
- host
- hash
- meta
- state
- component (+ props)


src/types/route.ts
```diff
@@ -41,12 +45,12 @@ export type Route<
   /**
    * The specific route properties that were matched in the current route.
   */
-  matched: CreateRouteOptionsMatched<TName>,
+  matched: TMatches extends [...any[], infer Last extends CreatedRouteOptions] ? Last : CreatedRouteOptions,
```
pleek91, line 48: Nit: Like to see a utility type for this rather than having a conditional type inlined here.

src/utilities/testHelpers.ts
```diff
@@ -30,6 +30,8 @@ const parentA = createRoute({
   path: '/parentA/[paramA]',
 })
 
+export type T1 = typeof parentA['path']
+
```
pleek91, line 34: I'm guessing this was just something you were testing with

pleek91 (commented):
This looks super clean. I wanna play around with it a bit still.

src/types/route.ts
stackoverfloweth: I added a `LastInArray` to @/types/utilities.ts, let me know what you think

pleek91: Yeah that looks nice, the fallback is for when we just use `Route` without any generics?

PR #411, opened 2025-01-08
simplify withState types

stackoverfloweth:
while working on https://github.com/kitbagjs/router/pull/407, we noticed that the types for `withState` were over-complicated. The types were built to behave like `withParent` and `withComponent`, but unlike those types we didn't actually benefit from the complexity because there are no overloads of `createRoute` that leverage `withState` vs `withoutState`.


pleek91 (approved):
❤️

PR #421, opened 2025-01-12
Move `props` option to second arg of `createRoute` utility

pleek91:
# Description
By moving the `props` option of `createRoute` to the second argument we can get much better typescript support and typescript errors. The utility will now correctly error when props are missing entirely if it detects the component has props that are required. And if the return type from the props function doesn't match the expected type for a component it gives much better errors.

```
const route = createRoute({
  ...
  component: echo,
}, () => ({
  value: 'hello world',
}))
```

Routes with named views will also expect an object by the same keys

```
const route = createRoute({
  ...
  components: {
    default: echo
  }
}, {
  default: () => ({ value: 'hello world' })
})
```


stackoverfloweth (commented):
looking great, I got it building (committed) and fixed all the tests that relied on props (also committed)

src/services/createRoute.ts
stackoverfloweth, line 1: so much simpler! 😍

stackoverfloweth (approved):
🚢 IT

PR #422, opened 2025-01-12
add nullable utility

stackoverfloweth:
currently, if users want to support `null` in their param types, they have to make their own with something like 

```ts
const numberParam = createParam(Number)
const nullishNumberParam = createParam({
  get: (value, extras) => {
    if (value === '') {
      return null
    }

    return numberParam.get(value, extras)
  },
  set: (value: number | null, extras) => {
    if (value === null) {
      return ''
    }

    return numberParam.set(value, extras)
  },
})
```

This approach requires that the param also be optional, since [getParamValue](https://github.com/kitbagjs/router/blob/main/src/services/params.ts#L109) sees an empty string as a `InvalidRouteParamValueError` if there is no default and the param is not optional. 

We could consider changing this logic inside services/params so that a custom `ParamGetSet` can choose what to do with any value but current param contract says the getter will get `string`.

```ts
export type ParamGetter<T = any> = (value: string, extras: ParamExtras) => T
export type ParamSetter<T = any> = (value: T, extras: ParamExtras) => string
```

However, I kind of thing an empty string SHOULD continue to be sent to the params getter. It's difficult to know if empty string was set by us because of an `undefined`. Furthermore, now that the supplied value could be `null`, it's even harder to know what value we should store and later return. My solution to this is the `nullKey`.

```ts
export const nullKey = 'NULL\u0000'
```

That value is supposed to be something a user is unlikely to supply by accident. It's used as the `defaultValue` for the param so when no value is supplied, this key gets substituted in and then replaced with `null`.  I wrote a test to ensure this value never gets actually put in the URL.


pleek91:
@stackoverfloweth does this approach mean that a user couldn't supply a default value for a nullable param?

PR #433, opened 2025-01-20
Add `parent` context with name and prop values to the callback context of route props

pleek91:
# Description
When parent routes have props, it can sometimes be useful to reuse one or more values in the props of the child route. This allows for sharing values across routes without duplicating the logic and also enables data dependent props where the child determines its props based the props value(s) from the parent. 

**sync props**
```ts
  const parent = createRoute({
    name: 'parent',
  }, () => ({ foo: 123 }))

  const child = createRoute({
    name: 'child',
    parent: parent,
    path: '/child',
  }, (__, { parent }) => ({
      propFromParent: parent.props.foo
  }))
```

**async props**
```ts
  const parent = createRoute({
    name: 'parent',
  }, async () => ({ foo: 123 }))

  const child = createRoute({
    name: 'child',
    parent: parent,
    path: '/child',
  }, async (__, { parent }) => {
      const { foo } = await parent.props

      return {
        propFromParent: parent.props.foo
      }
  }))
```
Note: Awaiting parent props creates waterfalls.

### DIFF
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -127,13 +127,13 @@ export function createPropStore(): PropStore {
         get props() {
           return getParentProps(parent, 'default', route, prefetch)
         },
-      } as PropsCallbackParent
+      }
     }
 
     if (isWithComponentPropsRecord(parent)) {
       return {
         name: parent.name ?? '',
-        props: new Proxy({} as any, {
+        props: new Proxy({}, {
           get(target, name) {
             if (typeof name !== 'string') {
               return Reflect.get(target, name)
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -116,7 +116,6 @@ export function createPropStore(): PropStore {
     return store.get(key)
   }
 
-  // todo: account for prefetching
   function getParentContext(parent: Route['matched'] | undefined, route: ResolvedRoute, prefetch: boolean = false): PropsCallbackParent {
     if (!parent) {
       return
@@ -146,7 +145,10 @@ export function createPropStore(): PropStore {
       }
     }
 
-    return
+    return {
+      name: parent.name ?? '',
+      props: undefined,
+    }
   }
 
   function getParentProps(parent: Route['matched'], name: string, route: ResolvedRoute, prefetch: boolean = false): unknown {
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -7,6 +7,7 @@ import { CallbackPushResponse, CallbackRejectResponse, CallbackSuccessResponse, 
 import { CallbackContextPushError } from '@/errors/callbackContextPushError'
 import { CallbackContextRejectionError } from '@/errors/callbackContextRejectionError'
 import { getPropsValue } from '@/utilities/props'
+import { PropsCallbackParent } from '@/types/props'
 
 export const propStoreKey: InjectionKey<PropStore> = Symbol()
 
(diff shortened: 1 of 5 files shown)


src/services/createPropStore.ts
```diff
@@ -111,6 +114,59 @@ export function createPropStore(): PropStore {
...
+      return
+    }
+
+    if (isWithComponentProps(parent)) {
+      return {
+        name: parent.name ?? '',
+        get props() {
+          return getParentProps(parent, 'default', route, prefetch)
+        },
+      } as PropsCallbackParent
```
pleek91, line 130: @stackoverfloweth I was struggling to get this file to be happy with the types. All the api types and functionality are tested. So I'm okay with some hacks in here. But be nice if they were not necessary. Maybe you can see a nicer solution.

src/services/createPropStore.ts
```diff
@@ -111,6 +114,59 @@ export function createPropStore(): PropStore {
...
+    if (!parent) {
+      return
+    }
+
+    if (isWithComponentProps(parent)) {
+      return {
+        name: parent.name ?? '',
+        get props() {
+          return getParentProps(parent, 'default', route, prefetch)
+        },
```
pleek91, line 129: I made this a get so that the prefetch warning wouldn't happen unless you actually tried to access the props. But I need to figure out how to test that.

src/services/createRoute.spec-d.ts
pleek91, line 1: These tests already existed I just moved them to this file.

src/services/createPropStore.ts
stackoverfloweth: actually I think we can pivot your `GetParentPropsReturnType` slightly and it all just works

src/services/createPropStore.ts
```diff
@@ -111,6 +114,59 @@ export function createPropStore(): PropStore {
...
+      } as PropsCallbackParent
+    }
+
+    if (isWithComponentPropsRecord(parent)) {
+      return {
+        name: parent.name ?? '',
+        props: new Proxy({} as any, {
```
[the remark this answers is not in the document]
stackoverfloweth: actually I don't think this is needed after my other suggestion

stackoverfloweth (commented):
Would be cool if we could extract some of the parent stuff to keep prop store as small as possible but I see some of the reasons you made that decision. 

I got a couple small questions and a commit that applies some type suggestions otherwise seems pretty straight forward

src/services/createPropStore.ts
```diff
@@ -111,6 +114,59 @@ export function createPropStore(): PropStore {
     return store.get(key)
   }
 
+  function getParentContext(route: ResolvedRoute, prefetch: boolean = false): PropsCallbackParent {
+    const parent = route.matches.at(-2)
```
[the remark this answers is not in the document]
pleek91: -1 is the current match. -2 is the previous match (the parent).

src/services/createPropStore.ts
```diff
@@ -111,6 +114,59 @@ export function createPropStore(): PropStore {
...
+      name: parent.name ?? '',
+      props: undefined,
+    }
+  }
+
+  function getParentProps(parent: Route['matched'], name: string, route: ResolvedRoute, prefetch: boolean = false): unknown {
+    const value = getProps(parent.id, name, route)
+
+    if (prefetch && !value) {
+      const parentName = parent.name ?? 'unknown'
```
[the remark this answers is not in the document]
pleek91: I want to use the same for both (either || or ??) but these are what eslint wanted. The types are different (`string | undefined` vs `string`)

src/services/createPropStore.ts
pleek91: I'm slightly worried that'll affect the types you get in createRoute. I'll have to write a type test to make sure it says props is undefined when there is a parent but there are no props.

pleek91: @stackoverfloweth I added a type test for this that is failing. If we change `GetParentPropsReturnType` to return `unknown` then when a parent has no props the child thinks they are `unknown` rather than `undefined.

stackoverfloweth (approved):

PR #434, opened 2025-01-22
add support for literals as param types

stackoverfloweth:
With this PR we're extending the possible param types to include literal values for `string | number | boolean`.

## Example

```ts
const example = createRoute({
  name: 'example',
  path: path('/[foo]' { foo: 14 })
})
```

This means when you supply a value for the `foo` param, it must be exactly the literal number `14`. The same works for literal strings and booleans.

## But Why

By itself, this feature doesn't provide much real world value to users.  If you had a literal value in mind for a param it's not exactly dynamic, which is the whole point of params in your url. However, we're collaborating on some exciting new utilities for `oneOf`, `anyOf`, and `allOf` which will expect an array of `Param`.


pleek91 (approved):
❤️

PR #439, opened 2025-01-24
[1st Attempt] Fix route components remounting rather than being updated when props change

pleek91:
# Description
When a route's props are updated the components should not be unmounted and recreated. This fixes an issue with how route components and props were combined that was causing new components to get mounted whenever the props changed. 

**Note**
In its current state this PR changes how the default slot of RouterView works. Which I'm hoping can be fixed so that users get the same functionality as before when using the default slot.

### DIFF
--- a/src/services/component.ts
+++ b/src/services/component.ts
@@ -1,6 +1,6 @@
+/* eslint-disable vue/require-prop-types */
 /* eslint-disable vue/one-component-per-file */
-import { AsyncComponentLoader, Component, FunctionalComponent, defineComponent, getCurrentInstance, h, ref } from 'vue'
-import { MaybePromise } from '@/types/utilities'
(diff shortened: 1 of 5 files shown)


src/services/component.ts
```diff
@@ -13,83 +13,75 @@ export type ComponentProps<TComponent extends Component> = TComponent extends Co
...
+    return ''
+  }
+}, {
+  props: ['component', 'props'],
+})
+
+const SuspenseAsyncComponentPropsWrapper = defineComponent(async (input: { component: Component, props: unknown }) => {
+  const values = ref()
+
+  values.value = await input.props
```
stackoverfloweth, line 72: So the suspense version is different because it uses top-level await for props, right?

And we only use this version when we know that the current instance has a Vue Suspense component above?

stackoverfloweth (approved):
Seems straight forward, glad it worked out with such a relatively small change.

PR #440, opened 2025-01-25
Fix route components remounting rather than being updated when props change

pleek91:
# Description
This approach builds on what I learned in https://github.com/kitbagjs/router/pull/439 and fixes the remounting of components when params change. What this does differently is it uses a factory to create a wrapper around the route's component that handles the prop binding, suspense, and async updates. And it stores that wrapper in a store so that it can be reused.

### DIFF
--- a/src/components/routerView.vue
+++ b/src/components/routerView.vue
@@ -1,31 +1,30 @@
 <template>
   <template v-if="component">
-    <slot name="default" v-bind="{ route, component, rejection }">
-      <component :is="component" />
+    <slot name="default" v-bind="{ route, component, props, rejection }">
+      <template v-if="rejection">
+        <component :is="rejection.component" />
+      </template>
+      <template v-else>
+        <component :is="component" :props="props" />
+      </template>
     </slot>
   </template>
 </template>
 
-<script lang="ts">
-/**
- * @ignore
- */
-</script>
-
 <script lang="ts" setup>
-  import { Component, UnwrapRef, VNode, computed, provide, resolveComponent } from 'vue'
+  import { Component, UnwrapRef, VNode, computed, provide } from 'vue'
   import { usePropStore } from '@/compositions/usePropStore'
   import { useRejection } from '@/compositions/useRejection'
   import { useRoute } from '@/compositions/useRoute'
   import { useRouterDepth } from '@/compositions/useRouterDepth'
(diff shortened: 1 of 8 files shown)


src/components/routerView.vue
```diff
@@ -1,94 +1,69 @@
...
-  import { component as componentUtil } from '@/services/component'
   import { RouterRejection } from '@/services/createRouterReject'
   import { RouterRoute } from '@/services/createRouterRoute'
-  import { CreateRouteOptions, isWithComponent, isWithComponents } from '@/types/createRouteOptions'
   import { depthInjectionKey } from '@/types/injectionDepth'
+  import { useComponentsStore } from '@/compositions/useComponentsStore'
 
   const { name = 'default' } = defineProps<{
     name?: string,
+    props?: unknown,
```
pleek91, line 27: Not sure about this. When a `RouterView` was rendered automatically it still ends up trying to pass props too it. But since this component doesn't have a prop called props vue would put a warning in the console "extraneous non props warning". Adding this hear silences that warning. Don't like it though. Will have to figure out a better way to silence the warning.

pleek91: Resolved by https://github.com/kitbagjs/router/pull/440/commits/8f2d7208d984f32c872863ddad2ad90f346a475f

stackoverfloweth:
@pleek91, just wanna make sure I fully understand this change. So RouterView is basically back to it's original form at least in the template. However, that `component` computed that was re-running when components changes is NOT being called anymore because the props are wrapped around component when it was added to the store?

1. If the user actually navigates away and Vue unmounts our RouterView, do we clear the components from the store? Otherwise when the user comes back, would the store re-use an old instance?
2. We have tests around props reactivity right? I would have been concerned that moving props out of the computed would break that reactivity.
3. Are we still considering the idea of `rejections` being out of users control?

pleek91:
@stackoverfloweth 
What we're creating and storing here is component definitions not component instances. So vue will create new component instances from these component definitions each time. But I'll update the new test to reflect that. 

And we could clear the definitions out but I think it makes more sense to just keep them indefinitely.

stackoverfloweth (approved):

pleek91:
@stackoverfloweth foe the rejections I think for now we should keep them as is otherwise transitions won't work going between routes and rejections.

PR #446, opened 2025-01-27
Experimental zod support for param types

pleek91:
# Description
Since [zod](https://github.com/colinhacks/zod) is a popular typescript schema validation library I wanted to see if we could support zod schemas as param types without any special public utilities or api.  I believe this PR adds support for most zod schemas to be used as param types out of the box. Now any `ZodSchema` is a valid param type and the correct param type will be inferred from the `ZodSchema` and used for the param itself. 

There are some schemas we cannot functionally support though. Most of which don't even make sense in a param context. 
Specifically:
- `ZodPromise`
- `ZodFunction`
- `ZodIntersection`
I believe we could support `ZodIntersection` but its a bit more complex then the others so I want to see if there's some demand for it first. 

These schemas are supported:
- `ZodString`
- `ZodBoolean`
- `ZodDate`
- `ZodNumber`
- `ZodLiteral`
- `ZodObject`
- `ZodEnum`
- `ZodNativeEnum`
- `ZodArray`
- `ZodTuple`
- `ZodUnion`
- `ZodDiscriminatedUnion`
- `ZodRecord`
- `ZodMap`
- `ZodSet`

This means you can do things like this
```ts
const route = createRoute({
  path: path('/[myParam], {
    myParam: z.string()
  }
})
```
And more complex params like
```ts
const route = createRoute({
  path: path('/[myParam], {
    myParam: z.object({
      foo: z.string(),
      bar: z.number(),
      baz: z.date()
  }
})
```

All support is experimental and subject to change or removal. And there are likely many bugs and edge cases (though I'm optimistic there are not many).


src/services/zod.ts
```diff
@@ -0,0 +1,259 @@
...
+        return stringifyZodValue(value, schema)
+      } catch {
+        throw invalid()
+      }
+    },
+  }
+}
+
+// regex for an iso date string
+const isoDateRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/
```
pleek91, line 45:
```suggestion
const isoDateRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/
```

stackoverfloweth (approved):

PR #447, opened 2025-01-28
Unified `withParams` utility to replace individual `path`, `query`, and `host` utilities

stackoverfloweth:
We ship separate utilities for assigning params for each of the different parts of the url. For the most part these all worked exactly the same (Hash was a little different, more on that later). This PR introduces a single, unified utility `withParams` to replace them all.

```ts
import { withParams } from '@kitbag/router'

const events = createRoute({
  name: 'events',
  path: withParams('/events/[year]/[?month]', {
    year: Number,
    month: withDefault(monthParam, 'january'),
  }),
  query: withParams('enabled=[?enabled]', {
    enabled: true,
  }),
})
```

I really didn't like the fact that `hash` didn't quite fit but still had extremely similar types. The biggest difference is that `hash` stored the value as `string | undefined`, but the others always store `string`. Hence our empty paths being defined as `Path<'', {}>`.  I actually prefer the idea of undefined so `withParams` stores value as possibly undefined. It has an overload that takes zero arguments and a separate property `toString: () => string` for instances where we need to normalize to a string.

Additionally, I decided to support params in the `hash` because it works exactly like path and I assumed it would be a small change. It's a bit bigger than I expected but still small enough that I feel like it was the right call.

Hash also had some logic for stripping a leading "#" when it's created but I just moved that so it would work like the other utilities.

To be kind to our current user base I've exported `path`, `query`, and `host` utilities but marked them as deprecated.


src/services/createExternalRoute.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    foo: 123,
+  })
+})
+
+test.each([
+  ['https://[foo].dev', '/[foo]', 'foo=[zoo]', '[bar]'],
+  ['https://[zoo].dev', '/[foo]', 'foo=[bar]', '[foo]'],
+  ['https://[zoo].dev', '/[bar]', 'foo=[foo]', '[foo]'],
+  ['https://[zoo].dev', '/[bar]', 'foo=[foo]', '[foo]'],
+])('given duplicate params across different parts of the route, throws DuplicateParamsError', (host, path, query, hash) => {
```
stackoverfloweth, line 66: this is the test I wanted to write, copied the rest from `createRoute.spec`, figured it doesn't hurt to have same tests for what we expect to behave the same way

src/services/createRoute.spec-d.ts
```diff
@@ -41,7 +40,7 @@ test('options with path', () => {
   type Source = typeof route['path']
   type Expect = { value: '/foo', params: {} }
 
-  expectTypeOf<Source>().toEqualTypeOf<Expect>()
+  expectTypeOf<Source>().toMatchTypeOf<Expect>()
```
stackoverfloweth, line 43: I switched these to `withParams` because the `WithParams` type also has the `toString: () => string` and I didn't want to include that in each of these `Expect` cases. It still tests what I think we're trying to assert

src/services/paramValidation.ts
```diff
@@ -18,22 +17,24 @@ export const routeParamsAreValid: RouteMatchRule = (route, url) => {
 }
 
 export const getRouteParamValues = (route: Route, url: string): Record<string, unknown> => {
-  const { pathname, search } = parseUrl(url)
+  const { protocol, host, pathname, search, hash } = parseUrl(url)
 
   return {
+    ...getPathParams(route.host, `${protocol}//${host}`),
```
stackoverfloweth, line 23: pretty sure this is a bug in production right now, `host` advertises support for params but doesn't validate them when matching

src/services/withParams.spec.ts
stackoverfloweth, line 1: I chose not to duplicate all the tests from `path`, `host`, `query`, etc since the specific formats aren't expected to make a difference

src/services/combineHash.spec.ts
pleek91, line 1: I thought hash had some special logic around NOT combining with the parent?

src/services/combineHash.ts
pleek91, line 1: I feel like the combinePath naming doesn't make sense now that its used for all the different pieces. Its also kinda interesting that since it accepts `WithParams` that there's no type safety preventing us from accidentally doing `WithParams<TParent['path'], TQuery>` 🤔

src/services/combinePath.ts
```diff
@@ -1,26 +1,31 @@
...
-    ? RemoveLeadingQuestionMarkFromKeys<TParentParams> & RemoveLeadingQuestionMarkFromKeys<TChildParams> extends PathParamsWithParamNameExtracted<`${TParentPath}${TChildPath}`>
-      ? Path<`${TParentPath}${TChildPath}`, RemoveLeadingQuestionMarkFromKeys<TParentParams> & RemoveLeadingQuestionMarkFromKeys<TChildParams>>
-      : Path<'', {}>
-    : Path<'', {}>
-  : Path<'', {}>
+  TParent extends WithParams,
+  TChild extends WithParams
+> = ToWithParams<TParent> extends { value: infer TParentPath extends string | undefined, params: infer TParentParams extends Record<string, unknown> }
+  ? ToWithParams<TChild> extends { value: infer TChildPath extends string | undefined, params: infer TChildParams extends Record<string, unknown> }
+    ? RemoveLeadingQuestionMarkFromKeys<TParentParams> & RemoveLeadingQuestionMarkFromKeys<TChildParams> extends ParamsWithParamNameExtracted<CombinePathString<TParentPath, TChildPath>>
```
pleek91, line 18: at some point it would be nice to change the way we're tracking params as optional. We're keeping the `?` in the param name but then we do a bunch of stuff to both "keep it" and "remove it" depending on the types we want. 

I think we should maybe update `WithParams` to return something like
```ts
{
  value: string,
  params: Record<string, param>,
  optional: string[]
}
```
Food for thought for future us

src/services/combinePath.ts
```diff
@@ -1,26 +1,31 @@
...
+      ? WithParams<CombinePathString<TParentPath, TChildPath>, RemoveLeadingQuestionMarkFromKeys<TParentParams> & RemoveLeadingQuestionMarkFromKeys<TChildParams>>
+      : WithParams<undefined, {}>
+    : WithParams<undefined, {}>
+  : WithParams<undefined, {}>
```
pleek91, line 22: Do we gain something from the `undefined` vs `{}` change?

src/services/createExternalRoute.ts
```diff
@@ -5,43 +5,40 @@ import { CombinePath } from '@/services/combinePath'
...
   const name = toName(options.name)
-  const path = toPath(options.path)
-  const query = toQuery(options.query)
-  const hash = toHash(options.hash)
+  const path = toWithParams(options.path)
+  const query = toWithParams(options.query)
+  const hash = toWithParams(options.hash)
   const meta = options.meta ?? {}
-  const host = isWithHost(options) ? toHost(options.host) : toHost('')
+  const host = isWithHost(options) ? toWithParams(options.host) : withParams()
```
pleek91, line 41: Why do we even need this ternary? Doesn't `toWithParams` accept undefined?

src/services/index.ts
```diff
@@ -1,10 +1,44 @@
...
 export { unionOf } from './unionOf'
 export { arrayOf } from './arrayOf'
 export { tupleOf } from './tupleOf'
+
+// backwards compatible withParams
+import { withParams, WithParams, toWithParams } from './withParams'
+/**
+ * @deprecated use `toWithParams` instead
+ */
+export const toPath = toWithParams
```
pleek91, line 16: These `toPath` and other "to" utilities aren't part of our public api are they? Just the `path`, 'query`, etc utilities. So I think we can just delete the "to" utilities yeah?

src/services/paramValidation.ts
pleek91: Should this be `getHostParams`?

src/services/routeMatchScore.ts
```diff
@@ -47,14 +47,14 @@ export function countExpectedPathParams(route: Route, actualPath: string): numbe
...
-  const missing = optionalParams.filter((expected) => getParamValueFromUrl(actualPath, route.path.value, expected) === undefined)
+  const missing = optionalParams.filter((expected) => getParamValueFromUrl(actualPath, route.path.toString(), expected) === undefined)
 
   return optionalParams.length - missing.length
 }
 
 export function countExpectedQueryParams(route: Route, actual: QuerySource): number {
   const actualQuery = new URLSearchParams(actual)
-  const expectedQuery = new URLSearchParams(route.query.value)
+  const expectedQuery = new URLSearchParams(route.query.toString())
```
pleek91, line 57: Why did we go back to having a `toString` method? Just for my own understanding.

src/services/urlAssembly.ts
```diff
@@ -23,16 +21,16 @@ export function assembleUrl(route: Route, options: AssembleUrlOptions = {}): Url
   const routeQuery = assembleQueryParamValues(route.query, paramValues)
   const searchParams = combineUrlSearchParams(routeQuery, queryValues)
   const pathname = assemblePathParamValues(route.path, paramValues)
-  const hash = createHash(route.hash.value ?? options.hash).value
+  const hash = route.hash.value ? assemblePathParamValues(route.hash, paramValues) : options.hash
```
pleek91, line 24: seeing utilities named `Path` being used for hash and host weirds me out. I'm guessing they're just functionally the same?

src/services/withParams.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function withParams<
+  const TValue extends string,
+  const TParams extends ParamsWithParamNameExtracted<TValue>
+>(value: TValue, params: TParams): WithParams<TValue, TParams>
+export function withParams(): WithParams<undefined, {}>
+export function withParams(value?: string, params?: Record<string, Param | undefined>): WithParams {
+  return {
+    value: stringHasValue(value) ? value : undefined,
+    params: value ? getParamsForString(value, params ?? {}) : {},
```
pleek91, line 60: seems like maybe `getParamsForString` should accept look like
```ts
function getParamsForString(value = '', params = {}) { ... }
```

src/services/paramValidation.ts
stackoverfloweth: if we reuse "path" functions, rename without the word "path".

stackoverfloweth: investigate if we actually need a "query" version as well

src/services/combineHash.spec.ts
stackoverfloweth: we currently DO combine hash currently in main

src/services/combineHash.ts
stackoverfloweth: I don't think we have that protection today, but it might be worth exploring. Something like assigning an internal `kind` (tagged type) to ensure wires don't get crossed

src/services/combinePath.ts
stackoverfloweth: I'll add this to my personal list and explore further, would love to get rid of leading "?" in param names

src/services/combinePath.ts
stackoverfloweth: if we just assume hash `''` (empty string) is the same as `undefined` we could remove all the `toStrings()`. See if you can remember why hash was special and expected `undefined` separately from empty string

src/services/createExternalRoute.ts
stackoverfloweth: good call, expect a couple slightly unrelated changes in this PR now

src/services/index.ts
stackoverfloweth: agreed. will remove

src/services/urlAssembly.ts
stackoverfloweth: part of larger discussion, will remove

src/services/withParams.ts
stackoverfloweth: agreed. fixed

pleek91 (approved):

PR #448, opened 2025-01-30
Run hooks and props within the context of the vue app

pleek91:
# Description
Since 3.3 vue has had a utility called [runWithContext](https://vuejs.org/api/application.html#app-runwithcontext) which allows for global injection to work within any callback. Using this when running hooks and props means that global injection also works within those callback. 

```ts
import { inject } from 'vue'

router.onAfterRouteEnter((to, context) => {
  const value = inject('global')

  ...
})
```

```ts
import { inject } from 'vue'

const route = createRoute({
  ...
}, async (route) => {
  const value = inject('global')

  return { value }
})
```

### DIFF
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -78,6 +78,8 @@ export function createRouter<
   const TOptions extends RouterOptions,
   const TPlugin extends RouterPlugin = EmptyRouterPlugin
 >(routesOrArrayOfRoutes: TRoutes | TRoutes[], options?: TOptions, plugins: TPlugin[] = []): Router<TRoutes, TOptions, TPlugin> {
+  let vueApp: App | null = null
+
   const routes = getRoutesForRouter(routesOrArrayOfRoutes, plugins, options?.base)
   const hooks = createRouterHooks()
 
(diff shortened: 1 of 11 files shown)


src/services/createRouter.ts
```diff
@@ -309,6 +311,10 @@ export function createRouter<
   }
 
   function install(app: App): void {
+    vueApp = app
+
+    propStore.setVueApp(app)
```
stackoverfloweth, line 316: I know this is overkill but thoughts on storing the vueApp in it's own "store"? That would solve 2 things I don't love about the current implementation. 

1. we use a `let`
2. the hooks and props technically have independent references and ideally they'd look in the same place

stackoverfloweth: that new "store" could also be responsible for exporting the `runWithContext` piece

pleek91: Don't hate it, I'll look into that tomorrow.

pleek91: @stackoverfloweth looking at this again not sure how the new store would work. We don't really have a way for the propsStore to get something from a separate store. Maybe we can chat about this

stackoverfloweth (approved):

src/services/createRouter.ts
pleek91: We might abstract this further in the future. But for now we're going to keep this simple.

PR #450, opened 2025-01-30
Make sure zod isn't required if zod params are not being used

pleek91:
# Description
https://github.com/kitbagjs/router/pull/446 introduced experimental support for zod params. However it didn't account for zod not being installed even though its clearly marked as an optional param. 

In order to account for this we can use an asyn import and only import actual zod code if we detect that a zod param is being used


stackoverfloweth (approved):
confirmed this works
- when target project doesn't use zod, zod isn't installed
- when target project has zod installed, uses zod params work as expected

PR #455, opened 2025-02-05
simplify extract param type, move check for optional

stackoverfloweth:
Currently there are some "extract param types" types were responsible for both finding out what type a param should be and also based on the key determining if the type should include `| undefined`. This PR moves that 2nd responsibility out into a separate type and also just aims to clean up the two similar but different extract types.

This needs to manual testing. I did a little testing and obviously the minimal type tests we have are passing but it would be easy to miss something that should have `| undefined` and doesn't or vice-versa.

### DIFF
--- a/src/types/params.ts
+++ b/src/types/params.ts
@@ -1,7 +1,7 @@
 import { LiteralParam, Param, ParamGetSet, ParamGetter } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { MakeOptional } from '@/utilities/makeOptional'
-import { ExtractParamTypeWithoutLosingOptional } from '@/types/routeWithParams'
+import { ZodSchema } from 'zod'
 
 export const paramStart = '['
 export type ParamStart = typeof paramStart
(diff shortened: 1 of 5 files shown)


src/types/params.ts
```diff
@@ -87,33 +87,66 @@ export type ExtractWithParamsParamType<
...
 export type ExtractRouteParamTypes<TRoute> = TRoute extends {
+  host: { params: infer HostParams extends Record<string, Param> },
   path: { params: infer PathParams extends Record<string, Param> },
   query: { params: infer QueryParams extends Record<string, Param> },
-  host: { params: infer HostParams extends Record<string, Param> },
   hash: { params: infer HashParams extends Record<string, Param> },
 }
   ? ExtractParamTypes<HostParams & PathParams & QueryParams & HashParams>
-  : {}
+  : Record<string, unknown>
```
pleek91, line 96: This change seems suspicious, we don't want to allow any string right?

pleek91 (commented):
Overall LGTM

src/types/routeWithParams.ts
```diff
@@ -1,48 +1,10 @@
...
-> = ExtractRouteParamTypesWithoutLosingOptional<RouteGetByKey<TRoutes, TKey>>
-
-type ExtractRouteParamTypesWithoutLosingOptional<TRoute> = TRoute extends {
-  host: { params: infer HostParams extends Record<string, Param> },
-  path: { params: infer PathParams extends Record<string, Param> },
-  query: { params: infer QueryParams extends Record<string, Param> },
-  hash: { params: infer HashParams extends Record<string, Param> },
-}
-  ? ExtractParamTypesWithoutLosingOptional<HostParams & PathParams & QueryParams & HashParams>
-  : Record<string, unknown>
```
stackoverfloweth, line 23: @pleek91 not saying it's correct but here's where that came from

pleek91: gotcha, so it just moved and the diff just makes it look like a change 👍

src/types/params.ts
```diff
@@ -1,7 +1,7 @@
 import { LiteralParam, Param, ParamGetSet, ParamGetter } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { MakeOptional } from '@/utilities/makeOptional'
-import { ExtractParamTypeWithoutLosingOptional } from '@/types/routeWithParams'
+import { ZodSchema } from 'zod'
```
pleek91, line 4:
```suggestion
import type { ZodSchema } from 'zod'
```

pleek91 (approved):
One change but LGTM

PR #464, opened 2025-02-08
alt approach to knowing what params are optional

stackoverfloweth:
## The Background

Currently, anything that can have params is normalized to `WithParams` type. This type has a `params` property that stores `Record<string, Param>`. The key for each param is where we store if we expect the param to be optional or not.

For example: 

```ts
{ value: '/[foo], params: { "foo": Boolean } }
```

This `WithParams` has a single param "foo", which we can see is **NOT** optional.

```ts
{ value: '/[?foo], params: { "?foo": Boolean } }
```

Versus this `WithParams`, where "foo" is now optional.

## The Motivation

We've never loved sticking the leading "?" in the params object for several reasons

- it's ugly
- it's changing the keys as defined by the user
- when we check for duplicate param names we need to look for the same name in twice where one has the question mark
- we combine two `withParams` objects we have to de-normalize param names back to exclude the question mark
- it creates a situation where there are 2 possibly conflicting sources of truth, the user already provides the optional declaration in the `value` property

That last reason is one that I found the most compelling and why I chose to try this syntax. This is also something that I didn't love about the suggestion to introduce a new `optional` property that only tracks what params are optional.

## The Solution

Now `WithParams` leaves params alone and looks at the `value` to determine what's optional.

```ts
{ value: '/[foo], params: { "foo": Boolean } } // required
{ value: '/[?foo], params: { "foo": Boolean } } // optional
```

### DIFF
--- a/src/services/withParams.ts
+++ b/src/services/withParams.ts
@@ -1,5 +1,5 @@
 import { getParamsForString } from '@/services/getParamsForString'
-import { ExtractParamName, ExtractWithParamsParamType, ParamEnd, ParamStart } from '@/types/params'
+import { ExtractParamName, ParamEnd, ParamStart } from '@/types/params'
 import { Param } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { isRecord } from '@/utilities/guards'
@@ -15,8 +15,15 @@ type WithParamsParamsOutput<
   TValue extends string,
   TParams extends Record<string, Param | undefined> = Record<never, never>
 > = TValue extends `${string}${ParamStart}${infer TParam}${ParamEnd}${infer Rest}`
-  ? Record<TParam, ExtractWithParamsParamType<TParam, TParams>> & WithParamsParamsOutput<Rest, TParams>
+  ? Record<ExtractParamName<TParam>, ParamTypeOrDefault<ExtractParamName<TParam>, TParams>> & WithParamsParamsOutput<Rest, TParams>
   : {}
+
+type ParamTypeOrDefault<
+  TKey extends string,
+  TParams extends Record<string, Param | undefined>
+> = TKey extends keyof TParams
+  ? TParams[TKey]
+  : StringConstructor
 
 export type WithParams<
   TValue extends string = string,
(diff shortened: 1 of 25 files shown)


pleek91 (approved):
Nice, I like this direction 👍

src/services/withParams.ts
```diff
@@ -15,9 +15,16 @@ type WithParamsParamsOutput<
...
-  ? Record<TParam, ExtractWithParamsParamType<TParam, TParams>> & WithParamsParamsOutput<Rest, TParams>
+  ? Record<ExtractParamName<TParam>, ParamTypeOrDefault<ExtractParamName<TParam>, TParams>> & WithParamsParamsOutput<Rest, TParams>
   : {}
 
+type ParamTypeOrDefault<
+  TKey extends string,
+  TParams extends Record<string, Param | undefined>
+> = TKey extends keyof TParams
+  ? TParams[TKey]
+  : StringConstructor
```
pleek91, line 26: What's the thought behind defaulting the params to the StringConstructor here? I believe we have other types that fallback to `string` for param types if a param is missing. But now it'll always be defined right?

src/types/params.ts
```diff
@@ -158,8 +122,3 @@ export type ExtractParamType<TParam extends Param> =
         : TParam extends LiteralParam
           ? TParam
           : string
-
-type RemoveLeadingQuestionMark<T extends PropertyKey> = T extends `?${infer TRest extends string}` ? TRest : T
-export type RemoveLeadingQuestionMarkFromKeys<T extends Record<string, unknown>> = {
-  [K in keyof T as RemoveLeadingQuestionMark<K>]: T[K]
-}
```
pleek91, line 165: Makes me happy to see this type deleted 🚀

pleek91 (approved):
I like this a lot better. Seems more maintainable and easier to grok too. Nice work

src/services/withParams.ts
stackoverfloweth: however, this IS accurate. Once `withParams` is called, any params not defined WILL be defined as `String`

pleek91: yep that makes sense, and https://github.com/kitbagjs/router/pull/464/commits/c153bd4890241d0f969af111fbdafcb579b47c3b simplifies this more how I'd expect. 

We probably have downstream types that are still expecting a param to possibly not be there that we can maybe simplify since we're normalizing `WithParams` to always include the param.

pleek91 (approved):

PR #469, opened 2025-02-10
Type bug assigning params with default

stackoverfloweth:
when we [simplified the extract type logic](https://github.com/kitbagjs/router/pull/455), we de-duped the ExtractType logic from src/types/routeWithParams.ts so now there are 2 types in src/types/params.ts. The types are `ExtractParamTypes` and `ExtractParamTypesWithOptional`. When testing some params I had the following test case

```ts
const keys = createRoute({
  parent: settings,
  name: 'settings.keys',
  path: '/keys',
  query: withParams('sort=[?sort]', { 
    sort: withDefault(unionOf('asc', 'desc'), 'asc') 
  }),
  component: defineAsyncComponent(() => import('../views/SettingsKeysView.vue'))
})
```

and I found that when I tried pushing to `settings.keys` it was expecting `sort` to be passed because our logic in `ExtractParamTypesWithOptional` says if an optional param has a default, it should REMOVE the `| undefined`. This is true when **reading** params but opposite when **writing**. 

While verbose, this PR further bifurcates the `ExtractParamTypesWithOptional` into 
- ExtractParamTypesOptionalReading
- ExtractParamTypesOptionalWriting

So now there are 3 "extract params" types (with comments in code to explain)
- **ExtractParamTypes**: Doesn't consider optionality
- **ExtractParamTypesOptionalReading**: Differs from ExtractParamTypes in that it also reads the string `value` from WithParams to determine what should be optional.
- **ExtractParamTypesOptionalWriting**: Differs from ExtractParamTypesOptionalReading in that optional params with defaults will remain optional.

### DIFF
--- a/src/services/withParams.ts
+++ b/src/services/withParams.ts
@@ -4,20 +4,22 @@ import { Param } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { isRecord } from '@/utilities/guards'
 
+type AssignableParamsFromString<
+  TValue extends string
+> = TValue extends `${string}${ParamStart}${infer TParam}${ParamEnd}${infer Rest}`
+  ? Record<ExtractParamName<TParam>, Param | undefined> & AssignableParamsFromString<Rest>
+  : {}
+
 type ExtractParamsFromString<
   TValue extends string,
   TParams extends Record<string, Param | undefined> = Record<never, never>
-> = TValue extends `${string}${ParamStart}${infer Param}${ParamEnd}${infer Rest}`
-  ? Record<Param, ExtractWithParamsParamType<Param, TParams>> & ExtractParamsFromString<Rest, TParams>
-  : Record<never, never>
-
-export type ParamsWithParamNameExtracted<TValue extends string> = {
-  [K in keyof ExtractParamsFromString<TValue> as ExtractParamName<K>]?: Param
-}
+> = TValue extends `${string}${ParamStart}${infer TParam}${ParamEnd}${infer Rest}`
+  ? Record<ExtractParamName<TParam>, ExtractWithParamsParamType<TParam, TParams>> & ExtractParamsFromString<Rest, TParams>
+  : {}
 
 export type WithParams<
   TValue extends string = string,
-  TParams extends ParamsWithParamNameExtracted<TValue> = Record<string, Param | undefined>
+  TParams extends Record<string, Param | undefined> = Record<string, Param | undefined>
 > = {
   value: TValue,
   params: string extends TValue ? Record<string, Param> : Identity<ExtractParamsFromString<TValue, TParams>>,
@@ -50,7 +52,7 @@ export function toWithParams<T extends string | WithParams | undefined>(value: T
 
 export function withParams<
   const TValue extends string,
-  const TParams extends ParamsWithParamNameExtracted<TValue>
+  const TParams extends AssignableParamsFromString<TValue>
 >(value: TValue, params: TParams): WithParams<TValue, TParams>
 export function withParams(): WithParams<'', {}>
 export function withParams(value?: string, params?: Record<string, Param | undefined>): WithParams {
--- a/src/services/withParams.ts
+++ b/src/services/withParams.ts
@@ -3,6 +3,7 @@ import { ExtractParamName, ExtractWithParamsParamType, ParamEnd, ParamStart } fr
 import { Param } from '@/types/paramTypes'
 import { Identity } from '@/types/utilities'
 import { isRecord } from '@/utilities/guards'
+import { MakeOptional } from '@/utilities/makeOptional'
 
 type AssignableParamsFromString<
   TValue extends string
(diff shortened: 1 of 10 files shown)


src/types/params.ts
```diff
@@ -96,18 +96,33 @@ export type ExtractRouteParamTypes<TRoute> = TRoute extends {
...
+ * @returns A record of parameter names to their respective types, extracted and merged from both path and query parameters.
+ */
+export type ExtractRouteParamTypesOptionalReading<TRoute> = TRoute extends {
+  host: { params: infer HostParams extends Record<string, Param> },
+  path: { params: infer PathParams extends Record<string, Param> },
+  query: { params: infer QueryParams extends Record<string, Param> },
+  hash: { params: infer HashParams extends Record<string, Param> },
+}
+  ? Identity<MakeOptional<ExtractParamTypesOptionalReading<HostParams> & ExtractParamTypesOptionalReading<PathParams> & ExtractParamTypesOptionalReading<QueryParams> & ExtractParamTypesOptionalReading<HashParams>>>
+  : Record<string, unknown>
```
pleek91, line 111: Do we even need to infer things here or can we just use `TRoute extends Route` and then `TRoute['path']['params]`

src/types/params.ts
```diff
@@ -149,6 +177,6 @@ export type ExtractParamType<TParam extends Param> =
           : string
 
 type RemoveLeadingQuestionMark<T extends PropertyKey> = T extends `?${infer TRest extends string}` ? TRest : T
-export type RemoveLeadingQuestionMarkFromKeys<T extends Record<string, unknown>> = {
+export type RemoveLeadingQuestionMarkFromKeys<T> = {
```
pleek91, line 180: Why remove the extends? This type won't work all `unknown` types which is essentially what this is now right?

src/types/params.ts
```diff
@@ -96,18 +96,33 @@ export type ExtractRouteParamTypes<TRoute> = TRoute extends {
...
  */
-export type ExtractRouteParamTypesWithOptional<TRoute> = TRoute extends {
+export type ExtractRouteParamTypesOptionalWriting<TRoute> = TRoute extends {
   host: { params: infer HostParams extends Record<string, Param> },
   path: { params: infer PathParams extends Record<string, Param> },
   query: { params: infer QueryParams extends Record<string, Param> },
   hash: { params: infer HashParams extends Record<string, Param> },
 }
-  ? ExtractParamTypesWithOptional<HostParams & PathParams & QueryParams & HashParams>
+  ? Identity<MakeOptional<ExtractParamTypesOptionalWriting<HostParams> & ExtractParamTypesOptionalWriting<PathParams> & ExtractParamTypesOptionalWriting<QueryParams> & ExtractParamTypesOptionalWriting<HashParams>>>
```
pleek91, line 125: Why do we need to call the extract type for each type of param now rather than passing all the params in as a union like we were before?

src/services/withParams.ts
```diff
@@ -3,21 +3,24 @@ import { ExtractParamName, ExtractWithParamsParamType, ParamEnd, ParamStart } fr
...
-> = TValue extends `${string}${ParamStart}${infer Param}${ParamEnd}${infer Rest}`
-  ? Record<Param, ExtractWithParamsParamType<Param, TParams>> & ExtractParamsFromString<Rest, TParams>
-  : Record<never, never>
-
-export type ParamsWithParamNameExtracted<TValue extends string> = {
-  [K in keyof ExtractParamsFromString<TValue> as ExtractParamName<K>]?: Param
-}
+> = TValue extends `${string}${ParamStart}${infer TParam}${ParamEnd}${infer Rest}`
+  ? Record<TParam, ExtractWithParamsParamType<TParam, TParams>> & ExtractParamsFromString<Rest, TParams>
+  : {}
```
pleek91, line 19: I don't understand why we need both of these. Seems like we only need `AssignableParamsFromString`

src/types/params.ts
stackoverfloweth: this was part of testing, I'll put it back but also this type will be removed [here](https://github.com/kitbagjs/router/pull/464) FYI

src/types/params.ts
stackoverfloweth: this was required at one point but I can't remember why and changing it doesn't _appear_ to break anything.

pleek91: I don't feel strongly it needs to change. Just seemed odd to not have an extends and to need all the infers.

src/types/params.ts
stackoverfloweth: in an upcoming PR I'm adjusting the `ExtractParam` types to also take the `string` part to determine what's optional as part of [this change](https://github.com/kitbagjs/router/pull/464). When that happens, this is how I want to call them so I left that in for now.

src/services/withParams.ts
stackoverfloweth: I'll adjust the names
- `AssignableParamsFromString` is the "input" for `withParams(..., THIS)`
- `ExtractParamsFromString ` is the "ouput" for `withParams(....) => { params: THIS }`

pleek91 (approved):

PR #475, opened 2025-02-16
moved storage and run with context into "vueAppStore" service

stackoverfloweth:
zero impact on users or functionality, just exploring an idea I had to encapsulate some [recent changes](https://github.com/kitbagjs/router/pull/448) into a service.


pleek91 (commented):
I don't think a singleton is a good idea. It works but limits users to having a single router in a single vue app. If a user has multiple vue apps each with a router the contexts will get mixed up and end up using a different vue app than the one the router was installed to. 

So far we've kept a router instance completely self contained (other than the register interface) and I think it would be a good idea to keep it that way. But let's discuss the pros and cons here. 

HMR is another thing to consider. In my experience singletons don't play very well with hmr

PR #436, opened 2025-01-23
`unionOf`, `arrayOf`, and `tupleOf` utilities

stackoverfloweth:
This PR introduces 3 new utilities for easily defining complex params

```ts
import { unionOf, arrayOf, tupleOf } from '@kitbag/router'

const category = unionOf('music', 'sports', 'art')
//         ^? ParamGetSet<"music" | "sports" | "art">

const categories = arrayOf('music', 'sports', 'art')
//         ^? ParamGetSet<("music" | "sports" | "art")[]>

const location = tupleOf(Number, Number)
//         ^? ParamGetSet<[Number, Number]>
```

PR #480, opened 2025-02-22
update param utilities to accept options, separator for arrays

stackoverfloweth:
in a [recent PR](https://github.com/kitbagjs/router/pull/436) we introduced 3 new param utilities

`arrayOf`, `tupleOf`, and `unionOf`

This PR modifies the call signature of these utilities to accept an array of param types as the 1st argument instead of spreading the argument. This allows for a 2nd argument to be specified for "options".

As of right now, array and tuple are the only ones that accept options, and their options are identical.

```ts
type options = {
  separator?: string,
}
```

This allows users to override the default behavior of the utilities to choose their own separator instead of the default comma `","`.  For example, users might choose a pipe `"|"` which changes the url from

> https://ticketing.service.dev?category=music,sports,art

to be

> https://ticketing.service.dev?category=music|sports|art

Thanks to @nclemeur for making the [suggestion](https://discord.com/channels/1240867447033172138/1240868059196030977/1341662846626435073) in our Discord

### DIFF
--- a/src/services/arrayOf.spec.ts
+++ b/src/services/arrayOf.spec.ts
@@ -8,7 +8,7 @@ test.each([
   ['true,23,foo', [true, 23, 'foo']],
   ['1,2,3,4,5', [1, 2, 3, 4, 5]],
 ])('given an array of params with valid values, returns an array of values', (input, expected) => {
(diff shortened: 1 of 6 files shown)


pleek91 (approved):
Love it

Issue #516, opened 2025-04-19
`RouterLink` `isExactMatch` is true on routes with different params


arpadgabor:
It seems like the `isExactMatch` provided by `RouterLink` is true for all of the following routes in my app:

![Image](https://github.com/user-attachments/assets/7f6a169f-9093-4faa-80c6-0e3c54dcb10f)

The code looks like this (removed some styles/text for brevity), I also included the workaround code (comparing the `href` from the router-link `route` and the route returned by `useRoute`). 

```vue
<script>
import { RouterLink, useRoute } from '@kitbag/router';

const currentRoute = useRoute()
</script>

...
<template>
  <RouterLink
    v-for="channel of channels"
    :key="channel.id"
    :to="(resolve) => resolve('JoinedRoom', { channelId: channel.id })"
    v-slot="{ isExactMatch, route }"
  >
    <div
      :class="[currentRoute.href === route?.href ? 'bg-gray-700' : '']"
    >
      {{ isExactMatch ? 'matched' : '' }}
    </div>
  </RouterLink>
</template>
```

pleek91:
Can you provide the three routes? Just the code where you call "createRoute"

pleek91:
Or wait it's the same route with different params right?

arpadgabor:
But this is the code in any case:
```ts
const room = createRoute(
	{
		name: 'JoinedRoom',
		parent: rooms,
		path: withParams('/[channelId]', {
			channelId: String
		}),
		component: Room
	},
	(route) => ({ channelId: route.params.channelId })
);
```

pleek91:
I'll double check but I believe this is working correctly. The match and exact match properties are about matching route definitions not URLs. 

If the current url resolves "Room" then any links on the page that also resolve to "Room" will be an exact match. And any URLs to "RoomLayout" (its parent) will be a match (but not an exact match).

pleek91:
Verified that the `isMatch` and `isExactMatch` are factoring in the route but not the url itself. I'm guessing you're looking for functionality to determine which link is "active"? We could add both `isActive` and `isExactActive` easily if that's what you're looking for.

arpadgabor:
Yeah `isActive`/`isExactActive` would help. Though I find it a bit confusing for the `isMatch`. When would I want to use `isMatch`/`isExactMatch` in a `RouterLink`?

pleek91:
@arpadgabor update to latest and you'll have new `isActive` and `isExactActive` properties to use. 

[...]

PR #518, opened 2025-04-20
Add classes and slot scope for "active" links

pleek91:
# Description
The `isMatch` and `isExactMatch` properties of `useLink` and `router-link` give users insight into whether or not the link's route matches the current route based on the actual route definitions. This means for the route `/posts/[postId]` both `/posts/foo` and `/posts/bar` are exact matches if the current route is `/posts/baz`. and `/posts` is a match for all three. 

These aren't very useful when trying to determine if a link is the same as the current routes href (the url the user sees). 

Introducing `isActive` and `isExactActive` to address this. These properties focus on the hrefs/urls themselves. So in the posts example, if the current url is `/posts/baz`
| Link | Active | Exact Active |
|----------|----------|----------|
| `/posts/foo`   | false    | false     |
| `/posts/bar`   | false    | false     |
| `/posts/baz`  | true      | true      |
| `/posts`         | true      | false     |

Closes https://github.com/kitbagjs/router/issues/516

### DIFF
--- a/src/compositions/useLink.ts
+++ b/src/compositions/useLink.ts
@@ -11,6 +11,7 @@ import { Url, isUrl } from '@/types/url'
 import { AllPropertiesAreOptional } from '@/types/utilities'
 import { isRoute } from '@/guards/routes'
 import { combineUrlSearchParams } from '@/utilities/urlSearchParams'
+import { isDefined } from '@/utilities/guards'
 
 export type UseLink = {
   /**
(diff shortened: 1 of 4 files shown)


src/compositions/useLink.ts
```diff
@@ -110,6 +119,8 @@ export function useLink(
 
   const isMatch = computed(() => isRoute(router.route) && router.route.matches.some((match) => match.id === route.value?.id))
   const isExactMatch = computed(() => router.route.id === route.value?.id)
+  const isActive = computed(() => isRoute(router.route) && isDefined(route.value) && router.route.href.startsWith(route.value.href))
+  const isExactActive = computed(() => router.route.href === route.value?.href)
```
stackoverfloweth, line 123: I think I'd be tempted to use `isMatch` and `isExactMatch` inside of `isActive`. I think of these as MORE narrow guards on top of those and without using them inside of the function it could lead to surprises like a route that uses something like `/[userId]` being discovered as `isActive` for the actual route of `/home`. That's probably not the best example, but I think you get the point

pleek91: Yeah it's technically possible to have a url that matches multiple routes. But the first route will always match based on the order they're defined.

I also don't think the route matters for whether the route is active or not. And it be weird I think if a link exactly matches the current url but wasn't marked as "exact active".

stackoverfloweth (approved):

PR #523, opened 2025-04-29
Make sure router view waits for the router to be started when doing SSR

pleek91:
# Description
Because the router's initialization is async the router view won't render anything when the vue app first starts. In we rely on reactivity to automatically render the route component when the router is finished initializing. But when doing SSR we need to make sure the router is initialized BEFORE we attempt to render anything. Leveraging `onServerPrefetch` let's vue know to wait to render the router view component until the router is finished initializing.

### DIFF
--- a/src/components/routerView.spec.ts
+++ b/src/components/routerView.spec.ts
@@ -1,0 +1,29 @@
+import { createSSRApp } from 'vue'
+import { describe, it, expect } from 'vitest'
+import { createRoute } from '@/services/createRoute'
+import { createRouter } from '@/services/createRouter'
+import { renderToString } from 'vue/server-renderer'
+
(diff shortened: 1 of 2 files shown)


src/components/routerView.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const app = createSSRApp({
+      template: '<RouterView/>',
+    })
+
+    app.use(router)
+
+    const html = await renderToString(app)
+
+    expect(html).toMatchInlineSnapshot('"<!--[-->hello world<!--]-->"')
```
pleek91, line 27: The snapshot has the comments because of the `v-if` in router view. This looks weird but its what's rendered.

PR #553, opened 2025-07-24
Bug Fix: router resolve panics when missing optional param

### LINKED ISSUE #552: Optional Parameters Require a Value
Let's say I have a route like:

```js
createRoute({name: 'item', path: withParams('/item/[id]/[?name]', {id: Number, name: withDefault(String, null)}), component: ItemView})
```

If I create a link like:

```js
<RouterLink :to="(resolve) => resolve('item', {id: 123})">View</RouterLink>
```

I get console errors. Since name is optional, I would expect it to function properly. If I instead do include the name property when generating the link, but either set it to null or an empty string, I also get an error.

If I create the same route using query parameters, I don't get an error and things work as expected.

```js
createRoute({name: 'item', path: '', query: withParams('item=[id]&name=[?name]', {id: Number, name: withDefault(String, null)}), component: ItemView})
```

Looking at the example here: https://router.kitbag.dev/core-concepts/params.html#optional-params, I'm unable to match the route if the month isn't also included.

Potentially related to https://github.com/kitbagjs/router/issues/414

stackoverfloweth:
When the resolve is called on a route that has multiple params where **at least one is optional and not supplied by URL**, it will throw an exception. 

Example route

```ts
const route = createRoute({
  name: 'item', 
  path: withParams('/item/[id]/[?name]', {
    id: Number, 
    name: withDefault(String, null)
  }), 
  component: ...
})
```

Example RouterLink (uses `useLink`, which calls `router.resolve`)

```html
<router-link :to="(resolve) => resolve('item', { id: 1 })">Item</router-link>
```

The path the router takes to resolve is

1. router is supplied with route name, params `name="item", params={id: 1}`
2. router finds the best match
3. router assembles the URL `"/item/1/"`
4. router parses that URL to find and parse any params

During step 4, the router iterates through the params and uses regex to extract the individual values. When the router was searching for the required `id` param first it adds a regex capture group to get the value for `id`.

`/item/[id]/[?name]` -> `/item/(.*)/[?name]`

then it's supposed to replace any other params with a regex "catch all"

`/item/(.*)/[?name]` -> `/item/(.*)/.+`

wherein lies the issue. You'll notice that "catch all" is `.+` which means **one** to unlimited characters. So [the fix](https://github.com/kitbagjs/router/commit/369d98eac42c3c97e6881844bd657886577f6f20) was to update our `replaceParamSyntaxWithCatchAlls` utility to use `.*` for a "catch all" instead (this is **zero** to unlimited characters).

TLDR: the regex for finding a param, assumes other params would be have at least 1 character

closes #552

### DIFF
--- a/src/services/routeRegex.spec.ts
+++ b/src/services/routeRegex.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'vitest'
 import { createRoute } from '@/services/createRoute'
-import { generateRoutePathRegexPattern, generateRouteQueryRegexPatterns, getParamName, splitByMatches } from '@/services/routeRegex'
+import { generateRoutePathRegexPattern, generateRouteQueryRegexPatterns, getParamName, regexCaptureAll, regexCatchAll, splitByMatches } from '@/services/routeRegex'
 import { component } from '@/utilities/testHelpers'
 
 describe('generateRoutePathRegexPattern', () => {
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
Nice, good find

PR #558, opened 2025-08-20
Add support for automatic reloading of the window if a vite:prefetchError event is detected

pleek91:
# Description
Attempting to add automatic detection of vite preload errors that will automatically refresh the app. Based off of [this article](https://paulau.dev/blog/handle-version-skew-after-new-deployment-with-vite-and-vue-router/) by @romansp 

This is untested and I need to write some tests as well as figure out how to test if this actually works in the real world.

### DIFF
--- a/src/services/createRouterPreloadErrorRefresh.ts
+++ b/src/services/createRouterPreloadErrorRefresh.ts
@@ -1,0 +1,65 @@
+import { isBrowser } from '@/utilities/isBrowser'
+import { onUnmounted } from 'vue'
+
+type RouterPreloadErrorOptions = {
+  refreshOnPreloadError?: boolean,
+}
+
+type RouterPreloadError = {
+  initialize: () => void,
+  onBeforePreloadErrorRefresh: (hook: RouterPreloadErrorHook) => void,
+}
+
+type RouterPreloadErrorHookParameters = {
+  event: Event,
+  abort: () => void,
+}
+
+export type RouterPreloadErrorHook = (parameters: RouterPreloadErrorHookParameters) => void
+
+export function createRouterPreloadErrorRefresh({ refreshOnPreloadError = true }: RouterPreloadErrorOptions): RouterPreloadError {
+  const hooks = new Set<RouterPreloadErrorHook>()
+
+  function initialize(): void {
+    if (isBrowser() && refreshOnPreloadError) {
+      window.addEventListener('vite:preloadError', (event) => {
+        try {
+          hooks.forEach((hook) => {
+            hook({ event, abort })
+          })
+
+          window.location.reload()
+        } catch (error) {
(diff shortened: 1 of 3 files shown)


src/services/createRouterPreloadErrorRefresh.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export type RouterPreloadErrorHook = (parameters: RouterPreloadErrorHookParameters) => void
+
+export function createRouterPreloadErrorRefresh({ refreshOnPreloadError = true }: RouterPreloadErrorOptions): RouterPreloadError {
+  const hooks = new Set<RouterPreloadErrorHook>()
+
+  function initialize(): void {
+    if (isBrowser() && refreshOnPreloadError) {
+      window.addEventListener('vite:preloadError', (event) => {
```
stackoverfloweth: nit: I'd invert this personally

src/services/createRouterPreloadErrorRefresh.ts
```diff
@@ -0,0 +1,65 @@
...
+    initialize,
+    onBeforePreloadErrorRefresh,
+  }
+}
+
+class RouterPreloadErrorAbort extends Error {
+  public constructor() {
+    super('Router preload error aborted')
+  }
+}
```
stackoverfloweth: is this a meaningful departure from re-using `createCallbackContext`? I kinda wonder if we could somehow plug this into the existing infrastructure like other before hooks

stackoverfloweth (commented):
Mostly just interested in trying to establish some "built in hooks". Did you try that? Would you want me to give it a shot?

src/services/createRouterPreloadErrorRefresh.ts
pleek91: It's meant to have a similar feel. But hooks are part of the regular lifecycle of a Route. This happens completely outside of the router itself basically. We're just using the router as an entry point to add the event listener. 

This "hook" only exists because I assume there's a use case where the error is expected and a dev will not want to refresh the page. 

But I'd rather rip this out before trying to tie it in with hooks. Hooks are a nicely scoped feature and I don't want to make them more complex trying to make every "onXyz" method part of the hooks service.

src/services/createRouterPreloadErrorRefresh.ts
pleek91: Nit: why

stackoverfloweth: sorry I thought it was obvious, to avoid nesting

pleek91:
@stackoverfloweth I'm going to close this PR for now. As is this doesn't do anything that actually needs the router. Which means it doesn't belong here and should remain in userland. I think we should keep this in mind and re-imagine how we can leverage the vite preload error to enhance the router itself and build a feature that sense.

PR #559, opened 2025-08-31
code review suggestions

stackoverfloweth:
Happy to merge the underlying PR either way. This is the branch I tested

### DIFF
--- a/src/errors/routerPreloadErrorAbortError.ts
+++ b/src/errors/routerPreloadErrorAbortError.ts
@@ -1,0 +1,5 @@
+export class RouterPreloadErrorAbort extends Error {
+  public constructor() {
+    super('Router preload error aborted')
+  }
+}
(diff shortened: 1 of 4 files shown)


src/errors/routerPreloadErrorAbortError.ts
```diff
@@ -0,0 +1,5 @@
+export class RouterPreloadErrorAbort extends Error {
```
pleek91: If there's no hook we don't need the error right? Nothing throws it.

src/services/createRouterPreloadErrorRefresh.ts
```diff
@@ -1,65 +1,35 @@
...
-  function onBeforePreloadErrorRefresh(hook: RouterPreloadErrorHook): void {
-    hooks.add(hook)
+    window.addEventListener('vite:preloadError', () => {
+      try {
+        console.log('vite:preloadError REFRESHING!')
+        window.location.reload()
+      } catch (error) {
+        if (error instanceof RouterPreloadErrorAbort) {
+          return
+        }
```
pleek91: We dont need to catch this error right

src/errors/routerPreloadErrorAbortError.ts
stackoverfloweth: right. Fixed

PR #563, opened 2025-09-05
Add support for non global routers

pleek91:
# Description
Having a global router with type safety provided by the `Register` interface is great for most use cases. But more complex use cases like monorepos with multiple apps, nested routers, etc require routers that don't depend on the `Register` interface. 

This refactor allows for these use cases by allowing users to pass `isGlobalRouter: false` as an option when creating a router. This will tell the router to not use the default injection key and it will create a unique injection key specifically for this router instance. 

```ts
   isGlobalRouter: false
})
```

When `isGlobalRouter` is `false` none of the exported assets (components, compositions, hooks, etc) will work with this router instance. 
- Components like `RouterLink` will have no types for created routes or rejections
- Compositions will throw "Router not installed" errors
- type guards like `isRoute` will always return false. 

To use router assets like components and hooks they will need to be created and exported using the new `createRouterAssets` utility

```ts
export const {
  RouterLink,
  RouterView,
  useRoute,
  useRouter,
  ...etc
} = createRouterAssets(router)
```

### Global and non global instances
There can only be one global router instance and it must register its type using the `Register` interface. But there can be multiple non global instances. Non global router instances can be used without or in addition to a global router instance.

### DIFF
--- a/src/components/routerView.spec.ts
+++ b/src/components/routerView.spec.ts
@@ -24,6 +24,6 @@ describe('SSR', () => {
 
     const html = await renderToString(app)
 
-    expect(html).toMatchInlineSnapshot(`"hello world"`)
+    expect(html).toMatchInlineSnapshot('"hello world"')
   })
 })
--- a/src/services/valibot.ts
+++ b/src/services/valibot.ts
@@ -5,7 +5,7 @@ import { isPromise } from '@/utilities/promises'
 import { StandardSchemaV1 } from '@standard-schema/spec'
 
 export interface ValibotSchemaLike extends StandardSchemaV1<any> {
-  type: string;
+  type: string,
 }
 
 // inferring the return type is preferred for this function
--- a/src/services/zod.ts
+++ b/src/services/zod.ts
@@ -6,7 +6,7 @@ import { StandardSchemaV1 } from '@standard-schema/spec'
 import type { ZodSchema } from 'zod'
 
 export interface ZodSchemaLike extends StandardSchemaV1<any> {
-  parse: (input: any) => any  
+  parse: (input: any) => any,
 }
 
 let zod: ZodSchemas | null = null
--- a/src/tests/hooks.spec.ts
+++ b/src/tests/hooks.spec.ts
@@ -1,9 +1,9 @@
-import echo from "@/components/echo";
-import { createRoute } from "@/services/createRoute";
-import { createRouter } from "@/services/createRouter";
-import { component } from "@/utilities";
-import { expect, test, vi } from "vitest";
-import { createApp, inject } from "vue";
+import echo from '@/components/echo'
+import { createRoute } from '@/services/createRoute'
+import { createRouter } from '@/services/createRouter'
+import { component } from '@/utilities'
+import { expect, test, vi } from 'vitest'
+import { createApp, inject } from 'vue'
 
 test('hooks are run with the correct context', async () => {
   const beforeRouteHook = vi.fn()
@@ -24,7 +24,7 @@ test('hooks are run with the correct context', async () => {
       const value = inject('global')
 
       afterRouteHook(value)
-    }
+    },
   }, () => ({ value: 'hello' }))
 
   const router = createRouter([route], {
@@ -38,7 +38,7 @@ test('hooks are run with the correct context', async () => {
       const value = inject('global')
 
       afterGlobalHook(value)
-    }
+    },
   })
 
   const app = createApp(component)


src/components/routerLink.browser.spec.ts
```diff
@@ -11,6 +10,7 @@ import { visibilityObserverKey } from '@/compositions/useVisibilityObserver'
 import { VisibilityObserver } from '@/services/createVisibilityObserver'
 import { Url } from '@/types/url'
 import { RouterPushOptions } from '@/types/routerPush'
+import { RouterLink } from '@/main'
```
pleek91, line 13: Eventually, we'll want to update all these tests to use the factories themselves. But continuing to test the global assets for this PR.

src/components/routerView.spec.ts
```diff
@@ -24,6 +24,6 @@ describe('SSR', () => {
 
     const html = await renderToString(app)
 
-    expect(html).toMatchInlineSnapshot('"<!--[-->hello world<!--]-->"')
+    expect(html).toMatchInlineSnapshot(`"hello world"`)
```
pleek91: Using a render function rather than a SFC actually made this test cleaner which was a nice surprise. No `v-if` means no weird comments.

src/compositions/useLink.ts
```diff
@@ -61,112 +61,111 @@ export type UseLinkOptions = RouterPushOptions & {
...
-  const route = computed(() => {
-    const sourceValue = toValue(source)
-    if (typeof sourceValue !== 'string') {
-      return sourceValue
-    }
+type UseLinkFunction<TRouter extends Router> = {
+  <TRouteKey extends RouterRouteName<TRouter>>(name: MaybeRefOrGetter<TRouteKey>, ...args: UseLinkArgs<TRouter, TRouteKey>): UseLink,
+  (url: MaybeRefOrGetter<Url>, options?: MaybeRefOrGetter<UseLinkOptions>): UseLink,
+  (resolvedRoute: MaybeRefOrGetter<ResolvedRoute | undefined>, options?: MaybeRefOrGetter<UseLinkOptions>): UseLink,
+  (source: MaybeRefOrGetter<string | ResolvedRoute | undefined>, paramsOrOptions?: MaybeRefOrGetter<Record<PropertyKey, unknown> | UseLinkOptions>, maybeOptions?: MaybeRefOrGetter<UseLinkOptions>): UseLink,
```
pleek91, line 75: This overload is very wide. I originally removed it (since it matched the implementation) but that caused a type error in RouterLink. So left it for now.

src/compositions/useQueryValue.ts
pleek91, line 1: I don't think this composition really needs to be router specific. But it calls useRouter, so it became router specific as well. Even though I don't think it actually does anything different. And query could technically be shared across routers

src/main.ts
```diff
@@ -1,10 +1,34 @@
...
 export { createRoute } from './services/createRoute'
 export { createExternalRoute } from './services/createExternalRoute'
+export { createRouterAssets } from './services/createRouterAssets'
+
+declare module 'vue' {
+  export interface GlobalComponents {
+    RouterView: typeof RouterView,
+    RouterLink: typeof RouterLink,
+  }
+}
```
pleek91, line 34: I think as some point we'll want to stop doing this automatically and include it in our global router installation just like the register interface.

src/services/createRouter.ts
```diff
@@ -318,20 +323,21 @@ export function createRouter<
...
+    app.provide(getRouterRejectionInjectionKey(routerKey), rejection)
+    app.provide(getRouterHooksKey(routerKey), hooks)
+    app.provide(getPropStoreInjectionKey(routerKey), propStore)
+    app.provide(getComponentsStoreKey(routerKey), componentsStore)
     app.provide(visibilityObserverKey, visibilityObserver)
 
-    // We cant technically guarantee that the user registered the same router that they installed
-    // So we're making an assumption here that when installing a router its the same as the RegisteredRouter
-    app.provide(routerInjectionKey, router as any)
+    app.provide(routerKey, router)
```
pleek91, line 340: Kinda cool this as goes away now

src/main.ts
stackoverfloweth: that would mean users have to manually do this when they have global routers right? idk how I feel about making everybody do this when 99% of users will only use global router. Maybe I missed something though

pleek91: It's basically the same as using the Reguster interface. And it's copy paste, so doesn't seem like it would add any friction. 

We also might be able to check if there is a registered router and conditionally add these types to the global vue interface. 

Either, problem for another day.

PR #577, opened 2025-10-19
Contained router hook types

pleek91:
# Description
Builds on https://github.com/kitbagjs/router/pull/563 by further removing reliance on the `Register` interface for parts of the router itself. This PR accomplishes two things

1. No reliance on `Register` for global hooks and component hooks
Introduces new router specific types for these hook methods. I did this to avoid breaking any of the types for the component and route level hooks. Eventually these will be unified

2. The new types for global hooks preserve type information about the route itself

Before
```ts
router.onBeforeRouteEnter((to) => {
  if(to.name === 'myRoute') {
    to.params.myParam /// unknown
  }
})
```

After
```ts
router.onBeforeRouteEnter((to) => {
  if(to.name === 'myRoute') {
    to.params.myParam // string
  }
})
```


src/models/RouterRouteHooks.ts
```diff
@@ -0,0 +1,10 @@
+import { RouterAfterRouteHook, RouterBeforeRouteHook, Routes } from '@/main'
```
stackoverfloweth: this import surprised me a bit, I wonder if that's intentional? Kinda feels like using barrels again

src/services/createRouter.spec-d.ts
```diff
@@ -0,0 +1,100 @@
+import { createRoute } from '@/services/createRoute'
+import { createRouter } from '@/services/createRouter'
+import { component } from '@/utilities/testHelpers'
+import { describe, test, expectTypeOf } from 'vitest'
+import { AddRouterAfterRouteHook, AddRouterBeforeRouteHook } from '@/types/router'
+import { createRouterPlugin } from './createRouterPlugin'
+import { CallbackContextAbort } from './createCallbackContext'
+
+describe('createRouter', () => {
```
stackoverfloweth: nit: this test doesn't describe its test case very well

src/services/createRouter.spec-d.ts
```diff
@@ -0,0 +1,100 @@
...
+  test('hook functions are correctly typed', () => {
+    expectTypeOf(router.onBeforeRouteEnter).toEqualTypeOf<AddRouterBeforeRouteHook<Routes>>()
+    expectTypeOf(router.onBeforeRouteLeave).toEqualTypeOf<AddRouterBeforeRouteHook<Routes>>()
+    expectTypeOf(router.onBeforeRouteUpdate).toEqualTypeOf<AddRouterBeforeRouteHook<Routes>>()
+    expectTypeOf(router.onAfterRouteEnter).toEqualTypeOf<AddRouterAfterRouteHook<Routes>>()
+    expectTypeOf(router.onAfterRouteLeave).toEqualTypeOf<AddRouterAfterRouteHook<Routes>>()
+    expectTypeOf(router.onAfterRouteUpdate).toEqualTypeOf<AddRouterAfterRouteHook<Routes>>()
+  })
+
+  test('to and from can be narrowed', () => {
```
stackoverfloweth, line 50: :heart:

src/types/router.ts
```diff
@@ -154,13 +156,47 @@ export type Router<
   key: InjectionKey<Router<TRoutes, TOptions, TPlugin>>,
 }
 
+type RouterHookContext<TRoutes extends Routes> = {
+  from: RouterResolvedRouteUnion<TRoutes> | null,
+  reject: RegisteredRouterReject,
```
stackoverfloweth, line 161: should we put this in our sights next?

src/types/router.ts
```diff
@@ -154,13 +156,47 @@ export type Router<
...
+type RouterBeforeRouteHookContext<TRoutes extends Routes> = RouterHookContext<TRoutes> & {
+  abort: CallbackContextAbort,
+}
+
+export type HookContext<TRoutes extends Routes> = {
+  to: RouterResolvedRouteUnion<TRoutes>,
+  from: RouterResolvedRouteUnion<TRoutes> | null,
+}
+
+export type RouterBeforeRouteHook<TRoutes extends Routes> = (to: RouterResolvedRouteUnion<TRoutes>, context: RouterBeforeRouteHookContext<TRoutes>) => MaybePromise<void>
```
stackoverfloweth, line 175: this type feels like it has a lot of overlap with `BeforeRouteHook`. I sort of expected it to replace `BeforeRouteHook` or at least deprecate it? Maybe that's part of the next push where we kill route hooks?

src/types/router.ts
```diff
@@ -154,13 +156,47 @@ export type Router<
...
+export type AddRouterBeforeRouteHook<TRoutes extends Routes> = (hook: RouterBeforeRouteHook<TRoutes>) => RouteHookRemove
+
+type RouterAfterRouteHookContext<TRoutes extends Routes> = RouterHookContext<TRoutes>
+
+export type RouterAfterRouteHook<TRoutes extends Routes> = (to: RouterResolvedRouteUnion<TRoutes>, context: RouterAfterRouteHookContext<TRoutes>) => MaybePromise<void>
```
stackoverfloweth, line 180: same thing here but for `AfterRouteHook`

stackoverfloweth: also `@/models/RouteHooks`

stackoverfloweth (approved):
some questions but mostly just for my understanding, nicely done! 👏

src/models/RouterRouteHooks.ts
pleek91: Yeah these shouldn't be from main. We should figure out a linting rule to prevent this. Wanna take a stab at that?

src/types/router.ts
pleek91: Yep exactly, though that'll be a small breaking change.

src/types/router.ts
pleek91: Yeah exactly, the end of this is to mark all hooks in createRoute as deprecated and all their associated types. 

These new types are after very similar, but they're not exactly the same and most have a generic.

src/models/RouterRouteHooks.ts
pleek91: Fixed

src/services/createRouter.spec-d.ts
pleek91: Good catch, updated

PR #594, opened 2025-11-14
fix: Registered Router Assets type fix

stackoverfloweth:
Adds references to `RouterAssets` with generic for correct export types of registered router assets.

Fixes type issue for all assets but specifically addresses the `useRoute` type issue submitted to us by @undoublethink on discord


pleek91 (approved):
Clean

PR #600, opened 2025-12-07
Add router.onError

pleek91:
## Description

Adds `router.onError()` to register global error handlers for errors thrown during route navigation, including route hooks and props callbacks.

### Error Handling
An error is handled if it an onError hook does not throw an error. Calling any context method (`push`, `replace`, or `reject`) handles the error and navigation is immediately updated. 

If an onError hook throws the original error, then it is considered unhandled. If there are more onError hooks to run then the next hook will be run. If there are no more onError hooks registered then the error will be rethrown within navigation itself and navigation will stope.

```ts
const router = createRouter([...routes])

// callback calls push, error is considered handled
router.onError((error, { push }) => {
  push('NewRoute')
})

// callback calls replace, error is considered handled
router.onError((error, { replace }) => {
  replace('NewRoute')
})

// callback calls reject, error is considered handled
router.onError((error, { reject }) => {
  reject('MyRejection')
})

// callback does not throw, error is considered handled
router.onError((error) => {
  console.error('Navigation error:', error)
  // Error is handled - subsequent handlers won't run
})

// callback throws error, error is not considered handled
router.onError((error) => {
  throw error
})
```

### Error Context
The error handler receives:

- **`error`**: The thrown error
- **`context`**: An object with:
  - `to`: Route being navigated to
  - `from`: Route being navigated from (or `null`)
  - `source`: Error source (`'hook'` or `'props'`)
  - `reject()`: Reject navigation with a rejection type
  - `push()`: Redirect to another route
  - `replace()`: Replace current route


src/services/createRouterHooks.ts
```diff
@@ -124,7 +132,19 @@ export function createRouterHooks<TRouter extends Router>(_routerKey: InjectionK
         return error.response
       }
 
-      throw error
+      try {
```
stackoverfloweth, line 136: is this `try catch` necessary if `runErrorHooks` also wraps each hook with a `try catch`?

src/services/createRouterHooks.ts
```diff
@@ -171,14 +191,51 @@ export function createRouterHooks<TRouter extends Router>(_routerKey: InjectionK
...
```
stackoverfloweth, line 195: same here

stackoverfloweth (approved):
love this, feels like a simple extension of hooks paradigm already established. 

I'm curious where you got the idea for the whole `return true` means stop chain, vs `return false` stops the chain. I feel like it's a nice little add but also feels like it maybe adds complexity/confusion on what's going to happen. I was sort of surprised to think that users would even have a concept of where their hook is the in the greater chain of error hooks?

src/services/createRouterHooks.ts
pleek91: Sorry I actually had a commit I hadn't pushed. `runErrorHooks` doesn't have a try/catch in the final version. I kept trying to make that work but doing a try catch there just ended up making things more complex everywhere else.

PR #601, opened 2025-12-08
Make sure meta property is always defined for matched and matches

pleek91:
# Description
When accessing the route from `router.route` or `useRoute` there are a couple different ways to access the route's meta. You can access the combined meta using `route.meta`. This value is always defined even if there is no meta. It also merges any meta that exists on the parent. 

Sometimes you might want to check the meta for the matched route, and ignore any meta that exists on the parent. You can do this using `route.matched.meta` which is an exact copy of the meta passed when creating the route. However since it was an exact match if no meta was passed then the property would be undefined. This is problematic from a devx perspective because you have to first check that the meta property is even defined in order for typescript to narrow the type and provide the correct type hints. 

```ts
const routeA = createRoute({
  name: 'routeA',
 })

const routeB = createRoute({
  name: 'routeB',
  meta: { pubic: true }
})

const route = useRoute()

route.matched.meta // Property 'meta' does not exist on type ...
```

A route's meta is already defaulted to an empty object at runtime. But that wasn't reflected by the types. After this change the types will reflect the runtime functionality. This makes for a much better devx with easier narrowing and fewer logical checks. 

```ts
const routeA = createRoute({
  name: 'routeA',
 })

const routeB = createRoute({
  name: 'routeB',
  meta: { pubic: true }
})

const route = useRoute()

route.matched.meta // {} | { public: true }
```

### DIFF
--- a/src/services/createRouter.spec-d.ts
+++ b/src/services/createRouter.spec-d.ts
@@ -136,3 +136,55 @@ describe('rejections', () => {
     expectTypeOf<Source>().toEqualTypeOf<Expect>()
   })
 })
+
+describe('route.matchedmeta', () => {
+  test('is always defined', () => {
+    const routeA = createRoute({
+      name: 'routeA',
+    })
+
+    const router = createRouter([routeA])
+
+    expectTypeOf(router.route.matched.meta).toEqualTypeOf<Readonly<{}>>()
+  })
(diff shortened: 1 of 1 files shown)


stackoverfloweth (approved):
❤️

PR #603, opened 2025-12-14
added new createRejection utility

stackoverfloweth:
This PR pivots our rejections from a simple object passed in when the router is created to a 2 step process where users first create the rejections and then pass the created rejections in when creating the router.

```ts
import { createRejection } from '@kitbag/router'

const customRejection = createRejection({
  type: 'Custom',
  component: ...
})

  rejections: [customRejection]
})

router.reject(...)
// expects `'NotFound' | 'Custom'`
```

This not only fits nicely with other paradigms of Kitbag router, but it also reduces dependency on the Register interface and more importantly sets us up nicely to use those rejections in other places.

### DIFF
--- a/docs/advanced-concepts/rejections.md
+++ b/docs/advanced-concepts/rejections.md
@@ -56,7 +56,7 @@ export const router = createRouter(routes, {
 
 ## Trigger Rejection
 
-Any of the [hooks](/advanced-concepts/hooks) will provided a `reject` function in the context argument. The [async prop function](core-concepts/component-props#async-prop-fetching) argument of `createRoute` also will provide a `reject` function.
+Any of the [hooks](/advanced-concepts/hooks) will provided a `reject` function in the context argument. The [async prop function](/core-concepts/component-props#async-prop-fetching) argument of `createRoute` also will provide a `reject` function.
 
 ```ts
 const route = {


src/services/createRouter.ts
```diff
@@ -69,19 +70,19 @@ type RouterUpdateOptions = {
  */
 export function createRouter<
   const TRoutes extends Routes,
-  const TOptions extends RouterOptions,
+  const TOptions extends RouterOptions = {},
```
pleek91, line 73: What is this default needed for?

src/types/routerPlugin.ts
```diff
@@ -1,19 +1,19 @@
...
 }
 
 export type RouterPlugin<
   TRoutes extends Routes = Routes,
-  TRejections extends Record<string, Component> = Record<string, Component>
+  TRejections extends Rejection[] = Rejection[]
 > = {
   routes: TRoutes,
-  rejections: TRejections,
+  rejections?: TRejections,
```
pleek91, line 16: Good catch

src/services/createRejection.ts
```diff
@@ -0,0 +1,10 @@
+import { genericRejection } from '@/components/rejection'
+import { Rejection } from '@/types/rejection'
+import { Component } from 'vue'
+
+export function createRejection<TType extends string>(type: TType, component?: Component): Rejection<TType> {
```
pleek91: Should we just make this an object rather than two positional arguments? Then it'll feel very similar to createRoute. And give us the ability to add options in the future without making a breaking change.

src/services/createRouter.ts
stackoverfloweth: the options argument for `createRouter` is optional, and this is literally the value that will be used. By default if you don't supply a value Typescript infers the whole `RouterOptions`, which has `rejections?: Rejection[]`, which widens the available rejections to `'NotFound' | string`

pleek91 (approved):

PR #607, opened 2025-12-21
Route context property, new route hooks syntax

stackoverfloweth:
This PR introduces an exciting new concept to routes that dramatically improves type safety. When creating a route, you can now also provide a `context` property, which can be other routes and/or rejections.

```ts
import { createRejection, createRoute } from '@kitbag/router'

const parentRejection = createRejection({ type: 'NeedsAuthentication' })
const homeRoute = createRoute({ name: 'home' })

const route = createRoute({
  context: [parentRejection, homeRoute],
})
```

Now both `parentRejection` and `homeRoute` will be accessible to my route hooks without any additional effort. Additionally this context is automatically passed down through any descendants as well.

```ts
route.onBeforeRouteEnter((to, { push, reject }) => {
  push('home')
  // or
  reject('NeedsAuthentication')
})
```

As you might have noticed, we have a new syntax for route level hooks too! Just like router, you'll now create the route and then call a function to register your hook.

```ts
route.onBeforeRouteEnter(() => ...)
route.onBeforeRouteUpdate(() => ...)
route.onBeforeRouteLeave(() => ...)
route.onAfterRouteEnter(() => ...)
route.onAfterRouteUpdate(() => ...)
route.onAfterRouteLeave(() => ...)
```

Or for external routes

```ts
const docsHomepage = createExternalRoute({
  host: 'https://router.kitbag.dev',
  path: '/'
})

docsHomepage.onBeforeRouteEnter(() => ...)
```

The old way of defining route level hooks as part of the options passed into `createRoute` is now deprecated and will be removed in a future release.

### DIFF
--- a/src/services/hooks.browser.spec.ts
+++ b/src/services/hooks.browser.spec.ts
@@ -3,7 +3,7 @@ import { createRoute } from './createRoute'
 import { createRouter } from './createRouter'
 import { h } from 'vue'
 import { mount } from '@vue/test-utils'
-import { RouterView, onBeforeRouteLeave, onBeforeRouteUpdate, onAfterRouteLeave, onAfterRouteUpdate } from '@/main'
+import { RouterView, onBeforeRouteLeave, onBeforeRouteUpdate, onAfterRouteLeave, onAfterRouteUpdate, createExternalRoute } from '@/main'
 import { routes } from '@/utilities/testHelpers'
 
 test('global hooks passed as options are called correctly', async () => {
@@ -198,6 +198,75 @@ test('route hooks are called correctly', async () => {
   expect(parentHooks.afterLeave).toHaveBeenCalledTimes(1)
   expect(parentHooks.afterUpdate).toHaveBeenCalledTimes(2)
   expect(parentHooks.afterEnter).toHaveBeenCalledTimes(1)
+})
+
+test('external route hooks are called correctly', async () => {
+  const internalHooks = {
+    beforeEnter: vi.fn(),
+    beforeLeave: vi.fn(),
+    beforeUpdate: vi.fn(),
+    afterEnter: vi.fn(),
+    afterLeave: vi.fn(),
+    afterUpdate: vi.fn(),
+  }
+
+  const internal = createRoute({
+    name: 'internal',
+    path: '/',
+  })
+
+  internal.onBeforeRouteEnter(() => internalHooks.beforeEnter())
+  internal.onBeforeRouteLeave(() => internalHooks.beforeLeave())
+  internal.onBeforeRouteUpdate(() => internalHooks.beforeUpdate())
+  internal.onAfterRouteEnter(() => internalHooks.afterEnter())
+  internal.onAfterRouteLeave(() => internalHooks.afterLeave())
+  internal.onAfterRouteUpdate(() => internalHooks.afterUpdate())
+
+  const externalHooks = {
+    beforeEnter: vi.fn(),
+  }
+
+  const external = createExternalRoute({
+    name: 'external',
+    host: 'https://kitbag.dev',
+    path: '/',
+  })
+
+  external.onBeforeRouteEnter(() => externalHooks.beforeEnter())
+
+  const router = createRouter([internal, external], {
+    initialUrl: '/',
+  })
+
+  const root = {
+    template: '<RouterView />',
+  }
+
+  mount(root, {
+    global: {
+      plugins: [router],
+    },
+  })
+
+  await router.start()
+
+  expect(internalHooks.beforeEnter).toHaveBeenCalledTimes(1)
+  expect(internalHooks.beforeUpdate).toHaveBeenCalledTimes(0)
+  expect(internalHooks.beforeLeave).toHaveBeenCalledTimes(0)
+  expect(internalHooks.afterLeave).toHaveBeenCalledTimes(0)
+  expect(internalHooks.afterUpdate).toHaveBeenCalledTimes(0)
+  expect(internalHooks.afterEnter).toHaveBeenCalledTimes(1)
+  expect(externalHooks.beforeEnter).toHaveBeenCalledTimes(0)
+
+  await router.push('external', { param: 'param2' })
+
+  expect(internalHooks.beforeEnter).toHaveBeenCalledTimes(1)
+  expect(internalHooks.beforeUpdate).toHaveBeenCalledTimes(0)
+  expect(internalHooks.beforeLeave).toHaveBeenCalledTimes(1)
+  expect(internalHooks.afterLeave).toHaveBeenCalledTimes(1)
+  expect(internalHooks.afterUpdate).toHaveBeenCalledTimes(0)
+  expect(internalHooks.afterEnter).toHaveBeenCalledTimes(1)
+  expect(externalHooks.beforeEnter).toHaveBeenCalledTimes(1)
 })
 
 test('component hooks are called correctly', async () => {
--- a/src/types/hooks.ts
+++ b/src/types/hooks.ts
@@ -64,11 +64,6 @@ export type InternalRouteHooks<TContext extends RouteContext[] | undefined = und
    * Registers a hook to be called after a route is updated.
    */
   onAfterRouteUpdate: AddRouterAfterRouteHook<RouteContextToRoute<TContext>, RouteContextToRejection<TContext>>,
-  /**
-   * Registers a hook to be called when an error occurs.
-   * If the hook returns true, the error is considered handled and the other hooks are not run. If all hooks return false the error is rethrown
-   */
-  onError: AddRouterErrorHook<RouteContextToRoute<TContext>, RouteContextToRejection<TContext>>,
 }
 
 export type ExternalRouteHooks<TContext extends RouteContext[] | undefined = undefined> = {


src/models/RouteHooks.ts
```diff
@@ -1,5 +1,8 @@
 import { AfterRouteHook, BeforeRouteHook } from '@/types/hooks'
 
+/**
+ * @deprecated will be removed in a future version
+ */
```
pleek91, line 5: This is deprecated because now we just use the hook store everywhere?

src/services/createExternalRoute.ts
```diff
@@ -5,14 +5,18 @@ import { toName } from '@/types/name'
...
   const TOptions extends CreateRouteOptions & WithHost & WithoutParent
 >(options: TOptions): ToRoute<TOptions, undefined>
+  & ExternalRouteHooks<TOptions['context']>
 
 export function createExternalRoute<
   const TOptions extends CreateRouteOptions & WithoutHost & WithParent
 >(options: TOptions): ToRoute<TOptions, undefined>
+  & ExternalRouteHooks<TOptions['context']>
 
 export function createExternalRoute(options: CreateRouteOptions & (WithoutHost | WithHost)): Route {
```
pleek91, line 21: Not part of this PR but do you know why `createExternalRoute` accepts `WithoutHost`? Isn't host what makes it external?

src/services/createExternalRoute.ts
```diff
@@ -22,12 +26,15 @@ export function createExternalRoute(options: CreateRouteOptions & (WithoutHost |
   const hash = toWithParams(options.hash)
   const meta = options.meta ?? {}
   const host = toWithParams(options.host)
+  const context = options.context ?? []
+  const { store, ...hooks } = createRouteHooks(context)
```
pleek91: we should probably specifically destructure the hooks we expect.  I know its enforces in types but we probably also don't want users hacking things into runtime that we don't support or expect to work. 


```suggestion
  const { store, onBeforeRouteLeave } = createRouteHooks(context)
```

src/services/createRoute.ts
```diff
@@ -30,21 +33,26 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
   const hash = toWithParams(options.hash)
   const meta = options.meta ?? {}
   const state = options.state ?? {}
+  const context = options.context ?? []
+  const { store, ...hooks } = createRouteHooks(context)
```
pleek91: Same here, we should desctrure and only return the hooks we expect (onError isn't supported on route level hooks for example).

src/services/createRouteHooks.ts
```diff
@@ -0,0 +1,83 @@
...
+  onBeforeRouteLeave: AddRouterBeforeRouteHook<TRoutes, TRejections>,
+  onAfterRouteEnter: AddRouterAfterRouteHook<TRoutes, TRejections>,
+  onAfterRouteUpdate: AddRouterAfterRouteHook<TRoutes, TRejections>,
+  onAfterRouteLeave: AddRouterAfterRouteHook<TRoutes, TRejections>,
+  onError: AddRouterErrorHook<TRoutes, TRejections>,
+  store: RouterRouteHooks,
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function createRouteHooks<TContext extends RouteContext[]>(_context: TContext): RouteHooks<
```
pleek91: Do we actually need to pass the context in or can we just use the type?

```suggestion
export function createRouteHooks<TContext extends RouteContext[]>(): RouteHooks<
```

src/services/createRouter.ts
```diff
@@ -150,6 +145,10 @@ export function createRouter<
         throw new Error(`Switch is not exhaustive for before hook response status: ${JSON.stringify(beforeResponse satisfies never)}`)
     }
 
+    if (isExternal(url)) {
+      return
+    }
```
pleek91: This makes me wonder if we can support after hooks on the `from` when navigating from an internal route to an external route. Something for future us to consider.

src/services/createRouterReject.ts
```diff
@@ -40,6 +40,7 @@ export function createRouterReject(rejections: Rejection[]): CreateRouterReject
       id: route.id,
       matched: route,
       matches: [route],
+      hooks: [],
```
pleek91, line 43: now that we have `createRejection`, I think we could technically add hooks to rejections 🤔 going to have to think about the implications/use cases there

src/services/createRouterRoute.ts
```diff
@@ -109,6 +109,7 @@ export function createRouterRoute<TRoute extends ResolvedRoute>(routerKey: Injec
     id,
     matched,
     matches,
+    hooks,
```
pleek91, line 112: hmmm, I can't think of a reason to not expose this on the route but also feels like internal implementation detail that users shouldn't really know about. Thoughts?

pleek91: Like `matched` and `matches` we have a contract for and expect users to access them (to get meta for example) but `hooks` I don't think we have a contract for an I'd expect us to change this without caring about end user code. So I wonder if we should either not expose this on the router route.

src/services/getRouteHooks.spec.ts
```diff
@@ -2,57 +2,51 @@ import { expect, test, vi } from 'vitest'
...
+  const from = createResolvedRoute(grandchildB, {})
 
   const hooks = getBeforeRouteHooksFromRoutes(to, from)
 
-  expect(Array.from(hooks.onBeforeRouteEnter)).toMatchObject([...asArray(childA.onBeforeRouteEnter), ...asArray(grandchildA.onBeforeRouteEnter)])
-  expect(Array.from(hooks.onBeforeRouteUpdate)).toMatchObject([...asArray(parent.onBeforeRouteUpdate)])
-  expect(Array.from(hooks.onBeforeRouteLeave)).toMatchObject([...asArray(childB.onBeforeRouteLeave), ...asArray(grandchildB.onBeforeRouteLeave)])
+  expect(Array.from(hooks.onBeforeRouteEnter)).toMatchObject([...Array.from(childA.hooks.at(0)?.onBeforeRouteEnter ?? []), ...Array.from(grandchildA.hooks.at(0)?.onBeforeRouteEnter ?? [])])
+  expect(Array.from(hooks.onBeforeRouteUpdate)).toMatchObject([...Array.from(parent.hooks.at(0)?.onBeforeRouteUpdate ?? [])])
+  expect(Array.from(hooks.onBeforeRouteLeave)).toMatchObject([...Array.from(childB.hooks.at(0)?.onBeforeRouteLeave ?? []), ...Array.from(grandchildB.hooks.at(0)?.onBeforeRouteLeave ?? [])])
```
pleek91: I was trying to grok this and took me a minute. Once I realized what it was doing it made me think of `matched` vs `matches`. I don't think we have a reason to do this other than tests, but having a similar pattern for hooks came to mind. Something to consider in the future if this sort of thing comes up again.

src/types/hooks.ts
```diff
@@ -38,6 +39,45 @@ export type WithHooks = {
...
+  onAfterRouteLeave: AddRouterAfterRouteHook<RouteContextToRoute<TContext>, RouteContextToRejection<TContext>>,
+  /**
+   * Registers a hook to be called after a route is updated.
+   */
+  onAfterRouteUpdate: AddRouterAfterRouteHook<RouteContextToRoute<TContext>, RouteContextToRejection<TContext>>,
+  /**
+   * Registers a hook to be called when an error occurs.
+   * If the hook returns true, the error is considered handled and the other hooks are not run. If all hooks return false the error is rethrown
+   */
+  onError: AddRouterErrorHook<RouteContextToRoute<TContext>, RouteContextToRejection<TContext>>,
```
pleek91: In a previous comment I mentioned `onError` not being supported for route level hooks. But I see it here. Does this work or is this here by mistake

src/types/resolved.ts
```diff
@@ -45,6 +46,11 @@ export type ResolvedRoute<TRoute extends Route = Route> = Readonly<{
    * String value of the resolved URL.
    */
   href: Url,
+  /**
+   * The stores for routes including ancestors.
+   * Order of routes will be from greatest ancestor to narrowest matched.
+   */
+  hooks: RouterRouteHooks[],
```
pleek91, line 54: Should this be marked as `internal`? For that matter should the entire `ResolvedRoute` type be `internal`? I'm trying to remember if we ever suggest users actually use the this object directly or if we just ask them to pass it to another utility or RouterLink. 

I think `hooks` should be internal. The whole type I'm not sure about (question for future us)

src/types/routeContext.ts
```diff
@@ -0,0 +1,36 @@
...
+  : undefined extends TContext
+    ? Route[]
+    : FilterRouteContextRoutes<TContext>
+
+type FilterRouteContextRoutes<TContext extends RouteContext[] | undefined> =
+TContext extends [infer First, ...infer Rest extends RouteContext[]]
+  ? First extends GenericRoute
+    ? [First, ...FilterRouteContextRoutes<Rest>]
+    : FilterRouteContextRoutes<Rest>
+  : []
```
pleek91, line 22: With this approach context is a tuple. But does it need to be? Dealing with tuple types often leads to more complexity (and slower types) than unions. Curious if you considered making context a union. Then rather than filtering a tuple and creating a new tuple you could do something like
```ts
type RouteContextRoutes<TContext extends RouteContext[]> = Extract<TContext[number], GenericRoute>

type RouteContextRejections<TContext extends RouteContext[]> = Extract<TContext[number], Rejection>
```

pleek91: Not 100% sure that would work, but that's what came to mind

src/models/RouteHooks.ts
stackoverfloweth: currently this is only used in the now deprecated `getRouteHooks`. I think it's been replaced by the very similar [RouterRouteHooks](https://github.com/kitbagjs/router/blob/main/src/models/RouterRouteHooks.ts)

src/services/createExternalRoute.ts
stackoverfloweth: not off the top of my head, but I'm pretty sure it's setup that way because you can create an external route without `host` if it has an ancestor that does have `host`

src/services/createRoute.ts
stackoverfloweth: yeah, as you notice lower - I _was_ expecting to support `onError` but currently does NOT actually work. I added it to my todos unless you feel strongly that it shouldn't be a route level hook

src/services/createRouteHooks.ts
stackoverfloweth: yeah, I swear I removed this once before 🤷

src/services/createRouter.ts
stackoverfloweth: good catch. I looked into this and actually think it should be pretty easy to fix. Take a look

src/services/createRouterRoute.ts
stackoverfloweth: originally I think this _was_ marked internal - so I think I agree

stackoverfloweth: I also took this opportunity to add all the other missing jsdoc strings to this file

src/services/getRouteHooks.spec.ts
stackoverfloweth: I also considered the same thing but ultimately felt like it would feel icky to add a thing for tests only

src/types/hooks.ts
stackoverfloweth: does not work but noted as something to revisit before we actually publish/update docs

src/types/resolved.ts
stackoverfloweth: I made `hooks` internal here, on `RouterRoute` and on `Route`. Also unsure about making the whole type internal

src/types/routeContext.ts
stackoverfloweth: I agree with this theory, though the types you provided didn't work for me. I spent 15-20 minutes trying to make a simpler union type work but didn't get anywhere. It could be the way we're capturing the context in `createRoute`? Problem for future us?

src/services/getRouteHooks.spec.ts
stackoverfloweth: ironic, this logic is actually wrong 😆 `hooks.at(0)` is the parent's hooks, was getting false positives until I switched to `.at(-1)`

src/services/createExternalRoute.ts
pleek91: Ah yeah that makes sense.

src/services/createRouter.ts
```diff
@@ -150,6 +145,33 @@ export function createRouter<
         throw new Error(`Switch is not exhaustive for before hook response status: ${JSON.stringify(beforeResponse satisfies never)}`)
     }
 
+    if (!isExternal(url)) {
+      setPropsAndUpdateRoute(navigationId, to, from)
+    }
+
+    const afterResponse = await hooks.runAfterRouteHooks({ to, from })
+
```
pleek91, line 153: This seems like trying to support after hooks when navigating to external routes. Does that work? I expect that as soon as history.update is called the browser starts navigating away.

src/services/hooks.browser.spec.ts
```diff
@@ -16,14 +16,15 @@ test('global hooks passed as options are called correctly', async () => {
 
   const router = createRouter(routes, {
     initialUrl: '/parentA/valueA',
-    onBeforeRouteEnter,
-    onBeforeRouteUpdate,
-    onBeforeRouteLeave,
-    onAfterRouteEnter,
-    onAfterRouteUpdate,
-    onAfterRouteLeave,
```
pleek91, line 24: Should we keep these in tests until they are removed?

src/types/hooks.ts
pleek91: Let's just remove it from this type then. We can always add it later.

src/services/hooks.browser.spec.ts
stackoverfloweth: 🤷 I don't think so but I'm open to it if you feel strongly

src/services/createRouter.ts
stackoverfloweth: added a test to confirm this does in fact work, take a look at `hooks.browser.spec.ts` "external route hooks are called correctly" in last commit

craig-pplx (approved):

PR #608, opened 2025-12-21
add new rule to route matching logic that checks for host

stackoverfloweth:
`host` property only exists on [external routes](https://router.kitbag.dev/core-concepts/external-routes.html#external-routes).  Currently, router has never had to worry about route matching for external routes because as soon as the `Set` function see's an external URL, it returns before attempting to find an underlying route.

```ts
if (isExternal(url)) {
  history.update(url, options)
  return
}
```
https://github.com/kitbagjs/router/blob/main/src/services/createRouter.ts#L117C9-L117C19

From the router's perspective, it doesn't matter if a user pushes to a named external route

```ts
router.push('docs.homepage')
```

or if the user navigates to a URL that's not internal

```ts
router.push('https://router.kitbag.io')
```

This is likely to change when we add support for `beforeRouteEnter` hooks on external routes.  So this PR enables better route matching by checking that the route either has no `host`, or the `host` matches the incoming URL. Note: because this follows the existing patterns in place for things like `path` and `query`, it will also work for hosts that have params.

### DIFF
--- a/src/services/routeMatchRules.spec.ts
+++ b/src/services/routeMatchRules.spec.ts
@@ -7,11 +7,10 @@ import { withParams } from '@/services/withParams'
 import { createExternalRoute } from '@/services/createExternalRoute'
 
 describe('routeHostMatches', () => {
-  test('given url without host, returns true', () => {
+  test('given url without host and a route without a host, returns true', () => {
     const urlWithoutHost = '/somewhere?with=query'
-    const route = createExternalRoute({
-      name: 'with-host',
-      host: 'www.kitbag.io',
+    const route = createRoute({
+      name: 'without-host',
     })
 
     const response = routeHostMatches(route, urlWithoutHost)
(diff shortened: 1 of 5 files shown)


src/services/routeMatchRules.spec.ts
```diff
@@ -1,9 +1,67 @@
...
+
+  test('given route.host without host, returns true', () => {
+    const route = createRoute({
+      name: 'without-host',
+    })
+
+    const response = routeHostMatches(route, 'http://www.kitbag.io/')
+
+    expect(response).toBe(true)
+  })
```
pleek91: These seem like they should expect false to me

src/services/routeMatchRules.ts
```diff
@@ -1,12 +1,23 @@
...
   return 'name' in route.matched && !!route.matched.name
 }
 
+export const routeHostMatches: RouteMatchRule = (route, url) => {
+  const { host } = parseUrl(url)
+  const hostPattern = generateRouteHostRegexPattern(route)
+
+  if (!host || !route.host.value) {
+    return true
+  }
```
pleek91: Should this be an && rather than an || ?

stackoverfloweth: this is covering all our internal routes that don't define a host, if `!host` skip this test

stackoverfloweth: WRT the `host` property on the URL, technically when the router uses this service I'd expect the URL to always include a host but because this utility takes `url: string`, it needed to be flexible and account for a potentially relative URL. I'm open to treating that as a fail? Basically remove the `!host` part and let the regex test an empty string

pleek91: Just by looking at the test it seems like this will pass if a regular route matches an external route. Is that not the case?

stackoverfloweth: good call @pleek91, this was certainly incorrect so I'm glad you stuck to your guns

PR #613, opened 2025-12-22
Lean into new Rejection type, reduce complexity of RouterHooks store

stackoverfloweth:
These changes are cherry-picked from the bigger RouterContext work.

This PR specifically addresses 2 side quests

1.) Now that we have [createRejection](https://github.com/kitbagjs/router/pull/603), our types that were previously expecting `string` for rejection types should instead expect the `Rejection` type

2.) When we added the concept of a hooks store to the route, we talked about how the types for what hooks are available and what should the context be only matters in the `Route` type we export to users. The store itself just knows that it has an array of hooks for each lifecycle type. I applied that same thinking to the `RouteHooks` type we have that we use in the router. The router was previously passing in the router key and adding a bunch of generics to communicate context between the router and the hooks. In practice, that's all behind the scenes and can be simplified to just arrays of hooks.

### DIFF
--- a/src/types/router.ts
+++ b/src/types/router.ts
@@ -11,7 +11,7 @@ import { RouterResolve, RouterResolveOptions } from '@/types/routerResolve'
 import { RouterReject } from '@/types/routerReject'
 import { RouterPlugin } from '@/types/routerPlugin'
 import { RoutesName } from '@/types/routesMap'
-import { Rejection } from '@/types/rejection'
+import { ExtractRejections, Rejection } from '@/types/rejection'
 
 /**
  * Options to initialize a {@link Router} instance.
@@ -51,8 +51,6 @@ export type RouterOptions = WithHooks & {
    */
   isGlobalRouter?: boolean,
 }
-
-type ExtractRejections<T> = T extends { rejections: infer TRejections extends Rejection[] } ? TRejections : []
 
 export type Router<
   TRoutes extends Routes = any,
--- a/src/types/routerPlugin.ts
+++ b/src/types/routerPlugin.ts
@@ -1,17 +1,13 @@
 import { BeforeRouteHook, AfterRouteHook } from './hooks'
 import { Routes } from './route'
 import { MaybeArray } from './utilities'
-import { Rejection } from './rejection'
+import { Rejection, ExtractRejections } from './rejection'
 
 export type EmptyRouterPlugin = RouterPlugin<[], []>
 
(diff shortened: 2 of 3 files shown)


src/types/router.ts
```diff
@@ -105,32 +107,32 @@ export type Router<
   /**
    * Registers a hook to be called before a route is entered.
    */
-  onBeforeRouteEnter: AddRouterBeforeRouteHook<TRoutes | TPlugin['routes'], RejectionType<TOptions['rejections']> | RejectionType<TPlugin['rejections']>>,
+  onBeforeRouteEnter: AddRouterBeforeRouteHook<TRoutes | TPlugin['routes'], [...ExtractRejections<TOptions>, ...ExtractRejections<TPlugin>]>,
```
pleek91, line 108: Can't this just be TOptions['rejections'] | TPlugin['rejections'] ?

src/types/router.ts
```diff
@@ -171,7 +173,7 @@ export type RouterRoutes<TRouter extends Router> = TRouter extends Router<infer
   : Routes
 
 export type RouterRejections<TRouter extends Router> = TRouter extends Router<any, infer TOptions extends RouterOptions, infer TPlugins extends RouterPlugin>
-  ? RejectionType<TOptions['rejections']> | RejectionType<TPlugins['rejections']>
+  ? [...TOptions['rejections'] extends Rejection[] ? TOptions['rejections'] : [], ...TPlugins['rejections'] extends Rejection[] ? TPlugins['rejections'] : []]
```
pleek91: I'm guessing this condition is why we need ExtractRejections. Can we use that here?

src/types/routerPlugin.ts
```diff
@@ -3,21 +3,37 @@ import { Routes } from './route'
...
-> = {
-  routes: TRoutes,
-  rejections?: TRejections,
+export type ToRouterPlugin<TPlugin extends CreateRouterPluginOptions> = TPlugin extends { routes: infer TRoutes extends Routes }
+  ? TPlugin extends { rejections: infer TRejections extends Rejection[] }
+    ? RouterPlugin<TRoutes, TRejections>
+    : RouterPlugin<TRoutes, []>
+  : TPlugin extends { rejections: infer TRejections extends Rejection[] }
+    ? RouterPlugin<[], TRejections>
+    : RouterPlugin<[], []>
```
pleek91: And here?

src/types/router.ts
stackoverfloweth: maybe for `TOptions`, but this was how I fixed that weird enumeration of `TPlugins` issue we found last time we chatted.

PR #615, opened 2025-12-23
check for missing route context

stackoverfloweth:
This PR adds a check that gets called on `createRouter` similar to `checkDuplicateNames`. This new check looks for route context provided but the underlying routes are not also supplied.

**Why do we need the user to supply routes if they're just needed for a single route?**
If a route hook pushes to a context route, the URL will reflect this change. If the end user refreshes the page but the route doesn't actually exist within the router, it would result in a NotFound.

**What about rejections?**
This PR does not check for missing rejections that are included in route.context but not in createRouter options. Unlike routes, the rejection state of the router is not tied to an external source like the URL. Users would always "dismiss" the rejection and go back to the underlying route for the URL if they refresh. So the idea of context-specific rejections is probably okay and might actually be preferred over forcing users to always have all rejections be globally accessible. Individual routes could even theoretically override the global rejection handling for their sub-route.

### DIFF
--- a/src/utilities/checkMissingContext.ts
+++ b/src/utilities/checkMissingContext.ts
@@ -3,13 +3,13 @@ import { Route, Routes } from '@/types/route'
 import { RouteContext } from '@/types/routeContext'
 
 export function checkMissingContext(routes: Routes): void {
-  const routeIds = new Set(routes.map(route => route.id)
+  const routeIds = new Set(routes.map((route) => route.id))
   const routeIdsFromContext = routes
     .flatMap((route) => route.context)
     .filter(contextIsRoute)
 
   for (const route of routeIdsFromContext) {
-    if (!routeIds.has(route.id) {
+    if (!routeIds.has(route.id)) {
       throw new MissingRouteContextError(route.name)
     }
   }


pleek91:
> If a route hook pushes to a context route, the URL will reflect this change. If the end user refreshes the page but the route doesn't actually exist within the router, it would result in a NotFound.

@stackoverfloweth FWIW I don't think it would even be able to do the push because it would first have to resolve the route right?

src/utilities/checkMissingContext.ts
```diff
@@ -0,0 +1,19 @@
...
+  const routeIdsFromContext = routes
+    .flatMap((route) => route.context)
+    .filter(contextIsRoute)
+
+  for (const route of routeIdsFromContext) {
+    if (routes.every(({ id }) => route.id !== id)) {
+      throw new MissingRouteContextError(route.name)
+    }
+  }
+}
```
pleek91, line 16: We should probably make this more efficient. It looks like we're looping over all the routes for each context route. Lets use a `Set` for all the route ids

```suggestion
export function checkMissingContext(routes: Routes): void {
  const routeIds = new Set(routes.map(route => route.id)
  const routeIdsFromContext = routes
    .flatMap((route) => route.context)
    .filter(contextIsRoute)

  for (const route of routeIdsFromContext) {
    if (!routeIds.has(route.id) {
      throw new MissingRouteContextError(route.name)
    }
  }
}
```

stackoverfloweth:
> > If a route hook pushes to a context route, the URL will reflect this change. If the end user refreshes the page but the route doesn't actually exist within the router, it would result in a NotFound.
> 
> @stackoverfloweth FWIW I don't think it would even be able to do the push because it would first have to resolve the route right?

correct, but we could _theoretically_ support ephemeral/contextual routes

PR #616, opened 2025-12-24
Add typed context to route props

pleek91:
## Description
Builds on top of https://github.com/kitbagjs/router/pull/607 to support the `context` property of `createRoute` for prop getters. The callback context for prop getters use to rely on the `Register` interface, now it is strictly typed based on the `context` option of `createRoute`. 

```ts
const contextRejection = createRejection({
  type: 'NotAuthorized',
})

const contextRoute = createRoute({
  name: 'contextRoute',
})

createRoute({
  name: 'route',
  component: echo,
  context: [contextRejection, contextRoute],
}, (route, { push, replace, reject ) => {
  reject('NotAuthorized') // rejection is valid and type safe
  push('contextRoute') // route name is valid and type safe
  replace('contextRoute') // route name is valid and type safe
})
```

### DIFF
--- a/src/services/createExternalRoute.ts
+++ b/src/services/createExternalRoute.ts
@@ -45,7 +45,7 @@ export function createExternalRoute(options: CreateRouteOptions & (WithoutHost |
     state: {},
     context,
     onBeforeRouteEnter,
-  } satisfies Route & ExternalRouteHooks
+  }
 
   const merged = isWithParent(options) ? combineRoutes(options.parent, route) : route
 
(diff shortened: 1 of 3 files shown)


stackoverfloweth (approved):
I think I can probably fix the `PropsGetter` type to not be so wide when `context` isn't provided, which I think is worthwhile so we can keep the satisfies on `createRoute` and `createExternalRoute`. However, that doesn't need to block this - looks good 👍

PR #620, opened 2025-12-25
Remove all deprecated hooks

pleek91:
# Description
Now that we have fully migrated hooks over to the new pattern where you add hooks after creating the route, router, or plugin, we're going to remove the deprecated options passed in when creating a route, router, or plugin. 

To migrate your hooks simply change this
```ts
const route = createRoute({
  name: 'MyRoute',
  onBeforeRouteEnter: () => { ... }
})
```
To this
```ts
const route = createRoute({
  name: 'MyRoute'
})

route.onBeforeRouteEnter(() => { ... })
```

The same applies to global routes on routers or in plugins
```ts
const router = createRouter([...], {
   onBeforeRouteEnter: () => { ... }
})
```
become
```ts
const router = createRouter([...])

router.onBeforeRouteEnter(() => { ... })
```

and
```ts
const plugin = createRouterPlugin({
   onBeforeRouteEnter: () => { ... }
})
```

becomes
```ts
const plugin = createRouterPlugin()

plugin.onBeforeRouteEnter(() => { ... })
```

### DIFF
--- a/src/types/routerPlugin.ts
+++ b/src/types/routerPlugin.ts
@@ -7,6 +7,7 @@ import { RouterReject } from './routerReject'
 import { RouterPush } from './routerPush'
 import { RouterReplace } from './routerReplace'
 import { CallbackContextAbort } from '@/services/createRouterCallbackContext'
+import { MaybePromise } from './utilities'
 
 export type EmptyRouterPlugin = RouterPlugin<[], []>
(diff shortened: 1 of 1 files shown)


stackoverfloweth (approved):
this is so beautiful I almost teared up 🥹

PR #621, opened 2025-12-26
Fix circular type annotation issue with Register type

pleek91:
## Description
Every since https://github.com/kitbagjs/router/pull/563 where `createRouterAssets` was introduced we've been chasing our tails around circular type annotation issues. 

image

Finally tracked this down to referencing `typeof RouterView` in `CreateRouteOptions` because when no component is provided `RouterView` is used as the default. But we can easily just statically type this based on the prop type for `RouterView` rather than trying to infer it. Which appears to fix the circular annotation issue.

### DIFF
--- a/src/types/createRouteOptions.ts
+++ b/src/types/createRouteOptions.ts
@@ -12,13 +12,13 @@ import { ResolvedRoute } from './resolved'
 import { ComponentProps } from '@/services/component'
 import { PropsCallbackContext } from './props'
 import { MaybePromise } from './utilities'
-import { RouterView } from '@/main'
 import { ToMeta } from './meta'
 import { ToState } from './state'
 import { ToName } from './name'
 import { WithHooks } from './hooks'
 import { ToWithParams, WithParams } from '@/services/withParams'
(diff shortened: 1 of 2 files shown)


src/types/createRouteOptions.ts
```diff
@@ -125,6 +125,10 @@ export type PropsGetter<
   TComponent extends Component = Component
 > = (route: ResolvedRoute<ToRoute<TOptions, undefined>>, context: PropsCallbackContext<TOptions>) => MaybePromise<ComponentProps<TComponent>>
 
+export type RouterViewPropsGetter<
+  TOptions extends CreateRouteOptions = CreateRouteOptions
+> = (route: ResolvedRoute<ToRoute<TOptions, undefined>>, context: PropsCallbackContext<TOptions>) => MaybePromise<RouterViewProps & Record<string, unknown>>
```
pleek91, line 130: Props generally accept additional props, and we have many tests that do not provide a component but then define random props for testing purposes. The `& Record<string, unknown>` make it so you can pass any additional props.

stackoverfloweth (approved):
🚀🚀🚀

PR #623, opened 2025-12-26
using readonly Rejection[] similar to Routes type

stackoverfloweth:
While updating documentation I went to add this example to plugins.md

```ts
import { createRouterPlugin } from '@kitbag/router'

const routes = [
  createRoute({ name: 'home', path: '/' }),
] as const

const rejections = [
  createRejection({ type: 'RequiresAuth' }),
] as const

const plugin = createRouterPlugin({
  routes,
  rejections,
})
```

which gave me the following error on `createRouterPlugin` under the `rejections` property. 

> The type `readonly [Rejection<'RequiresAuth'>]` is readonly and cannot be assigned to the mutable type `Rejection[]`

and confirmed a similar error on `createRouter` when rejections is readonly and passed into router options. This is also consistent with how we handle `Routes` which is `readonly Route[]`.


pleek91 (approved):
I like it

PR #626, opened 2025-12-27
Docs tweaks

pleek91:
# Description
Re-read through our entire docs and made some updates/tweaks where I felt necessary.

### DIFF
--- a/docs/core-concepts/router.md
+++ b/docs/core-concepts/router.md
@@ -129,7 +129,7 @@ router.reject('NotFound')
 
 ### Find
 
-Returens a resolved route for a url or `undefined` if the url does not match any route.
+Returns a resolved route for a url or `undefined` if the url does not match any route.
 
 ```ts
 router.find('/users/1')


docs/core-concepts/router.md
```diff
@@ -129,17 +129,17 @@ router.reject('NotFound')
 
 ### Find
 
-Creates a URL for a given route and params.
+Returens a resolved route for a url or `undefined` if the url does not match any route.
```
stackoverfloweth:
```suggestion
Returns a resolved route for a url or `undefined` if the url does not match any route.
```

Can't believe you're not using cursor/spell-checker plugin and still out here misspelling words in 2026

docs/core-concepts/router.md
```diff
@@ -129,17 +129,17 @@ router.reject('NotFound')
...
 
 ```ts
-router.find('users', { id: 1 })
+router.find('/users/1')
```
stackoverfloweth, line 135: big oversight here, good catch

PR #627, opened 2025-12-27
simplify hook syntax without changing hooks

stackoverfloweth:
I started down the path of fixing types for before/after hooks so that `to` and `from` would be more accurate. The current syntax for hooks feels overly verbose, mostly in the overuse of the word "route". This is partly confusing because we now use hooks in places other than routes (like plugins) but also because when we created the new paradigm we prefixed the new stuff with `RouterRoute..`.

This PR seeks to simplify the naming convention of hooks so that it's easier to grok and offers us the ability to meaningfully differentiate "route" hooks from "plugin" hooks or "router" hooks if needed..

99% of this PR is F2 rename
the last 1% is 
- updating `createRouterHooks` service to reuse the `createHooksFactory` service where it was previously duplicating the same idea.
- splitting the args for hook runners from `(context: {to, from})` to 2 arguments `(to, from)` which helps avoid a duplicate `HookContext` that was only used in 1 other file

### DIFF
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -120,7 +120,7 @@ export function createRouter<
     const to = find(url, options) ?? getRejectionRoute('NotFound')
     const from = getFromRouteForHooks(navigationId)
 
-    const beforeResponse = await hooks.runBeforeRouteHooks(to, from)
+    const beforeResponse = await hooks.runBeforeRouteHooks({ to, from })
 
     switch (beforeResponse.status) {
       // On abort do nothing
@@ -153,7 +153,7 @@ export function createRouter<
       setPropsAndUpdateRoute(navigationId, to, from)
     }
 
-    const afterResponse = await hooks.runAfterRouteHooks(to, from)
+    const afterResponse = await hooks.runAfterRouteHooks({ to, from })
 
     switch (afterResponse.status) {
       case 'PUSH':
--- a/src/services/createRouterHooks.ts
+++ b/src/services/createRouterHooks.ts
@@ -38,7 +38,7 @@ export function createRouterHooks(): RouterHooks {
 
   const { reject, push, replace, abort } = createRouterCallbackContext()
 
-  const runBeforeRouteHooks: BeforeHookRunner = async (to, from) => {
+  const runBeforeRouteHooks: BeforeHookRunner = async ({ to, from }) => {
     const routeHooks = getBeforeHooksFromRoutes(to, from)
     const globalHooks = getGlobalBeforeHooks(to, from, globalStore)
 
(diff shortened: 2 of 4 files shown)


src/services/createRouter.ts
```diff
@@ -120,7 +120,7 @@ export function createRouter<
     const to = find(url, options) ?? getRejectionRoute('NotFound')
     const from = getFromRouteForHooks(navigationId)
 
-    const beforeResponse = await hooks.runBeforeRouteHooks({ to, from })
+    const beforeResponse = await hooks.runBeforeRouteHooks(to, from)
```
pleek91: I'd prefer we don't use positional arguments for this sort of thing.

src/services/createRouter.ts
```diff
@@ -153,7 +153,7 @@ export function createRouter<
       setPropsAndUpdateRoute(navigationId, to, from)
     }
 
-    const afterResponse = await hooks.runAfterRouteHooks({ to, from })
+    const afterResponse = await hooks.runAfterRouteHooks(to, from)
```
pleek91: Same here

src/services/createRouterHooks.ts
```diff
@@ -142,19 +98,18 @@ export function createRouterHooks(): RouterHooks {
...
-    const routeHooks = getAfterRouteHooksFromRoutes(to, from)
-    const globalHooks = getGlobalAfterRouteHooks(to, from, global)
+  const runAfterRouteHooks: AfterHookRunner = async (to, from) => {
+    const routeHooks = getAfterHooksFromRoutes(to, from)
+    const globalHooks = getGlobalAfterHooks(to, from, globalStore)
 
-    const allHooks: RouterAfterRouteHook[] = [
-      ...component.onAfterRouteLeave,
+    const allHooks: AfterHook[] = [
+      ...componentStore.onAfterRouteLeave,
```
pleek91, line 106: Seems like this should be named "componentHooks" since the others are "globalHooks" and "routeHooks"

src/types/hooks.ts
```diff
@@ -8,110 +8,98 @@ import { RouterReplace } from './routerReplace'
...
-export type RouteHookBeforeRunner = (context: BeforeHookContext) => Promise<BeforeRouteHookResponse>
-
-export type AfterHookContext = {
-  to: ResolvedRoute,
-  from: ResolvedRoute | null,
-}
-
-export type RouteHookAfterRunner = (context: AfterHookContext) => Promise<AfterRouteHookResponse>
+export type HookBeforeRunner = (to: ResolvedRoute, from: ResolvedRoute | null) => Promise<BeforeHookResponse>
+export type HookAfterRunner = (to: ResolvedRoute, from: ResolvedRoute | null) => Promise<AfterHookResponse>
```
pleek91: Yeah this is different but not better IMO.

PR #629, opened 2025-12-28
add onError hook support to route level

stackoverfloweth:
In [a recent PR](https://github.com/kitbagjs/router/pull/600) we added a new `onError` hook to the router. This PR extends support for this hook to individual routes.

```ts
const route = createRoute({ name: 'my-route' })

route.onError((error, context) => {
  ...
})
```

Just like the router hooks, these route level hooks are passed the error as well as the same context

  | Property | Type | Description |
  |--|--|--|
  | to       | RouterResolvedRouteUnion<TRoutes>        | The destination route that was being navigated to when the error occurred                             |
  | from     | RouterResolvedRouteUnion<TRoutes> \| null | The source route being navigated from. Will be null if there is no previous route                     |
  | source   | 'props' \| 'hook' \| 'component'           | The origin of the error - indicates whether the error occurred in route props, a hook, or a component |
  | reject   | RouterReject<TRejections>                | Function to reject the navigation with a custom rejection                                             |
  | push     | RouterPush<TRoutes>                      | Function to navigate to a different route using push navigation                                       |
  | replace  | RouterReplace<TRoutes>                   | Function to navigate to a different route using replace navigation 

```ts
const route = createRoute({ name: 'my-route' })

route.onError((error, { to, from, source, push, replace, reject }) => {
  if (error instanceof MyExpectedError) {
    push(...)
  }

  throw reject('AppError')
})
```

### DIFF
--- a/src/services/createExternalRoute.ts
+++ b/src/services/createExternalRoute.ts
@@ -27,7 +27,7 @@ export function createExternalRoute(options: CreateRouteOptions & (WithoutHost |
   const meta = options.meta ?? {}
   const host = toWithParams(options.host)
   const context = options.context ?? []
-  const { store, onBeforeRouteEnter } = createHooksFactory()
+  const { store, onBeforeRouteEnter, onError } = createHooksFactory()
   const rawRoute = markRaw({ id, meta: {}, state: {}, ...options })
 
   const route = {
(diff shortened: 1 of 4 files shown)


pleek91:
@stackoverfloweth If I remember the question I had about this PR was seems like we shouldn't need to do extra wrappers around the hooks because there's already a try catch around running the hooks that runs the global onError hooks. So I think we could add route level onError hooks to that existing system.

PR #630, opened 2025-12-28
Add `update` method to hooks

pleek91:
# Description
Hooks already support `push` and `replace` but previously they did not support `update`. This PR adds the `update` method to hooks so that the `to` route can be modified. This works exactly the same was as `route.update` on `router.route` but its pre bound to the `to` route.

```ts
const route = createRoute({
  name: 'MyRoute',
  path: '/my/path/[Mparam]',
})

route.onBeforeRouteEnter((to, { update }) => {
  if (to.params.myParam === 'this-value') {
    update({ Mparam: 'that-value' })
  }
})
```

So when `MyRoute` is entered the above hook will run. If the url is `/my/path/this-value` it will be updated to `/my-path/that-value`. 

## Note About Types
Since `update` is typed based on the `to` route, in route level hooks it will be narrowed to the specific route the hook is for. But for global and component level hooks the `to` is a union of all the possible routes. This means that unless all of your routes have a param in common, the `update` method's first argument will be type `never`. This is an accurate reflection of your routes, but means that for global and component level hooks the `update` method is likely unusable in most cases and the `push` method will be better suited since `push` requires specifying the desired route by its name.

### DIFF
--- a/src/services/createRoute.spec-d.ts
+++ b/src/services/createRoute.spec-d.ts
@@ -630,7 +630,7 @@ describe('props', () => {
   })
 
   describe('update', () => {
-    test('accepts params when no context is provided', () => {
+    test('accepts params based on the current route', () => {
       createRoute({
         name: 'route',
         path: '/[paramName]',


src/types/hooks.ts
```diff
@@ -96,47 +107,54 @@ export type AfterHookLifecycle = 'onAfterRouteEnter' | 'onAfterRouteUpdate' | 'o
...
-  TRejections extends Rejections = Rejections
-> = HookContext<TRoutes, TRejections>
+> = (hook: BeforeHook<TRoute, TRoutes, TRejections>) => HookRemove
 
 export type AfterHook<
+  TRoute extends Route = Route,
   TRoutes extends Routes = Routes,
   TRejections extends Rejections = Rejections
-> = (to: RouterResolvedRouteUnion<TRoutes>, context: AfterHookContext<TRoutes, TRejections>) => MaybePromise<void>
+> = (to: RouterResolvedRouteUnion<TRoutes>, context: AfterHookContext<TRoute, TRoutes, TRejections>) => MaybePromise<void>
```
stackoverfloweth, line 151: I'm a bit confused by the simplicity of just adding `TRoute` to before/after hooks. I expected needing both `TTo` and `TFrom` generics. The single route implies that the only route we can accurately narrow is the `from` in context but that's not true right? If the hook is `beforeRouteLeave`, `afterRouteLeave`, and maybe `..Update` this makes sense. But if the hook is `beforeRouteEnter` or `afterRouteEnter` wouldn't `TRoute` actually be the `to`? 

```ts
const myRoute = createRoute({ ... })
myRoute.onBeforeRouteEnter((to, context) => {
  to
// ^? typeof myRoute??
  context.from
// ^? typeof RouterResolvedRouteUnion<TRoutes>??
})
```

pleek91: Valid confusion. 

So this PR kinda skirts the to/from issue which I'll fix in a follow up. 

This TRoute generic will end up being either the to or the from depending on the hook. But this follows our current behavior. But you're right the types aren't quite right yet because of the issues we've discussed with the to/from types.

PR #633, opened 2025-12-29
attempt to automate docs generation

stackoverfloweth:


.github/workflows/auto-docs.yml
```diff
@@ -0,0 +1,37 @@
+name: Auto Generate Docs
+
+on:
+  schedule:
+    # Run every Monday at 9:00 AM UTC
```
pleek91: Rather than on a cron can we just have this run in commits to main? Then if we merge a PR it should run automatically.

pleek91: Or maybe in new tags in main? That would generate the docs after each release which might make the most sense.

PR #634, opened 2025-12-31
Accurately type to and from routes on hooks

pleek91:
# Description
Specifically for route level hooks the types for `to` and `from` are not accurate. Specifically they are typed based on the context which is completely inaccurate. 

For enter and update hooks on routes we know that the `to` value is the route the hook was defined for. Same for the `from` on leave hooks. 

So we can accurately reflect those types in the hooks themselves. This fixes a few issues
1. `params.foo` will be correctly typed by default rather than being type `unknown`
2. the `from` route is not longer typed based on the context and correctly is typed as `ResolvedRoute` because it could be any route not just one the user passed as context (same goes for the `to` argument of leave hooks)
3. the `from` route for leave hooks is no longer nullish. A leave hook always has a from route. This applies to all levels of hooks (router, component, plugin) not just route level hooks. 

```ts
const blog = createRoute({
  name: 'blog',
  path: '/blog'
})

const blogPost = createRoute({
  context: [blog],
  parent: [blog],
  name: 'blogPost',
  path: '/[blogPostId]'
})

// Before this PR
blogPost.onBeforeRouteEnter((to, ({ from }) => {
  to.name // "blog" | "blogPost" 
  from?.name // "blog" | "blogPost" 
})

// After this PR
blogPost.onBeforeRouteEnter((to, ({ from }) => {
  to.name //  | "blogPost" 
  from?.name // string
})
```

### DIFF
--- a/src/services/createComponentHooks.ts
+++ b/src/services/createComponentHooks.ts
@@ -6,12 +6,12 @@ import { Routes } from '@/types/route'
 import { Router, RouterRejections, RouterRoutes } from '@/types/router'
 import { Rejections } from '@/types/rejection'
 
-function createComponentHook<TRouter extends Router>(routerKey: InjectionKey<TRouter>, lifecycle: 'onBeforeRouteEnter'): AddBeforeEnterHook<RouterRoutes<TRouter>[number], RouterRoutes<TRouter>[number], RouterRoutes<TRouter>, RouterRejections<TRouter>>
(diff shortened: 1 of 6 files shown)


src/services/createRoute.spec-d.ts
```diff
@@ -756,6 +757,44 @@ describe('matches[number].meta', () => {
 })
 
 describe('hooks', () => {
+  test('to and from are typed correctly', () => {
```
pleek91, line 760: Seems like a lot of changes. But seeing this test pass felt good.

PR #637, opened 2026-01-02
Basic devtools integration that lists routes

pleek91:
# Description
Adds a new devtools integration for Kitbag Router. Right now it just lists the available routes with their details. Which we can hopefully build off of

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "homepage": "https://github.com/kitbagjs/router#readme",
   "scripts": {
     "build": "vite build",
-    "build:watch": "vite build --watch --minify=false",
+    "dev": "vite build --watch --minify=false",
     "test": "vitest",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
   },
   "dependencies": {
     "@standard-schema/spec": "^1.0.0",
+    "@vue/devtools-api": "^8.0.5",
```
stackoverfloweth, line 73: should this be in `dependencies`?

pleek91: I'm honestly not 100% sure. But I was referencing both [vue-router](https://github.com/kitbagjs/router/pull/637) and [pinia](https://github.com/vuejs/pinia/blob/v3/packages/pinia/package.json#L75) for this and they both have it as a dependency.

PR #643, opened 2026-01-08
Automatically include context routes

stackoverfloweth:
Previously if a router was given a route that has a route in `context` that's not also provided to the same router, it would throw a `MissingRouteContextError` error at runtime. This was our solution to ensuring that if the user choses to `push` to this route in a hook, the router _must_ need to know how to actually resolve that route!

However we came up with a potentially better DX - just automatically add missing routes to the router at runtime. These discovered routes will not be available in the types but otherwise work exactly the same as if they were explicitly provided.

### DIFF
--- a/src/services/getRoutesForRouter.spec.ts
+++ b/src/services/getRoutesForRouter.spec.ts
@@ -39,10 +39,13 @@ test('given named routes inside plugins, includes them in the response', () => {
 
 test('given named routes inside route context, includes them in the response', () => {
   const relatedRoute = createRoute({ name: 'related' })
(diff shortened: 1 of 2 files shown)


src/devtools/createRouterDevtools.ts
```diff
@@ -133,7 +133,7 @@ type RouterDevtoolsProps = {
 /**
  * Sets up Vue DevTools integration for Kitbag Router.
  */
-export function setupRouterDevtools({ router, app, routes: routesArray }: RouterDevtoolsProps): void {
```
pleek91: Side quest? I prefer having the names reflect what they are since both the map and the array are used in this file.

PR #644, opened 2026-01-08
update routeMatchScore logic, specifically around host and hash

stackoverfloweth:
I stumbled upon a bug with both host and hash matching logic. These properties are optionally defined on the route AND technically optional on the URL. 

We have route matching rules for each, which will basically return `true` if the route doesn't define it. For example, let's consider host. If the url is `https://router.kitbag.dev/foo` and you have both an internal and external routes with `path: '/foo'` BOTH routes will pass the host matches rule. Similarly if hash is on a url `/bar#42` and you have 2 routes with `path: '/bar'` and 1 has hash and 1 doesn't BOTH routes will pass the hash matches rule.

Then these routes are sorted by 
- prefer lower depth
- prefer truthy host
- prefer truthy hash
- prefer optional params fulfilled

### DIFF
--- a/src/tests/routeProps.browser.spec.ts
+++ b/src/tests/routeProps.browser.spec.ts
@@ -40,7 +40,6 @@ test('components are not remounted when props change', async () => {
   })
 
   await router.start()
-  console.log('After start, current route:', router.route.name, router.route.href, 'id:', router.route.id)
 
   await router.route.update({ param: 'foo' })
 
@@ -55,10 +54,8 @@ test('components are not remounted when props change', async () => {
   expect(setup).toHaveBeenCalledTimes(1)
 
   await router.push('routeB')
-  console.log('After push to routeB, current route:', router.route.name, router.route.href, 'id:', router.route.id)
 
   await router.push('routeA', { param: 'foo' })
-  console.log('After push to routeA, current route:', router.route.name, router.route.href, 'id:', router.route.id, 'setup count:', setup.mock.calls.length)
 
   expect(setup).toHaveBeenCalledTimes(2)
 })


src/services/getMatchesForUrl.spec.ts
```diff
@@ -153,25 +152,3 @@ test('given route with equal matches, returns route with highest score', () => {
 
   expect(match.name).toBe('second-route')
 })
-
-test.each([
```
stackoverfloweth, line 157: this test was moved to routeMatchScore, since that's really what's being tested here

src/services/getMatchesForUrl.ts
```diff
@@ -1,11 +1,12 @@
 import { routeParamsAreValid } from '@/services/paramValidation'
-import { isNamedRoute, routePathMatches, routeQueryMatches, routeHashMatches } from '@/services/routeMatchRules'
+import { isNamedRoute, routeHostMatches, routePathMatches, routeQueryMatches, routeHashMatches } from '@/services/routeMatchRules'
 import { getRouteScoreSortMethod } from '@/services/routeMatchScore'
 import { Route, Routes } from '@/types/route'
 import { RouteMatchRule } from '@/types/routeMatchRule'
 
 const rules: RouteMatchRule[] = [
   isNamedRoute,
+  routeHostMatches,
```
stackoverfloweth, line 9: we actually do want this check here, if a route has a host defined that does not match, it should NOT be considered

src/services/routeMatchRules.ts
```diff
@@ -37,13 +29,7 @@ export const routeQueryMatches: RouteMatchRule = (route, url) => {
...
 export const routeHashMatches: RouteMatchRule = (route, url) => {
   const { hash } = parseUrl(url)
-  const { value } = route.hash
-
-  if (!stringHasValue(value)) {
-    return true
-  }
-
-  const cleanHash = `#${value.replace(/^#*/, '')}`
+  const hashPattern = generateRouteHashRegexPattern(route)
```
stackoverfloweth, line 32: I updated this to follow the same paradigm of using regexp, actual implementation is the same but this sets us up nicely for adding params hopefully in the near future

src/services/routeMatchScore.spec.ts
stackoverfloweth, line 1: this is how I found these bugs, by testing the sort BOTH directions, I removed some false positive results

src/services/routeMatchScore.ts
stackoverfloweth, line 1: I think this file is a lot more readable now, apologies for the difficult diff to grok

src/tests/routeProps.browser.spec.ts
stackoverfloweth, line 1: this test started failing and I couldn't understand why.. Then I realized it should NOT have worked as it was written

So notice that RouteA has path `/[param]` and RouteB has path `/routeB`.

When we push to RouteB on 56, the router builds a URL ‘/routeB’, which when run through the route matching logic SHOULD correctly match that to RouteA.

src/services/routeMatchRules.spec.ts
```diff
@@ -274,6 +263,25 @@ describe('routeQueryMatches', () => {
 })
 
 describe('routeHashMatches', () => {
+  test.each([
+    ['we*23mf#0'],
```
pleek91, line 267: What is this lol

PR #650, opened 2026-01-13
Add route level redirects

pleek91:
# Description
Adds the ability to add redirects to routes to allow redirecting one route to another. Route redirects can be added in two ways, using new `route.redirectTo` and `route.redirectFrom` methods on the route

```ts
const routeA = createRoute({...})
const routeB = createRoute({...})

routeA.redirectTo(routeB)

or

routeB.redirectFrom(routeA)
```

When the `to` route has params you'll be required to pass those params as a getter. If the `from` route also has params they will be passed to the getter.

```ts
const from = createRoute({
  path: '/from-path/[fromParam]`
  ...
})

const to = createRoute({
  path: '/to-path/[toParam]'
  ...
})

from.redirectTo(to, ({ fromParam }) => {
  return { toParam: fromParam }
})
```

A route can only have a single redirect. So if you attempt to add a second redirect to a route a `MultipleRouteRedirectsError` will be thrown.

```ts
const to = createRoute({ ... })
const from = createRoute({ ... })

from.redirectTo(to) // no error

from.redirectTo(anotherRoute) // error

to.redirectFrom(from) // error because `from` already has a redirect assigned to it.
```

### DIFF
--- a/src/services/createExternalRoute.ts
+++ b/src/services/createExternalRoute.ts
@@ -15,6 +15,7 @@ export function createExternalRoute<
   const TOptions extends CreateRouteOptions & WithHost & WithoutParent
 >(options: TOptions): ToRoute<TOptions>
   & ExternalRouteHooks<ToRoute<TOptions>, TOptions['context']>
+  & RouteRedirects<ToRoute<TOptions>>
 
 export function createExternalRoute<
   const TOptions extends CreateRouteOptions & WithoutHost & WithParent
@@ -31,7 +32,7 @@ export function createExternalRoute(options: CreateRouteOptions & (WithoutHost |
   const meta = options.meta ?? {}
   const host = toWithParams(options.host)
   const context = options.context ?? []
-  const { store, onBeforeRouteEnter } = createRouteHooks()
+  const { store, ...hooks } = createRouteHooks()
   const redirects = createRouteRedirects({
     getRoute: () => route,
   })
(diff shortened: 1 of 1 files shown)


src/services/createRouteRedirects.ts
```diff
@@ -0,0 +1,53 @@
...
+   * The name of the route that that is being redirected to in the redirectFrom callback.
+   */
+  getRoute: () => Route & InternalRouteRedirects,
+}
+
+export function createRouteRedirects({ getRoute }: CreateRouteRedirectsContext): InternalRouteRedirects {
+  const redirectTo: RouteRedirectTo = (to, convertParams) => {
+    const from = getRoute()
+
+    const redirects = from.hooks.at(-1)?.redirects
```
pleek91: this is the current problem. Since route has an array of hooks that is merged with the parent (and includes any children) we don't actually have a reference to the current route's hook store.

src/services/createRouteRedirects.ts
```diff
@@ -0,0 +1,53 @@
...
+      throw new Error('Route hooks not found')
+    }
+
+    if (redirects.size > 0) {
+      throw new MultipleRouteRedirectsError(from.name)
+    }
+
+    redirects.add((from, { replace }) => {
+      replace(to.name, convertParams(from.params))
+    })
```
pleek91: This also type errors, but I'm not to worried about that at the moment.

src/services/createExternalRoute.ts
```diff
@@ -28,7 +31,10 @@ export function createExternalRoute(options: CreateRouteOptions & (WithoutHost |
   const meta = options.meta ?? {}
   const host = toWithParams(options.host)
   const context = options.context ?? []
-  const { store, onBeforeRouteEnter } = createHooksFactory()
+  const { store, onBeforeRouteEnter } = createRouteHooks()
```
pleek91: needs to grab `redirect`

PR #652, opened 2026-01-14
createUrl utility, integrates with route, replaces several services

stackoverfloweth:
This PR adds a new low-level utility `createUrl`. It enables users to parse full url strings into parts, join url parts into a url string, and more importantly can leverage router's powerful params syntax. 

```ts
const hasParams = createUrl({
  host: 'https://kitbag.dev',
  pathname: '/[id]',
  query: 'color=red',
  hash: '#[page]',
})

// can call toString on the whole url, any params will be required
hasParams.toString({ id: '14', page: '123' })
//  https://kitbag.dev/14?color=red#123

// can also call toString on individual parts
hasParams.host.toString()
//  https://kitbag.dev

// if "part" requires params, it'll be requires by toString
hasParams.path.toString({ id: '14' })
//  /14

// can call parse to get the params from the url
const params = hasParams.parse('https://kitbag.dev/14?color=red#123')
//  { id: '14', page: '123' }

// retain access to the WithParams object
const hostParams = hasParams.host.schema.params
//     ^?
```

This new utility assumes the responsibility of a bunch of old services
- paramValidation
- urlAssembly
- urlCreator
- urlParser

All of those spec files were combined into the new `createUrl.spec.ts` to prove that it successfully accomplishes what these services used to do.

### DIFF
--- a/src/types/createRouteOptions.ts
+++ b/src/types/createRouteOptions.ts
@@ -19,7 +19,6 @@ import { ToWithParams, WithParams } from '@/services/withParams'
 import { RouteContext, ToRouteContext } from './routeContext'
 import { RouterViewProps } from '@/components/routerView'
 import { createUrl, CreateUrlOptions } from '@/services/createUrl'
-import { Url } from '@/types/url'
 
 export type WithHost<THost extends string | WithParams = string | WithParams> = {
   /**


src/services/createIsExternal.spec.ts
```diff
@@ -12,7 +12,7 @@ test('given undefined host, returns true', () => {
 })
 
 test('given host with url that matches, returns false', () => {
-  const host = 'router.kitbag.dev'
+  const host = 'https://router.kitbag.dev'
```
stackoverfloweth, line 15: `parseUrl` took separate `protocol` and `host` and if protocol was omitted, it would assume "https". New utility assumes protocol inside host, doesn't make assumptions

src/services/createResolvedRoute.ts
```diff
@@ -1,30 +1,32 @@
...
-import { getRouteParamValues } from './paramValidation'
+import { createUrl } from '@/services/createUrl'
+import { combineUrlSearchParams } from '@/utilities/urlSearchParams'
 
 export function createResolvedRoute(route: Route, params: Record<string, unknown> = {}, options: RouterResolveOptions = {}): ResolvedRoute {
-  const href = assembleUrl(route, {
-    params: params,
-    query: options.query,
-    hash: options.hash,
+  const updatedRoute = createUrl({
```
stackoverfloweth, line 10: assembleUrl took extra `query` and `host`. I played around with the idea of `toString()` on url doing the same, but ultimately decided `createResolvedRoute` was a good spot

src/services/createUrl.ts
```diff
@@ -0,0 +1,197 @@
...
+      ...getParams(path.schema, parts.pathname),
+      ...getQueryParams(query.schema, parts.search),
+      ...getParams(hash.schema, parts.hash),
+    }
+  }
+
+  return { host, path, query, hash, toString, parse }
+}
+
+function assembleParamValues(part: WithParams, paramValues: Record<string, unknown>): string {
```
stackoverfloweth, line 96: Everything below here was previously non-exported functions inside the services I removed. I'm open to moving them into their own files and exporting them

src/services/routeMatchRules.spec.ts
```diff
@@ -347,3 +348,156 @@ describe('routeHashMatches', () => {
     expect(response).toBe(true)
   })
 })
+
+describe('routeParamsAreValid', () => {
```
stackoverfloweth, line 352: the method `routeParamsAreValid` used to be part of param validation service. Now it's a simple try catch around `route.parse()`. I choose to put that with other routeMatchRules and moved the tests accordingly

src/services/routeMatchRules.ts
```diff
@@ -1,31 +1,42 @@
...
 export const routeHashMatches: RouteMatchRule = (route, url) => {
-  const { hash } = parseUrl(url)
+  const { hash } = createUrl(url)
   const hashPattern = generateRouteHashRegexPattern(route)
 
-  return hashPattern.test(hash)
+  return hashPattern.test(hash.schema.value)
+}
+
+export const routeParamsAreValid: RouteMatchRule = (route, url) => {
```
stackoverfloweth, line 34: here's that rule I moved

src/utilities/makeOptional.ts
```diff
@@ -7,3 +7,6 @@ export type MakeOptional<T> = {
 } & {
   [P in Exclude<keyof T, WithOptionalProperties<T>>]: T[P];
 }
+
+export type UnionToIntersection<U> =
```
stackoverfloweth: this is used in my "simplification" of `ExtractRouteParamTypesReading` and `ExtractRouteParamTypesWriting`

src/types/params.ts
```diff
@@ -67,36 +67,49 @@ export type ExtractParamName<
...
  * @returns A record of parameter names to their respective types, extracted and merged from both path and query parameters.
  */
-export type ExtractRouteParamTypesReading<TRoute extends Route> =
-  Identity<
-    MakeOptional<
-      ExtractParamTypesReading<TRoute['host']> &
-      ExtractParamTypesReading<TRoute['path']> &
-      ExtractParamTypesReading<TRoute['query']> &
-      ExtractParamTypesReading<TRoute['hash']>
+export type ExtractRouteParamTypesReading<TParts extends Record<PropertyKey, unknown> | Url> =
```
stackoverfloweth: these changes were intended to make this type of more generic. Rather than expecting `Route` or new `Url`, it takes any record and combines any values that are `WithParams` type

src/services/createUrl.ts
```diff
@@ -0,0 +1,174 @@
...
+  ToWithParams<T['hash']>
+>
+
+// https://en.wikipedia.org/wiki/.invalid
+const FALLBACK_HOST = 'https://internal.invalid'
+
+export function createUrl<const T extends CreateUrlOptions>(options: T): ToUrl<T>
+export function createUrl(url: string): Url
+export function createUrl(url: string | CreateUrlOptions): Url
+export function createUrl(urlOrOptions: CreateUrlOptions | string): Url {
```
pleek91, line 28: Shouldn't we be using `UrlString` here rather than just `string`? Or is the expectation that the URL class will error if its invalid?

src/services/createUrl.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  return { host, path, query, hash, toString, parse }
+}
+
+function assembleParamValues(part: WithParams, paramValues: Record<string, unknown>): string {
+  return Object.keys(part.params).reduce((url, name) => {
+    return setParamValueOnUrl(url, part, name, paramValues[name])
+  }, part.value)
```
pleek91, line 99: nit: I know this was pre-existing but `Object.entries` seems like the right choice here

```suggestion
  return Object.entries(part.params).reduce((url, [name, value]) => {
    return setParamValueOnUrl(url, part, name, value)
  }, part.value)
```

src/types/params.ts
```diff
@@ -67,36 +67,60 @@ export type ExtractParamName<
...
+export type ExtractRecordParamTypesReading<TParts extends Record<PropertyKey, unknown>> =
   Identity<
     MakeOptional<
-      ExtractParamTypesReading<TRoute['host']> &
-      ExtractParamTypesReading<TRoute['path']> &
-      ExtractParamTypesReading<TRoute['query']> &
-      ExtractParamTypesReading<TRoute['hash']>
+      UnionToIntersection<
+        { [K in keyof ExtractWithParams<TParts>]: ExtractParamTypesReading<ExtractWithParams<TParts>[K]> }[keyof ExtractWithParams<TParts>]
+      >
```
pleek91: The previous version is much easier to understand/grok. Why not just accept the known parts like we were before and do the same `TParts['host'] & TParts['path']` type logic we were before? I don't think we need to be dynamic here.

src/types/params.ts
```diff
@@ -67,36 +67,60 @@ export type ExtractParamName<
...
+ */
+export type ExtractUrlParamTypesWriting<TParts extends Url> = ExtractRecordParamTypesWriting<{ host: TParts['host']['schema'], path: TParts['path']['schema'], query: TParts['query']['schema'], hash: TParts['hash']['schema'] }>
+
+/**
+ * Extracts combined types for any properties that are WithParams, creating a unified parameter object.
  * @template TParams - The record of parameter types, possibly including undefined.
  * @returns A new type with the appropriate properties marked as optional.
  */
-type ExtractParamTypesReading<TWithParams extends WithParams> = {
+type ExtractParamTypesReading<TWithParams extends { value: string, params: Record<string, Param> }> = {
```
pleek91: Isn't this just a more generic version of what the `WithParams` type is?

src/types/createRouteOptions.ts
```diff
@@ -169,10 +166,10 @@ export type ToRoute<
...
     ? Route<
       ToName<TOptions['name']>,
-      ToWithParams<TParent['host']>,
-      CombinePath<ToWithParams<TParent['path']>, ToWithParams<TOptions['path']>>,
-      CombineQuery<ToWithParams<TParent['query']>, ToWithParams<TOptions['query']>>,
-      CombineHash<ToWithParams<TParent['hash']>, ToWithParams<TOptions['hash']>>,
+      ExtractRouteUrlGenerics<TParent>['host'],
+      CombinePath<ExtractRouteUrlGenerics<TParent>['path'], ToWithParams<TOptions['path']>>,
+      CombineQuery<ExtractRouteUrlGenerics<TParent>['query'], ToWithParams<TOptions['query']>>,
+      CombineHash<ExtractRouteUrlGenerics<TParent>['hash'], ToWithParams<TOptions['hash']>>,
```
pleek91: I don't understand why we need this extra extract type. TParent and TOptions have these properties why cant we just access them directly as before?

src/types/resolved.ts
```diff
@@ -37,7 +37,7 @@ export type ResolvedRoute<TRoute extends Route = Route> = Readonly<{
   /**
    * Key value pair for route params, values will be the user provided value from current browser location.
   */
-  params: ExtractRouteParamTypesReading<TRoute>,
+  params: ExtractUrlParamTypesReading<TRoute>,
```
pleek91: nit: Its a bit confusing that we're passing a route into a url params type here. We should probably just have ExtractRouteParamTypes utilities too

src/types/route.ts
```diff
@@ -36,7 +37,7 @@ export type Route<
   TState extends Record<string, Param> = Record<string, Param>,
   TMatches extends CreatedRouteOptions[] = CreatedRouteOptions[],
   TContext extends RouteContext[] = RouteContext[]
-> = {
+> = Url<THost, TPath, TQuery, THash> & {
```
pleek91, line 40: Would it simplify anything to make the url a property on route rather than making it a union with Url?

src/types/url.ts
```diff
@@ -0,0 +1,65 @@
...
+  hash?: string,
+}
+
+type UrlParamsArgs<
+  TParts extends Record<string, unknown> = Record<string, unknown>
+> = ExtractRecordParamTypesWriting<TParts> extends Record<string, unknown>
+  ? [params?: Record<string, unknown>, options?: ToStringOptions]
+  : AllPropertiesAreOptional<ExtractRecordParamTypesWriting<TParts>> extends true
+    ? [params?: ExtractRecordParamTypesWriting<TParts>, options?: ToStringOptions]
+    : [params: ExtractRecordParamTypesWriting<TParts>, options?: ToStringOptions]
```
pleek91: I feel like there is a simpler way of accomplishing this. Something like
```ts
host: {
  schema: THost,
  toString: ToStringCallback<THost>
},
toString: ToStringCallback<THost & TPath & TQuery & THash>
```

```ts
type ToStringCallback<TParams extends WithParams> = AllPropertiesAreOptional<ExtractParamTypesWriting<TParams>> extends true
 ? (params?: ExtractParamTypesWriting<TParams>, options?: ToStringOptions]) => string
  :  (params: ExtractParamTypesWriting<TParams>, options?: ToStringOptions]) => string
```

src/types/url.ts
```diff
@@ -0,0 +1,65 @@
...
+  TQuery extends WithParams = WithParams,
+  THash extends WithParams = WithParams
+> = {
+  /**
+   * Represents the host for this route. Used for external routes.
+  */
+  host: {
+    schema: THost,
+    toString: (...params: UrlParamsArgs<{ host: THost }>) => string,
+  },
```
pleek91: thoughts on eliminating `schema` and doing something like 

```ts
host: THost & { toString: ... }
```

src/services/createRouter.ts
```diff
@@ -246,11 +246,11 @@ export function createRouter<
     if (isUrlString(source)) {
       const options: RouterPushOptions = { ...paramsOrOptions }
       const url = combineUrl(source, {
-        searchParams: options.query,
+        query: new URLSearchParams(options.query).toString(),
```
pleek91, line 249: Could we just use `createUrl` again here rather than using URLSearchParams directly?

src/services/createRoute.ts
```diff
@@ -40,27 +41,31 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
...
-    host: withParams(),
     prefetch: options.prefetch,
+    ...url,
     ...hooks,
   } satisfies Route & InternalRouteHooks
 
   const merged = isWithParent(options) ? combineRoutes(options.parent, route) : route
 
-  checkDuplicateParams(merged.path.params, merged.query.params, merged.hash.params)
+  checkDuplicateParams(merged.path.schema.params, merged.query.schema.params, merged.hash.schema.params)
```
pleek91: For future us: This utility should probably just accept `Route` or `Url`

src/services/createRoute.ts
```diff
@@ -40,27 +41,31 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
   const { store, ...hooks } = createHooksFactory()
   const rawRoute = markRaw({ id, meta, state, ...options, props })
 
+  const url = createUrl({
+    host: withParams(),
+    path,
+    query,
+    hash,
+  })
```
pleek91, line 55: can't host be undefined? Doesn't seem like you should have to pass a host in here.

stackoverfloweth: yeah, good catch

src/types/url.ts
stackoverfloweth: (note to self) also remove `options`, that not needed?

src/types/url.ts
stackoverfloweth: you're a genius 🧠

src/services/createRouter.ts
stackoverfloweth: this required a bit of effort to widen the available input for `query`, which was previously `string | WithParams | undefined`, now includes `QuerySource` but that feels worthwhile for the benefit of users. This also means users can create `query` on routes with new syntax options

```ts
createRoute({
  query: new URLSearchParams('....')
})

createRoute({
  query: {
    foo: 'bar',
    zoo: 'zar',
  }
})

createRoute({
  query: [['foo', 'bar'], ['zoo', 'zar']]
})
```

pleek91: FWIW this type of syntax change should probably be a separate PR so we can communicate to users.

stackoverfloweth: gonna "stash" my POC here and tackle this in a separate follow-up PR 

```ts
export type QuerySourceToWithParams<T extends QuerySource | string | WithParams | undefined> = T extends string | WithParams | undefined
  ? ToWithParams<T>
  : WithParams<string, {}>

export function querySourceToWithParams<T extends QuerySource | string | WithParams | undefined>(querySource: T): QuerySourceToWithParams<T>
export function querySourceToWithParams(querySource: QuerySource | string | WithParams | undefined): WithParams {
  if (typeof querySource === 'string' || typeof querySource === 'undefined' || isWithParams(querySource)) {
    return toWithParams(querySource)
  }

  const normalized = new URLSearchParams(querySource)

  return withParams(normalized.toString(), {})
}
```

src/types/params.ts
stackoverfloweth: yeah, this is leftovers from an abandoned test

src/services/createUrl.ts
stackoverfloweth: actually this is incorrect. `paramValues` is provided by the user, params entries value is the param definition. We can't use object entries on paramValues either because it might not have all the keys

src/services/createUrl.ts
stackoverfloweth: I like the idea of being wider, accept more inputs but be narrow in outputs

pleek91: But wont it error if you don't pass in a value UrlString?

stackoverfloweth: no, `new URL()` works with quite a lot of different strings

src/services/createRoute.ts
```diff
@@ -40,21 +41,24 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
   const { store, ...hooks } = createHooksFactory()
   const rawRoute = markRaw({ id, meta, state, ...options, props })
 
+  const url = createUrl({
+    path,
+    query,
+    hash,
+  })
```
craig-pplx, line 55: To be honest I don't really understand why we're using createUrl internally like this. What is the benefit vs just using the existing WithParams types and utilities.

src/services/createUrl.ts
```diff
@@ -0,0 +1,170 @@
...
+    return {
+      ...getParams(host, `${parts.protocol}//${parts.host}`),
+      ...getParams(path, parts.pathname),
+      ...getQueryParams(query, parts.search),
+      ...getParams(hash, parts.hash),
+    }
+  }
+
+  return { host, path, query, hash, toString, parse }
+}
```
craig-pplx, line 94: Part of my expectation with this utility is that you could not only extract params but you could determine if a string is a match. 

How does this work if you provide a string that doesn't match the url you created?

src/types/params.ts
```diff
@@ -68,31 +67,48 @@ export type ExtractParamName<
...
+  query?: WithParams,
+  hash?: WithParams,
+}> =
   Identity<
     MakeOptional<
-      ExtractParamTypesReading<TRoute['host']> &
-      ExtractParamTypesReading<TRoute['path']> &
-      ExtractParamTypesReading<TRoute['query']> &
-      ExtractParamTypesReading<TRoute['hash']>
+      & ExtractParamTypesReading<TUrl['host'] extends WithParams ? TUrl['host'] : never>
```
craig-pplx, line 82: Why do we need the extra conditions? Don't the url parts always extend WithParams?

src/utilities/makeOptional.ts
craig-pplx: Is this type needed anymore? I don't see it used this time through.

src/services/createRoute.ts
stackoverfloweth: I think it's valuable that we offloaded all the url logic out from createRoute/createExternalRoute, which also means the params stuff will work the same for users creating random URLs as it does for routes. Also, by spreading the `url` into `route`, makes route a lot more powerful. Each url "parts" get fancy `toString(params?)` methods, and route gets `.toString(params?)` and `.parse(url)`. From the users perspective `createRoute` is almost the same API as `createUrl` except route expects `component`, `props`, things like that. That made a lot of sense to me personally.

src/services/createUrl.ts
stackoverfloweth: it throws `InvalidRouteParamValueError`. I considered adding another method like `isValid(url) => boolean` but felt meh about it. Take a look at the updated internal `routeParamsAreValid` route match rule

```ts
export const routeParamsAreValid: RouteMatchRule = (route, url) => {
  try {
    route.parse(url)
  } catch {
    return false
  }

  return true
}
```

src/types/params.ts
stackoverfloweth: I updated the generic from 

```ts
  host: WithParams,
  path: WithParams,
  query: WithParams,
  hash: WithParams,
```

to

```ts
  host?: WithParams,
  path?: WithParams,
  query?: WithParams,
  hash?: WithParams,
```

to support the `toString(params?)` on each url part. This means the types for `host` can use `ExtractParamTypesReading<{ host: THost }>` and `path` can use `ExtractParamTypesReading<{ path: TPath }>`

src/types/params.ts
stackoverfloweth: I was able to get this mostly back to how it was before

src/types/resolved.ts
stackoverfloweth: this should be back

src/types/route.ts
stackoverfloweth: I don't think so, this also feels like a better UX

src/services/createRoute.ts
stackoverfloweth: not sure about this one, we check duplicate params on individual `combinePath`, `combineQuery`, etc. We could do an overload but I feel like it's working fine and pretty straight forward for an internal service

src/types/createRouteOptions.ts
stackoverfloweth: yep, fixed

PR #658, opened 2026-01-24
createUrl utility

stackoverfloweth:
This PR adds a new low-level utility `createUrl`. It enables users to parse full url strings into parts, join url parts into a url string, and more importantly can leverage router's powerful params syntax. 

```ts
const hasParams = createUrl({
  host: withParams('https://[?subdomain].kitbag.dev', { subdomain: withDefault(String, 'abc') }),
  path: '/[id]',
  query: 'color=[color]',
  hash: withParams('#[?page]', { page: Number }),
})

// can call toString on the whole url, any params will be required
hasParams.toString({ id: '14', color: 'white', page: 1 })
// @ts-expect-error - color is required
hasParams.toString({ id: '14' })
// @ts-expect-error - page must be number
hasParams.toString({ id: '14', color: 'white', page: 'ABC' })
//  https://kitbag.dev/14?color=red#123

// can call parse to get the params from the url
const params = hasParams.parse('https://kitbag.dev/14?color=red#123')
//  { page?: number | undefined; subdomain: string; id: string; color: string; }

// can combine urls
const combined = combineUrl(hasParams, { path: withParams('/[id2]', { id2: Number }), query: 'color=blue' })
//  https://kitbag.dev/[id2]?color=blue#[page]

combined.toString({ id: '14', id2: 14, color: 'white' })
// @ts-expect-error - id2 is required
combined.toString({ id: '14', page: '123', color: 'white' })
```

### DIFF
--- a/src/services/createUrl.ts
+++ b/src/services/createUrl.ts
@@ -10,16 +10,7 @@ import { checkDuplicateParams } from '@/utilities/checkDuplicateParams'
 const FALLBACK_HOST = 'https://internal.invalid'
 
 export function createUrl<const T extends CreateUrlOptions>(options: T): ToUrl<T>
-export function createUrl(url: string): Url
-export function createUrl(urlOrOptions: CreateUrlOptions | string): Url {
-  if (typeof urlOrOptions === 'string') {
-    const url = new URL(urlOrOptions, FALLBACK_HOST)
-    const includesHost = urlOrOptions.startsWith('http')
-    const host = includesHost ? `${url.protocol}//${url.host}` : undefined
-
-    return createUrl({ host, path: url.pathname, query: url.search, hash: url.hash })
-  }
-
+export function createUrl(urlOrOptions: CreateUrlOptions): Url {
   const options = {
     host: toWithParams(urlOrOptions.host),
     path: toWithParams(urlOrOptions.path),
@@ -85,8 +76,8 @@ export function createUrl(urlOrOptions: CreateUrlOptions | string): Url {
   function tryParse(url: string): { success: true, params: Record<string, unknown> } | { success: false, error: Error } {
     try {
       return { success: true, params: parse(url) }
-    } catch (error) {
-      return { success: false, error: error as Error }
+    } catch (cause) {
+      return { success: false, error: new Error('Failed to parse url', { cause }) }
     }
   }
 


src/services/createUrl.ts
```diff
@@ -0,0 +1,200 @@
...
+import { CreateUrlOptions, ToUrl, Url } from '@/types/url'
+import { asUrlString, UrlString } from '@/types/urlString'
+import { checkDuplicateParams } from '@/utilities/checkDuplicateParams'
+
+// https://en.wikipedia.org/wiki/.invalid
+const FALLBACK_HOST = 'https://internal.invalid'
+
+export function createUrl<const T extends CreateUrlOptions>(options: T): ToUrl<T>
+export function createUrl(url: string): Url
+export function createUrl(url: string | CreateUrlOptions): Url
```
pleek91: Do we need these overloads? We don't have param support for just a url string yet. So maybe we omit these for now?

src/types/url.ts
```diff
@@ -0,0 +1,108 @@
...
+  & ToUrlParams<ToWithParams<TOptions['host']>>
+  & ToUrlParams<ToWithParams<TOptions['path']>>
+  & ToUrlParams<ToWithParams<TOptions['query']>>
+  & ToUrlParams<ToWithParams<TOptions['hash']>>
+>
+
+type UrlParams = Record<string, [param: Param, isOptional: boolean]>
+
+type ToUrlParams<TWithParams extends WithParams> = {
+  [K in keyof TWithParams['params']]: TWithParams['value'] extends `${string}${ParamStart}?${K & string}${ParamEnd}${string}`
```
pleek91: Would it make sense to make a type for the optional param string? So we can do things like
```ts
TWithParams['value'] extends OptionalParamTemplate<K>
  ? ...
  : ..
```

src/types/url.ts
```diff
@@ -0,0 +1,108 @@
...
+* @returns A record of parameter names to their respective types, extracted and merged from both path and query parameters.
+*/
+export type UrlParamsReading<TUrl extends Url> = ToUrlParamsReading<TUrl['_params']>
+
+type ToUrlParamsReading<
+  TParams extends UrlParams
+> =
+Identity<
+  MakeOptional<{
+    [K in keyof TParams]: TParams[K] extends [infer TParam extends Param, true]
```
pleek91: Similar for these, should we make a type like this? Idk if this would even work 
```ts
TParams[K] extends OptionalParam<infer TParam>
//or

TParams[K] extends RequiredParam<infer TParam>
```

src/services/createUrl.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  const internal = {
+    _schema: { host, path, query, hash },
+    _params: {},
+  }
+
+  return { ...internal, stringify, parse, match }
+}
+
+function assembleParamValues(part: WithParams, paramValues: Record<string, unknown>): string {
```
pleek91, line 111: I know these utilities are only used in this file, but should we put them in separate files, export them, and unit test them? Or do we consider these part of the internal logic of CreateUrl and we should just test that as a whole (fine with either i think).

src/services/createUrl.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  function match(_url: string): number {
+    return 0
+  }
+
+  const internal = {
+    _schema: { host, path, query, hash },
+    _params: {},
+  }
```
pleek91, line 101: We haven't used `_` prefixed "private" variable names anywhere else I don't think. So Idk that I'd start here. 

I think potentially `schema` could be left off the types and we could type guard them back similar to what we did with the `redirect` hook on Route. 

For params, I know we have to have it to hold the generic. But I'd still just name it normal and mark it as internal. Then add a jsdoc stating that it has no runtime value.

src/services/createUrl.ts
stackoverfloweth: not sure what you mean? I think url _does_ support params in this PR

src/services/createUrl.ts
stackoverfloweth: yes that's my plan, I just want to finish this body for work first so I can be certain I removed all external access to these utilities

src/services/createUrl.ts
pleek91: There's no types though? The url string isn't generic

src/services/createUrl.ts
```diff
@@ -0,0 +1,185 @@
...
+    query: toWithParams(urlOrOptions.query),
+    hash: toWithParams(urlOrOptions.hash),
+  }
+
+  checkDuplicateParams(options.path.params, options.query.params, options.host.params, options.hash.params)
+
+  const host = {
+    ...options.host,
+    regexp: replaceParamSyntaxWithCatchAllsAndEscapeRest(options.host.value),
+    toString(params: Record<string, unknown> = {}): string {
```
pleek91: Any objection to using stringify for these internal pieces as well to be consistent?

src/services/createUrl.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    return {
+      ...getParams(host, `${parts.protocol}//${parts.host}`),
+      ...getParams(path, parts.pathname),
+      ...getQueryParams(query, parts.search),
+      ...getParams(hash, parts.hash),
+    }
+  }
+
+  function tryParse(url: string): { success: true, params: Record<string, unknown> } | { success: false, error: Error } {
```
pleek91, line 76: What is the error? Is it anything useful?

src/types/params.ts
```diff
@@ -50,6 +50,16 @@ export function isLiteralParam(value: Param): value is LiteralParam {
   return typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean'
 }
 
+/**
+ * Determines if a given value is an optional parameter template.
+ * @template TKey - The key of the parameter.
+ * @template TValue - The value of the parameter.
```
pleek91, line 56: nit: value is kinda non descriptive. I know that's what it's called in WithParams but it's really the template string right? Maybe we should consider changing the naming.

src/types/url.ts
```diff
@@ -0,0 +1,117 @@
...
+  & ToUrlParams<ToWithParams<TOptions['query']>>
+  & ToUrlParams<ToWithParams<TOptions['hash']>>
+>
+
+type OptionalParam<TParam extends Param = Param> = [param: TParam, isOptional: true]
+type RequiredParam<TParam extends Param = Param> = [param: TParam, isOptional: false]
+type UrlParams = Record<string, OptionalParam | RequiredParam>
+
+type ToUrlParams<TWithParams extends WithParams> = {
+  [K in keyof TWithParams['params']]: [TWithParams['params'][K], IsOptionalParamTemplate<K & string, TWithParams['value']>]
```
pleek91, line 33: Does this account for params with defaults? Idk how that works.

src/types/url.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * Type guard to assert that a url has a schema.
+ * @internal
+ */
+export function isUrlWithSchema(url: unknown): url is Url & { schema: Record<string, WithParams> } {
+  return typeof url === 'object' && url !== null && 'schema' in url
+}
```
pleek91, line 42: We might want to use a symbol or something more specific for these sorts of things. For future us

src/types/url.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Represents the structure of a url parts. Can be used to create a url with support for params.
+ */
+export type Url<TParams extends UrlParams = UrlParams> = {
+  /**
+   * @internal
+   * The parameters type for the url.
```
pleek91:
```suggestion
   * The parameters type for the url. Non functional and undefined at runtime. 
```

src/types/url.ts
```diff
@@ -0,0 +1,117 @@
...
+  parse(url: string): ToUrlParamsReading<TParams>,
+  /**
+   * Parses the url supplied and returns any params found.
+   */
+  tryParse(url: string): { success: true, params: ToUrlParamsReading<TParams> } | { success: false, error: Error },
+  /**
+   * @internal
+   * Checks if the supplied url matches this url. Any value above 0 is a match. Can be used to compare to other partial matches. Max score is 100.
+   */
+  match(url: string): { score: number, params: ToUrlParamsReading<TParams> },
```
pleek91, line 69: If match is 0 what is params? Just an empty object? I wonder if we want to type this more similar to tryParse with a bool for matched.

src/services/createUrl.ts
stackoverfloweth: currently, it's whatever error gets thrown. In the future it probably makes sense to coalesce this into a safe made-for-user error. 

For example, if you try calling parse and provide a url that omits or otherwise incorrectly passes a param value
Screenshot 2026-01-25 at 17 13 55@2x

src/types/url.ts
stackoverfloweth: there's nothing specific here to handle defaults, which is fine - not relevant at this point. We need to know if the user defined the param as optional `[?foo]` vs `[bar]` and capture whatever the `Param` is without modifying it. If the param is `ParamGetSet` and assigned `defaultValue` (or if user used `withDefault()`) that is still captured in the param and will be relevant when it comes to differentiating reading/writing types later in this file

stackoverfloweth: FTR, even if a user defines a default value the param is considered **required**. The only way to make a param optional is to use the `[?mark]` syntax.

src/types/params.ts
stackoverfloweth: I agree, but to only change here would be worse IMO. Better to be more consistent than more correct

src/types/url.ts
stackoverfloweth: Very possible, but at this point idk how important the type is. Technically speaking this is theoretically possible

- `score: 0`: all params are actually satisfied
- `score: 100`: all params are missing (because they were optional)

We also score a route higher if it supplied more optional params than another route, so you could theoretically expect these results for the same route

- `score: 80`: 5 params provided, 3 optional params omitted
- `score: 85`: 6 params provided, 2 optional params omitted

My gut says as long as this is internal, fancy types aren't going to pay off here. Even after running route matching logic and then calling a separate utility to set `params: getRouteParamValues(route, url)`, that utility returns `Record<string, unknown>`.

src/services/createUrl.ts
stackoverfloweth: sorry, I'm not sure if this comment was intended for a specific overload? Yes we could remove line 14

src/types/params.ts
pleek91: Agreed, future us problem.

src/services/createUrl.ts
pleek91: I don't think that error is intended for users. It's just part of how we do matching. 

For now I'd prefer to just return { matches: false }

src/services/createUrl.ts
pleek91: 13 and 14. Both overloads that accept string. Because we're not capturing the string and we're not returning a typed url just a generic one with no param types.

src/types/url.ts
pleek91: score 0 can mean all params are satisfied?

stackoverfloweth: sure, if it's just totally the wrong path, but it happens to have all the same params

src/services/createUrl.ts
stackoverfloweth: agreed, we really need to add something here though. Users are going to be easily frustrated if we don't give some sort of details

pleek91: I could see adding a reason or something about why it didn't match. But I think for now a boolean is fine.

src/types/url.ts
pleek91: But if the path is wrong then params shouldn't move returned right? I'd expect a 0 score to not return any params because it's not a match.

src/services/createUrl.ts
stackoverfloweth: yeah I see your point, when I first added these overloads this utility had more value than JUST for situations where you have params.  Maybe in a future PR we can play around with supporting creating a URL without specifying parts AND support params.

```ts
const url = createUrl('https://[subdomain].kitbag.dev/foo/[id]?color=[?color]#size=[size]')
```

pleek91: Yeah exactly. I think it's inevitable that the syntax goes that direction. We just don't have it quite yet.

src/types/url.ts
stackoverfloweth: if the params are in the path, yes. The point was params isn't strictly tied to match value

src/services/createUrl.ts
stackoverfloweth: I've added a note for me to come back and try to add something more user friendly here

PR #665, opened 2026-01-30
Route match scoring

stackoverfloweth:
## Primary Quest

Simpler and more performant logic for route matching! 

This also means we're ditching the route scoring, which was previously used to sort all potential matches. Previously it would prioritize
- depth (more below)
- host matches (more below)
- hash matches
- optional params satisfied

The goal was to try to find the best possible match based on all routes, now it's less magical - it simply checks for regex pattern and checks that params (if any) are satisfied. **So after this PR route order might matter more than it did before.**

### Depth

We still prioritize depth by pre-sorting routes by `depth` descending and then instead of searching all routes and checking each routes regex, we find the 1st one that passes and exit early.

### Host

When the router is looking for matches, it first checks `isExternal(url)`. If the url is external it only checks external routes, else it only checks internal routes.

## Side Quests

- update urlParser (takes a url string, returns the parts)
  - update parts to be query/path instead of search/pathnam
  - combine protocol and host
  - add `stringifyUrl` (takes the parts, returns a string)
  - add `updateUrl` (takes pair of string|parts, joins together)
- combine nearly duplicate services `createResolvedRoute` and `createResolvedRouteForUrl`
  - allowed me to move `getMatchForUrl` out to router, where I have access to `isExternal`

### DIFF
--- a/src/services/urlParser.spec.ts
+++ b/src/services/urlParser.spec.ts
@@ -81,10 +81,10 @@ describe('updateUrl', () => {
   })
 
   test.each([
-    [{ query: new URLSearchParams('?foo=456') }, { query: new URLSearchParams('?bar=123') }],
-    [{ query: new URLSearchParams('?foo=456') }, { query: '?bar=123' }],
-    [{ query: new URLSearchParams('?foo=456&&bar=123') }, { query: '' }],
-    [{ query: new URLSearchParams('?foo=456&&bar=123') }, { query: undefined }],
+    [{ query: new URLSearchParams('?bar=123') }, { query: new URLSearchParams('?foo=456') }],
+    [{ query: new URLSearchParams('?bar=123') }, { query: '?foo=456' }],
+    [{ query: new URLSearchParams('?foo=456&bar=123') }, { query: '' }],
+    [{ query: new URLSearchParams('?foo=456&bar=123') }, { query: undefined }],
   ])('given previous query (%s) and updated query (%s), returns updated query', (previous, updated) => {
     const url = updateUrl(previous, updated)
 
--- a/src/services/urlParser.ts
+++ b/src/services/urlParser.ts
@@ -50,7 +50,7 @@ export function updateUrl(url: string | Partial<UrlParts>, updates: UrlPartsInpu
   return {
     host: stringHasValue(updates.host) ? updates.host : url.host,
     path: stringHasValue(updates.path) ? updates.path : url.path ?? '',
-    query: combineUrlSearchParams(url.query, updatedQuery),
+    query: combineUrlSearchParams(updatedQuery, url.query),
     hash: stringHasValue(updates.hash) ? updates.hash : url.hash ?? '',
   }
 }


src/services/getMatchesForUrl.ts
```diff
@@ -14,10 +13,8 @@ const rules: RouteMatchRule[] = [
...
 ]
 
-export function getMatchesForUrl(routes: Routes, url: string): Route[] {
-  const sortByRouteScore = getRouteScoreSortMethod(url)
-
+export function getMatchForUrl(routes: Routes, url: string, isExternal: boolean = false): Route | undefined {
   return routes
-    .filter((route) => rules.every((test) => test(route, url)))
-    .sort(sortByRouteScore)
+    .filter((route) => !!route.host.value === isExternal)
```
stackoverfloweth, line 18: this is how I fixed the is external thread we had @pleek91

src/services/getRoutesForRouter.ts
```diff
@@ -67,7 +67,7 @@ export function getRoutesForRouter(routes: Routes | Routes[], plugins: RouterPlu
   }
 
   return {
-    routes: Array.from(routerRoutes.values()),
+    routes: Array.from(routerRoutes.values()).sort(sortByDepthDescending),
```
stackoverfloweth, line 70: this is important, this is how we effectively keep the prefer higher depth logic

src/services/urlCreator.spec.ts
stackoverfloweth, line 1: this is now handled by `stringifyUrl` in urlParser.ts

src/services/urlCombine.ts
stackoverfloweth, line 1: this is now handled by `updateUrl` in urlParser.ts

src/services/createIsExternal.spec.ts
pleek91, line 1: Good sign, these tests probably shouldn't have passed before since protocol is part of what we consider a host

src/services/routeMatchRules.ts
pleek91, line 1: ❤️

src/services/createResolvedRoute.ts
```diff
@@ -1,31 +1,40 @@
...
 
-export function createResolvedRoute(route: Route, params: Record<string, unknown> = {}, options: RouterResolveOptions = {}): ResolvedRoute {
-  const href = assembleUrl(route, {
-    params: params,
-    query: options.query,
-    hash: options.hash,
-  })
+export function createResolvedRoute(route: Route, url: UrlString, options?: RouterResolveOptions): ResolvedRoute
+export function createResolvedRoute(route: Route, params?: Record<string, unknown>, options?: RouterResolveOptions): ResolvedRoute
+export function createResolvedRoute(route: Route, ulrOrParams: UrlString | Record<string, unknown> = {}, options: RouterResolveOptions = {}): ResolvedRoute {
```
pleek91, line 13: I couldn't figure out what the first overload was for where its `route, url, options`. What is that for? Seems odd

stackoverfloweth: That's used by router.find. This is the overload that enabled me to delete the nearly identical `createResolvedRouteForUrl`

pleek91: Ah right. Do you think maybe it would read more nicely to keep them both but have createResolvedRoute just call/wrap createResolvedRouteForUrl? The overload is a bit confusing IMO.

stackoverfloweth: Idk, overloads feels better to me personally. There are other examples of similar decisions in this project where we used overloads instead of separate files/functions.

pleek91 (approved):

PR #666, opened 2026-01-31
Adds option to remove trailing slashes from URLs

pleek91:
# Description
Adds a new router option, `removeTrailingSlashes`. Defaults to `true`. 

When `true` when doing route matching any trailing slashes in the urls path will be trimmed. 

- `/foo/` becomes `/foo`
- `/foo/?greeting=hello` becomes `/foo?greeting=hello`


stackoverfloweth (approved):
Love this. Might want to update documentation? Might warrant something outside of generated api docs

PR #667, opened 2026-02-01
Add support for lazy vs eager params and change default behavior

pleek91:
# Description

This PR introduces two new concepts to params. Lazy vs Eager params

## Today: All Params Are "Eager"

Today, every path param matches **across slashes**. A single param can span multiple path segments.

Example: path `/foo/[param]/bar`

- Matches `/foo/baz/bar` → param = `baz`
- Matches `/foo/baz/buz/bar` → param = `baz/buz`
- Matches `/foo/a/b/c/bar` → param = `a/b/c`

So in practice, all params behave like "eager": they match until the next literal segment, not just until the next `/`.

## Introducing: Lazy vs Eager Params

We're introducing an explicit **lazy vs eager** distinction:

- **Lazy (new default):** param matches only **one path segment** (stops at the next `/`).
- **Eager (opt-in):** param matches **one or more segments** (can include `/`), same as today.

**Lazy will be the new default.** Existing params that don't use the new syntax will become lazy and may change which URLs match (see migration below).

### Lazy params (default): `[param]` / `[?param]`

| Route path | URL | Match? | Param value(s) |
|------------|-----|--------|----------------|
| `/foo/[id]/bar` | `/foo/123/bar` | Yes | `id` = `"123"` |
| `/foo/[id]/bar` | `/foo/123/456/bar` | No | — |
| `/foo/[id]/bar` | `/foo/a/b/c/bar` | No | — |
| `/foo/[?id]/bar` | `/foo/bar` | Yes | `id` = `undefined` |
| `/foo/[?id]/bar` | `/foo/123/bar` | Yes | `id` = `"123"` |
| `/foo/[?id]/bar` | `/foo/123/456/bar` | No | — |

### Eager params (opt-in): `[param*]` / `[?param*]`

| Route path | URL | Match? | Param value(s) |
|------------|-----|--------|----------------|
| `/foo/[rest*]/bar` | `/foo/123/bar` | Yes | `rest` = `"123"` |
| `/foo/[rest*]/bar` | `/foo/123/456/bar` | Yes | `rest` = `"123/456"` |
| `/foo/[rest*]/bar` | `/foo/a/b/c/bar` | Yes | `rest` = `"a/b/c"` |
| `/foo/[rest*]/bar` | `/foo/bar` | No | — |
| `/foo/[?rest*]/bar` | `/foo/bar` | Yes | `rest` = `undefined` |
| `/foo/[?rest*]/bar` | `/foo/a/b/c/bar` | Yes | `rest` = `"a/b/c"` |

## Migration

If a route today relies on a param spanning multiple segments (e.g. `/foo/[param]/bar` matching `/foo/a/b/c/bar`), that param will **stop** matching after this change unless you make it eager by adding `*`:

- Change `[param]` → `[param*]`
- Change `[?param]` → `[?param*]`

Use eager params when the value can contain slashes (file paths, multi-segment IDs, etc.).

### DIFF
--- a/src/services/routeRegex.ts
+++ b/src/services/routeRegex.ts
@@ -1,11 +1,11 @@
 import { paramEnd, paramStart } from '@/types/params'
 import { Route } from '@/types/route'
 import { stringHasValue } from '@/utilities/guards'
-import { WithParams } from '@/services/withParams'
 
-export const paramRegex = `\\${paramStart}\\??([\\w-_]+)(\\*?)\\${paramEnd}`
+export const paramRegex = `\\${paramStart}\\??([\\w-_]+)\\*?\\${paramEnd}`
 export const optionalParamRegex = `\\${paramStart}\\?([\\w-_]+)\\*?\\${paramEnd}`
 export const requiredParamRegex = `\\${paramStart}([\\w-_]+)\\*?\\${paramEnd}`
+export const eagerParamRegex = `\\${paramStart}\\??([\\w-_]+)\\*\\${paramEnd}`
 export const regexCatchAll = '[^/]*'
 export const regexEagerCatchAll = '.*'
 export const regexCaptureAll = '([^/]*)'
@@ -88,22 +88,28 @@ export function replaceParamSyntaxWithCatchAllsAndEscapeRest(value: string): str
 }
 
 export function replaceParamSyntaxWithCatchAlls(value: string): string {
-  return value.replace(new RegExp(paramRegex, 'g'), (_match, _name, eagerStar) => {
-    return eagerStar ? regexEagerCatchAll : regexCatchAll
+  return value.replace(new RegExp(paramRegex, 'g'), (match) => {
+    return isEagerParamSyntax(match) ? regexEagerCatchAll : regexCatchAll
   })
 }
 
-export function replaceIndividualParamWithCaptureGroup(path: WithParams, paramName: string): string {
+export function replaceIndividualParamWithCaptureGroup(path: string, paramName: string): string {
   const pattern = getParamRegexPattern(paramName)
   const capturePattern = paramIsEager(path, paramName) ? regexEagerCaptureAll : regexCaptureAll
 
-  return path.value.replace(pattern, capturePattern)
+  return path.replace(pattern, capturePattern)
 }
 
-export function paramIsOptional(path: WithParams, paramName: string): boolean {
+export function paramIsOptional(path: string, paramName: string): boolean {
   const paramRegex = getOptionalParamRegexPattern(paramName)
 
-  return paramRegex.test(path.value)
+  return paramRegex.test(path)
+}
+
+export function paramIsEager(path: string, paramName: string): boolean {
+  const eagerPattern = getEagerParamRegexPattern(paramName)
+
+  return eagerPattern.test(path)
 }
 
 export function isOptionalParamSyntax(value: string): boolean {
@@ -112,6 +118,10 @@ export function isOptionalParamSyntax(value: string): boolean {
 
 export function isRequiredParamSyntax(value: string): boolean {
   return new RegExp(requiredParamRegex, 'g').test(value)
+}
+
+export function isEagerParamSyntax(value: string): boolean {
+  return new RegExp(eagerParamRegex, 'g').test(value)
 }
 
 export function getParamName(value: string): string | undefined {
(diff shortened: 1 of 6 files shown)


docs/core-concepts/params.md
```diff
@@ -28,6 +28,21 @@ const events = createRoute({
 })
 ```
 
+## Eager Params
```
stackoverfloweth, line 31: We should also consider using the alternative "Greedy Params". I probably would have used greedy but I don't feel strongly.

docs/core-concepts/params.md
```diff
@@ -28,6 +28,21 @@ const events = createRoute({
...
+By default, a path param matches only up to the next `/` (a single path segment). For example, the path `/foo/[param]/bar` matches `/foo/baz/bar` (param is `baz`) but not `/foo/baz/buz/bar`.
+
+An **eager param** matches across one or more path segments (including `/`). Add a `*` after the param name: `[param*]` or `[?param*]`. The path `/foo/[param*]/bar` then matches both `/foo/baz/bar` (param is `baz`) and `/foo/baz/buz/bar` (param is `baz/buz`).
+
+```ts {3}
+const route = createRoute({
+  name: 'support',
+  path: '/supports/[slashes*]/bookmarked',
+})
+```
```
stackoverfloweth: I like that our docs have mostly used realistic examples

```suggestion
By default, a path param matches only up to the next `/` (a single path segment). For example, the path `/photos/[date]` matches `/photos/2026` (param is `"2026"`) but not `/photos/2026/01`.

An **eager param** matches across one or more path segments (including `/`). Add a `*` after the param name: `[param*]` or `[?param*]`. The path `/photos/[date*]` then matches both `/photos/2026` (param is `"2026"`) and `/photos/2026/01` (param is `"2026/01`).

```ts {3}
const route = createRoute({
  name: 'photos',
  path: '/photos/[date*]',
})
```
```

Comment (stackoverfloweth) [reply docs/core-concepts/params.md]: alternatively if you feel we need the example to show a bookend


```suggestion
By default, a path param matches only up to the next `/` (a single path segment). For example, the path `/photos/[date]/search` matches `/photos/2026/upload` (param is `"2026"`) but not `/photos/2026/01/upload`.

An **eager param** matches across one or more path segments (including `/`). Add a `*` after the param name: `[param*]` or `[?param*]`. The path `/photos/[date*]/upload` then matches both `/photos/2026/upload` (param is `"2026"`) and `/photos/2026/01/upload` (param is `"2026/01`).

```ts {3}
const route = createRoute({
  name: 'upload-photos',
  path: '/photos/[date*]/upload',
})
```
```

File: src/services/routeRegex.ts [hunk 3]
```diff
@@ -80,13 +88,16 @@ export function replaceParamSyntaxWithCatchAllsAndEscapeRest(value: string): str
...
 export function replaceParamSyntaxWithCatchAlls(value: string): string {
-  return value.replace(new RegExp(paramRegex, 'g'), regexCatchAll)
+  return value.replace(new RegExp(paramRegex, 'g'), (_match, _name, eagerStar) => {
+    return eagerStar ? regexEagerCatchAll : regexCatchAll
+  })
 }
 
-export function replaceIndividualParamWithCaptureGroup(path: string, paramName: string): string {
-  const paramRegex = getParamRegexPattern(paramName)
+export function replaceIndividualParamWithCaptureGroup(path: WithParams, paramName: string): string {
```
stackoverfloweth: why does this need to be `WithParams` vs `string`?

src/services/routeRegex.ts
```diff
@@ -3,6 +3,14 @@ import { Route } from '@/types/route'
 import { stringHasValue } from '@/utilities/guards'
 import { WithParams } from '@/services/withParams'
 
+export const paramRegex = `\\${paramStart}\\??([\\w-_]+)(\\*?)\\${paramEnd}`
```
stackoverfloweth: I noticed that this pattern added a 2nd capture group. Did you try keeping the capture as is and just testing the param name against your `eagerParamRegex`>

src/services/routeRegex.ts
stackoverfloweth: [this is addressed in my suggestion PR]

src/services/routeRegex.ts
stackoverfloweth: [this is addressed in my suggestion PR]

PR #668, opened 2026-02-01
PR review suggestions

stackoverfloweth:


src/types/params.ts
stackoverfloweth, line 1: I didn't actually change anything here, this was lint. Not sure if you have linter running in your IDE

PR #669, opened 2026-02-01
Update route to utilize new url service under the hood

stackoverfloweth:
In [a recent PR](https://github.com/kitbagjs/router/pull/658), we added a great new service `createUrl`. This service is meant to encapsulate all of the responsibility around what are the parts of a URL, how do you find + interpolate params, and how do you know if a url "matches". 

This PR is the culmination of this work by stripping out all of that from `route` and utilizing `createUrl` under the hood of `createRoute`.

This SHOULD greatly reduce complexity and keep 99.9% of the app from ever reaching into route.path, route.query, etc. The only exceptions as of right now are 
- combineUrl
- devtools

This PR also attempts to improve performance of route matching by generating all of the regex when we create the URL and then using it internally when `route.match` is called.  It also returns params since that's required to determine if a url actually matches, so now we're using those params. By using the returned params to create the resolved route, we avoid the previous paradigm of having to go back a 2nd time to find the param value and run param getters.


src/services/createRouter.ts
```diff
@@ -27,7 +27,7 @@ import { EmptyRouterPlugin, RouterPlugin } from '@/types/routerPlugin'
 import { getRoutesForRouter } from './getRoutesForRouter'
 import { getGlobalHooksForRouter } from './getGlobalHooksForRouter'
 import { createComponentsStore } from './createComponentsStore'
-import { initZod, zotParamsDetected } from './zod'
+import { initZod, zodParamsDetected } from './zod'
```
stackoverfloweth, line 30: 😉

src/services/createUrl.ts
```diff
@@ -22,53 +21,52 @@ export function createUrl(urlOrOptions: CreateUrlOptions): Url {
...
+      host: host.stringify(params),
+      path: path.stringify(params),
+      query: query.stringify(params),
+      hash: hash.stringify(params),
+    })
   }
 
   function parse(url: string): Record<string, unknown> {
-    const parts = new URL(url, FALLBACK_HOST)
+    const parts = parseUrl(url)
```
stackoverfloweth, line 64: this was low hanging fruit

src/services/createUrl.ts
```diff
@@ -183,16 +213,3 @@ function getQueryParams(query: WithParams, url: string): Record<string, unknown>
   }
   return values
 }
-
-function generateQueryRegexPatterns(value: string): RegExp[] {
```
stackoverfloweth, line 187: this was accidentally duplicated from routeRegex. I think that's a better spot for this logic

src/services/createExternalRoute.spec.ts
```diff
@@ -14,12 +14,7 @@ test('given parent, path is combined', () => {
     path: withParams('/child/[id]', { id: Number }),
   })
 
-  expect(child.path).toMatchObject({
-    value: '/parent/child/[id]',
-    params: {
-      id: Number,
-    },
-  })
+  expect(child.stringify({ id: 123 })).toBe('https://kitbag.dev/parent/child/123')
```
pleek91, line 17: Should these tests check the schema instead?

src/services/createResolvedRoute.browser.spec.ts
```diff
@@ -7,10 +7,11 @@ test('given a url with a query returns all query values', () => {
   const route = createRoute({
     name: 'route',
     path: '/',
+    query: '?foo=foo1&foo=foo2&bar=bar&baz',
     component,
   })
 
-  const response = createResolvedRoute(route, '/?foo=foo1&foo=foo2&bar=bar&baz')
+  const response = createResolvedRoute(route)
```
pleek91: Is this test still accurate? It was testing that the extra query from the browsers url appeared in the resolved route I think? Vs now it's not testing that it's just asserting they existed on the route itself.

src/services/createResolvedRoute.browser.spec.ts
```diff
@@ -61,9 +62,10 @@ test('given a url with hash, returns hash property', () => {
     name: 'route',
     path: '/foo',
     component,
+    hash: '#bar',
   })
 
-  const response = createResolvedRoute(route, '/foo#bar')
+  const response = createResolvedRoute(route)
```
pleek91: Same here about hash. I don't think this test is accurate now.

src/services/createRoute.spec-d.ts
```diff
@@ -11,13 +10,16 @@ import { BuiltInRejectionType } from '@/services/createRouterReject'
...
 test('empty options returns an empty route', () => {
   const route = createRoute({})
 
-  type RouteWithHooks = Route & InternalRouteHooks<typeof route, []> & RouteRedirects<typeof route>
+  type Source = typeof route
 
-  expectTypeOf<typeof route>().toEqualTypeOf<RouteWithHooks>()
+  expectTypeOf<Source>().toMatchObjectType<Url>()
+  expectTypeOf<Source>().toMatchObjectType<InternalRouteHooks<Source>>()
+  expectTypeOf<Source>().toMatchObjectType<RouteRedirects>()
```
pleek91, line 22: ❤️

src/services/createRoute.spec-d.ts
```diff
@@ -44,10 +46,10 @@ test('options with name and parent', () => {
...
   const route = createRoute({ path: '/foo' })
 
-  type Source = typeof route['path']
-  type Expect = { value: '/foo', params: {} }
+  type Source = typeof route['params']
+  type Expect = {}
 
-  expectTypeOf<Source>().toEqualTypeOf<Expect>()
+  expectTypeOf<Source>().toMatchObjectType<Expect>()
 })
```
pleek91: Is this test even valuable anymore?

src/services/createRoute.spec-d.ts
```diff
@@ -57,18 +59,18 @@ test('options with path and parent', () => {
     parent,
   })
 
-  type Source = typeof route['path']
-  type Expect = { value: '/foo/bar', params: {} }
+  type Source = typeof route['params']
+  type Expect = {}
 
-  expectTypeOf<Source>().toEqualTypeOf<Expect>()
+  expectTypeOf<Source>().toMatchObjectType<Expect>()
```
pleek91: Same with this one

src/services/createRoute.spec.ts
```diff
@@ -16,12 +16,7 @@ test('given parent, path is combined', () => {
     path: withParams('/child/[id]', { id: Number }),
   })
 
-  expect(child.path).toMatchObject({
-    value: '/parent/child/[id]',
-    params: {
-      id: Number,
-    },
-  })
+  expect(child.stringify({ id: 123 })).toBe('/parent/child/123')
```
pleek91, line 19: Should these test the schema rather than the stringify output?

src/services/createRoute.ts
```diff
@@ -47,28 +48,41 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
...
     context,
     depth: 1,
-    host: withParams(),
     prefetch: options.prefetch,
     ...redirects,
+    ...url,
     ...hooks,
   } satisfies Route & InternalRouteHooks & RouteRedirects
 
-  const merged = isWithParent(options) ? combineRoutes(options.parent, route) : route
```
pleek91, line 69: ❤️

src/services/createUrl.ts
```diff
@@ -91,7 +114,14 @@ export function createUrl(urlOrOptions: CreateUrlOptions): Url {
     [IS_URL_SYMBOL]: true,
   } as const
 
-  return { ...internal, stringify, parse, tryParse, match }
+  return {
+    ...internal,
+    isRelative: !stringHasValue(options.host.value),
```
pleek91, line 109: I think this is added so the router can figure out relative vs external. 

I'd suggest instead we omit this from url (keep url agnostic of relative/external) and add an isExternal boolean in the external route type. 

Then in the router we don't have to switch from "is external" to "is relative".

src/services/getMatchesForUrl.ts
```diff
@@ -1,20 +1,22 @@
...
-const rules: RouteMatchRule[] = [
-  isNamedRoute,
-  routeHostMatches,
-  routePathMatches,
-  routeQueryMatches,
-  routeHashMatches,
-  routeParamsAreValid,
-]
+export function getMatchForUrl(routes: Routes, url: string, options: RouterResolveOptions = {}): ResolvedRoute | undefined {
+  const { query, hash } = parseUrl(updateUrl(url, options))
```
pleek91, line 9: ❤️

src/services/getMatchesForUrl.ts
```diff
@@ -1,20 +1,22 @@
...
+    if (!isNamedRoute(route)) {
+      continue
+    }
+
+    const { isMatch, params } = route.match(url)
+
+    if (isMatch) {
+      return createResolvedRoute(route, params, { ...options, query, hash })
+    }
+  }
```
pleek91, line 21: ❤️❤️❤️

src/services/insertBaseRoute.ts
```diff
@@ -1,16 +1,15 @@
...
   if (!stringHasValue(base)) {
     return route
   }
 
-  const value = `${base}${route.path.value}`
-
   return {
     ...route,
-    path: withParams(value, route.path.params),
+    ...combineUrl(createUrl({ path: base }), route),
```
pleek91, line 13: why do we have combineUrl and updateUrl? Do we need both?

src/types/createRouteOptions.ts
```diff
@@ -60,27 +59,12 @@ export function isWithComponentPropsRecord<T extends Record<string, unknown>>(op
 
 export type CreateRouteOptions<
   TName extends string | undefined = string | undefined,
-  TPath extends string | WithParams | undefined = string | WithParams | undefined,
-  TQuery extends string | WithParams | undefined = string | WithParams | undefined,
-  THash extends string | WithParams | undefined = string | WithParams | undefined,
   TMeta extends RouteMeta = RouteMeta
-> = {
+> = Omit<CreateUrlOptions, 'host'> & {
```
pleek91: I think we should just duplicate these properties here. It'll be a better devx for end users trying to read the type hints.

src/types/redirects.ts
```diff
@@ -44,7 +44,7 @@ export type RedirectHook<
 export type RouteRedirectCallback<
   TRouteTo extends Route = Route,
   TRouteFrom extends Route = Route
-> = (params: ExtractRouteParamTypesReading<TRouteFrom>) => ExtractRouteParamTypesWriting<TRouteTo>
+> = (params: UrlParamsReading<TRouteFrom>) => UrlParamsWriting<TRouteTo>
```
pleek91, line 47: ❤️❤️

src/types/url.ts
```diff
@@ -93,12 +96,12 @@ type ToUrlParamsReading<
 Identity<
   MakeOptional<{
     [K in keyof TParams]: TParams[K] extends OptionalParam<infer TParam>
-      ? TParam extends ParamWithDefault
```
pleek91, line 96: This type and the next seem unrelated to the rest. What's being fixed here?

Also never vs unknown seems like a big difference. Why the change?

src/services/createExternalRoute.spec.ts
stackoverfloweth: my goal was to avoid schema as much as possible. Ideally that could be completely removed some day. This effectively tests the same thing just _slightly_ less specific

src/services/createResolvedRoute.browser.spec.ts
stackoverfloweth: I see your point, I'll remove. Technically this does still work, just harder to test. Browser with extra query is captured in createRouter `set` method, sent to `find`/`getMatchForUrl`, which combines any query sent in with `RouterResolveOptions` with any query from `url`. IMO this is a nice simplification to lean on existing paradigm (RouterResolveOptions) to add extra stuff to the query/hash

src/services/createRoute.spec.ts
stackoverfloweth: this is a duplicate, depends on your answer [here](https://github.com/kitbagjs/router/pull/669#discussion_r2752359772)

src/services/insertBaseRoute.ts
stackoverfloweth: it's a bit confusing because we named our special thing `Url`, partly why I wanted to name it something like `DynamicUrl`. 

combineUrl -> combines to Url objects (our type)
updateUrl -> combines to string | UrlParts (built-in utility wrapper)

I'm open to suggestions

src/types/url.ts
stackoverfloweth: yeah I hoped I could use `ParamWithDefault` here because that's actually what we're checking and it reads more nicely than `Required<ParamGetSet>`.

[this test](https://github.com/kitbagjs/router/blob/main/src/types/route.spec-d.ts#L30) will fail with this change reverted. For some reason it doesn't extend `ParamWithDefault` even though it feels like the same but with a generic 

```ts
export type ParamWithDefault<TParam extends Param = Param> = Required<ParamGetSet<ExtractParamType<TParam>>>
```

stackoverfloweth: WRT to the `never`, it's actually a miss from before. This updates params for reading to match the type for params for writing when not explicit. 

[this test](https://github.com/kitbagjs/router/blob/main/src/types/resolved.spec-d.ts#L23) will fail with this change reverted. It prevents the generic `Record<string, unknown>` on router.route params without a specific route supplied.

src/services/createResolvedRoute.browser.spec.ts
pleek91: Maybe we just need to test this elsewhere

src/services/createExternalRoute.spec.ts
pleek91: I don't feel strongly. Agreed we should rely on the schema as little as possible.

src/services/insertBaseRoute.ts
pleek91: Ah interesting. Yeah they feel very similar. Fine for now 👍

PR #674, opened 2026-02-03
hotfix: insert base route logic

stackoverfloweth:
we recently pivoted out logic inside [getRoutesForRouter](https://github.com/kitbagjs/router/blob/main/src/services/getRoutesForRouter.ts) AND [insertBaseRoute](https://github.com/kitbagjs/router/blob/main/src/services/insertBaseRoute.ts)

At some point we failed to account for the fact that `insertBaseRoute` no longer modifies the route and instead returns a new route. This means that while we _are_ currently calling `insertBaseRoute`, it isn't doing anything.

Separately, we also started [removing trailing slashes](https://github.com/kitbagjs/router/pull/666) by default. So I added a hot fix inside insert base route logic to account for the "root" or "home" . Basically if the path is `path: "/"`, don't combine, just use the base. Otherwise we'd see `/my-base/` for the root route.

This might end up being our default behavior for all routes in the future

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "homepage": "https://github.com/kitbagjs/router#readme",
   "scripts": {
     "build": "vite build",
+    "build:watch": "vite build --watch --minify=false",
     "dev": "vite build --watch --minify=false",
     "test": "vitest",
     "lint": "eslint ./src",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
   "homepage": "https://github.com/kitbagjs/router#readme",
   "scripts": {
     "build": "vite build",
+    "build:watch": "vite build --watch --minify=false",
```
craig-pplx, line 20: Is this intentional?

stackoverfloweth: yes, very useful

PR #678, opened 2026-02-04
fix: router reject types

stackoverfloweth:
Our types for router hooks weren't working. As you can see in the test I added, it only ever expected the built-in rejection type.

Screenshot 2026-02-03 at 18 48 34@2x

### DIFF
--- a/src/services/createRouter.browser.spec.ts
+++ b/src/services/createRouter.browser.spec.ts
@@ -1,8 +1,9 @@
-import { mount } from '@vue/test-utils'
-import { expect, test } from 'vitest'
+import { flushPromises, mount } from '@vue/test-utils'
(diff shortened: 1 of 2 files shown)


src/services/createRouter.browser.spec.ts
```diff
@@ -31,3 +32,79 @@ test('Router is automatically started when installed', async () => {
...
+      // ok
+      push('root')
+      // @ts-expect-error does not know about rejections sent as router options
+      push('fakeRoute')
+      // ok
+      reject('NotFound')
+      // ok
+      reject('CustomRejection')
+      // @ts-expect-error should not accept an invalid rejection type
+      reject('fakeRejection')
```
pleek91, line 84: These should probably be type tests but fine for now.

PR #679, opened 2026-02-04
fix: route reject types

stackoverfloweth:
Our types for route hooks reject weren't working. As you can see in the test I added, it would fail if you tried using a rejection type defined by the parent `context`.

Screenshot 2026-02-03 at 18 52 00@2x

### DIFF
--- a/src/services/createRoute.spec.ts
+++ b/src/services/createRoute.spec.ts
@@ -107,6 +107,19 @@ test('given parent, context is combined', () => {
     },
   })
 
+  child.onAfterRouteEnter((_to, { push, reject }) => {
+    // ok
+    push('bRoute')
+    // @ts-expect-error should not accept an invalid route name
+    push('fakeRoute')
+    // ok
+    reject('aRejection')
+    // ok
+    reject('NotFound')
+    // @ts-expect-error should not accept an invalid rejection type
+    reject('fakeRejection')
+  })
+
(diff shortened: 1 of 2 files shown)


src/services/createRoute.spec.ts
```diff
@@ -107,6 +107,19 @@ test('given parent, context is combined', () => {
...
+    push('bRoute')
+    // @ts-expect-error should not accept an invalid route name
+    push('fakeRoute')
+    // ok
+    reject('aRejection')
+    // ok
+    reject('NotFound')
+    // @ts-expect-error should not accept an invalid rejection type
+    reject('fakeRejection')
+  })
```
pleek91, line 121: Can these go in the type tests?

src/types/routeContext.ts
```diff
@@ -11,8 +11,12 @@ export type ToRouteContext<TContext extends RouteContext[] | readonly RouteConte
 export type ExtractRouteContext<
   TOptions extends CreateRouteOptions
 > = TOptions extends { context: infer TContext extends RouteContext[] }
-  ? TContext
-  : []
+  ? TOptions extends { parent: infer TParent extends Route }
+    ? [...ToRouteContext<TParent['context']>, ...TContext]
+    : TContext
+  : TOptions extends { parent: infer TParent extends Route }
+    ? ToRouteContext<TParent['context']>
```
pleek91, line 18: Ah nice good catch

PR #682, opened 2026-02-05
reject should abort

stackoverfloweth:
## Expectation

When a user sets up a before hook that calls `reject` it should **NOT** continue updating the current url.

## Background

Our rejections are intended to be a layer "on top of" the underlying route. Unlike other routers, you don't need to actually define separate error routes and redirect the user. Your rejection component overrides the actual router at the `<RouterView/>` level. This also means that if the user is able to resolve the rejection (say by logging in), the `onSuccess` of that login can simply call `router.refresh()` which avoids the common pattern of stashing "where was the user trying to go" somewhere like browser storage and then checking it after successful login.

If we do not make this change, the test in this file changes to:

Screenshot 2026-02-05 at 12 16 14@2x

## What this means

If the rejection happens as part of a url change - could be before route change or on route update (maybe params changed), the route wouldn't actually be updated. So user stays where they were, rejection gets mounted, we lose the action the user previously wanted.

### DIFF
--- a/src/services/hooks.browser.spec.ts
+++ b/src/services/hooks.browser.spec.ts
@@ -300,10 +300,13 @@ test('async beforeEnter hook with reject prevents navigation after awaiting', as
     },
   })
 
+  const targetSetup = vi.fn()
+
   const target = createRoute({
     name: 'target',
     path: '/target',
     component: {
+      setup: targetSetup,
       template: '<h1>Target</h1>',
     },
   })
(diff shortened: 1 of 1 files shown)


stackoverfloweth:
users issue was rending component from `route`, which does NOT consider rejection

```html
<RouterView v-slot="{route)",
  <Transition name="slide-fade">
    <component :is="route.matched.component" />
  </Transition>
</RouterView>
```

correct way to handle transitions is [already in docs](https://router.kitbag.dev/components/router-view.html#transitions)

```html
<router-view>
  <template #default={ component }>
    <transition name="fade">
      <component :is="component" />
    </transition>
  </template>
</router-view>
```

PR #686, opened 2026-02-07
withParams refactor

stackoverfloweth:
This PR refactors our `withParams` utility (and types) to adopt a simplified output. Previously, withParams kept track of both the Params `Record<string, Param | undefined>` and the Value literal `string`. This is because the string value is what tells us if a param is optional or greedy. 

When we added the `createUrl` utility, we pivoted within that service from the WithParams approach to a tuple `[param: Param, isOptional: boolean]`. This PR adopts that tuple idea but at a higher level. It also expands that 2nd member for a single boolean to `ParamOptions`.

```ts
export type ParamOptions = {
  isOptional?: boolean,
  isGreedy?: boolean,
}
```

Not only does this make the types simpler, it also means we figure out the param options (optional/greedy) when the user calls `withParams` and then we never check the string again, so no regex checks every time we get or set param values. To help with the transition, I've also renamed the output from `WithParms` to `UrlPart`.

### DIFF
--- a/src/services/withParams.ts
+++ b/src/services/withParams.ts
@@ -24,6 +24,8 @@ type WithParamsParamsOutput<
 const UrlPartsWithParamsSymbol = Symbol('UrlPartsWithParams')
 
 export type UrlParam<TParam extends Param = Param> = { param: TParam, isOptional: boolean, isGreedy: boolean }
+export type RequiredUrlParam<TParam extends Param = Param> = { param: TParam, isOptional: false, isGreedy: false }
+export type OptionalUrlParam<TParam extends Param = Param> = { param: TParam, isOptional: true, isGreedy: false }
 export type UrlParams = Record<string, UrlParam>
 
 export type UrlPart<TParams extends UrlParams = UrlParams> = {
--- a/src/types/url.ts
+++ b/src/types/url.ts
@@ -1,9 +1,9 @@
 /* eslint-disable @typescript-eslint/method-signature-style */
-import { ToUrlPart, UrlParams, UrlPart } from '@/services/withParams'
+import { OptionalUrlParam, RequiredUrlParam, ToUrlPart, UrlParams, UrlPart } from '@/services/withParams'
 import { ExtractParamType } from '@/types/params'
 import { AllPropertiesAreOptional, Identity } from '@/types/utilities'
 import { UrlString } from '@/types/urlString'
-import { Param, ParamGetSet } from '@/types/paramTypes'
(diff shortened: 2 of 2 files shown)


src/services/createRoute.spec-d.ts
```diff
@@ -16,7 +16,7 @@ test('empty options returns an empty route', () => {
 
   type Source = typeof route
 
-  expectTypeOf<Source>().toMatchObjectType<Url>()
+  expectTypeOf<Source>().toExtend<Url>()
```
stackoverfloweth, line 19: no idea why this had to change. Worked fine with original tuple syntax but when I changed to record it broke?

src/types/url.ts
```diff
@@ -90,11 +82,11 @@ type ToUrlParamsReading<
 > =
 Identity<
   MakeOptional<{
-    [K in keyof TParams]: TParams[K] extends OptionalParam<infer TParam>
+    [K in keyof TParams]: TParams[K] extends { param: infer TParam extends Param, isOptional: true }
       ? TParam extends Required<ParamGetSet>
         ? ExtractParamType<TParam>
         : ExtractParamType<TParam> | undefined
-      : TParams[K] extends RequiredParam<infer TParam>
+      : TParams[K] extends { param: infer TParam extends Param, isOptional: false }
```
craig-pplx: The utility params seem like they would be worth keeping, or do they not work with the new syntax?

src/services/withParams.ts
```diff
@@ -56,11 +66,12 @@ export function toWithParams<T extends string | WithParams | undefined>(value: T
 export function withParams<
   const TValue extends string,
   const TParams extends MakeOptional<WithParamsParamsInput<TValue>>
->(value: TValue, params: TParams): WithParams<TValue, TParams>
-export function withParams(): WithParams<'', {}>
-export function withParams(value?: string, params?: Record<string, Param | undefined>): WithParams {
+>(value: TValue, params: TParams): WithParamsParamsOutput<TValue, TParams> extends UrlParams ? UrlPart<WithParamsParamsOutput<TValue, TParams>> : never
+export function withParams(): UrlPart<{}>
```
craig-pplx, line 66: Seems like we might want a utility here that handles the logic rather than doing infers inline with the return type.

stackoverfloweth: found a way to write without the `Identity<T>`, which is why `WithParamsParamsOutput` wasn't saying it satisfies `UrlParams`

PR #687, opened 2026-02-07
clean unused exports

stackoverfloweth:


src/compositions/useQueryValue.ts
```diff
@@ -5,7 +5,7 @@ import { Param } from '@/types/paramTypes'
 import { ExtractParamType } from '@/types/params'
 import { safeGetParamValue, setParamValue } from '@/services/params'
 
-export type UseQueryValue<T> = {
```
pleek91, line 8: Isn't this exported for users to use? If not should it be?

stackoverfloweth: it's not currently explicitly exported, only through the return type of `useQueryValue`. I feel like that probably is sufficient? For the record, this is the same for `useRouter`, `useLink`, etc. The type isn't exported

pleek91 (approved):

PR #694, opened 2026-02-09
Expand API for query inputs

stackoverfloweth:
Previously, defining the query part of a url was the exact same input as other parts like `path`. (`string | withParams | undefined'). 

With this PR we expand the available inputs to support
- Record<string, string>
- [string, string][]

Or if you want to use params
- Record<string, Param>
- [string, Param][]

```ts
createRoute({
  name: 'example',
  query: {
    static: 'value',
    dynamic: Number
   }
 }
})
```

Note: because we define optional params by the key, these new syntax options to **NOT** support optional params yet. In theory we could follow this up with something like

```ts
createRoute({
  name: 'example',
  query: {
    '?dynamic': Number
   }
 }
})
```

### DIFF
--- a/src/services/withParams.spec.ts
+++ b/src/services/withParams.spec.ts
@@ -101,6 +101,16 @@ describe('toUrlQueryPart', () => {
     })
   })
 
+  test('given a record with Param values and optional key, converts to parameterized query string', () => {
+    const response = toUrlQueryPart({ 'foo': Number, '?baz': Boolean, '?zoo': 'zoo' })
+
+    expect(response.value).toBe('foo=[foo]&baz=[?baz]&?zoo=zoo')
+    expect(response.params).toMatchObject({
+      foo: { param: Number, isOptional: false, isGreedy: false },
+      baz: { param: Boolean, isOptional: true, isGreedy: false },
+    })
+  })
+
   test('given an array with string values, converts to query string', () => {
     const response = toUrlQueryPart([['foo', 'bar'], ['baz', 'qux']])
 
@@ -115,6 +125,16 @@ describe('toUrlQueryPart', () => {
     expect(response.params).toMatchObject({
       foo: { param: Number, isOptional: false, isGreedy: false },
       baz: { param: Boolean, isOptional: false, isGreedy: false },
+    })
+  })
+
+  test('given a record with Param values and greedy key, converts to parameterized query string', () => {
+    const response = toUrlQueryPart({ 'foo': Number, '?baz': Boolean, '?zoo': 'zoo' })
+
+    expect(response.value).toBe('foo=[foo]&baz=[?baz]&?zoo=zoo')
+    expect(response.params).toMatchObject({
+      foo: { param: Number, isOptional: false, isGreedy: false },
+      baz: { param: Boolean, isOptional: true, isGreedy: false },
     })
   })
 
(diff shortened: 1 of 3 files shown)


src/services/withParams.spec.ts
```diff
@@ -27,9 +27,7 @@ test('given value with optional params, returns each param name as type String',
 })
 
 test('given value with param types, returns each param with corresponding param', () => {
-  const response = withParams('[parentId]-[childId]', {
-    parentId: Boolean,
-  })
+  const response = withParams('[parentId]-[childId]', { parentId: Boolean, childId: String })
```
pleek91: Why did this test change?

src/services/withParams.ts
```diff
@@ -71,3 +73,60 @@ export function withParams(value?: string, params?: Record<string, Param | undef
...
+ * Supports
+ * { query: 'foo=bar' }
+ * { query: 'foo=[bar]' }
+ * { query: { foo: 'bar' } }
+ * { query: [['foo', 'bar']] }
+ * { query: { foo: Param } }
+ * { query: [[ 'foo', Param ]] }
+*/
+export type QuerySourceOrUrlPart = QuerySource | UrlPart | undefined | Record<string, Param> | [string, Param][]
+export type QuerySourceToUrlPart<T extends QuerySourceOrUrlPart> = T extends string
```
pleek91: Nit: any reason we need to have "OrUrlPart" ins this? QuerySource can be many things including a UrlPart but I don't think it needs to be in the name.

src/types/paramTypes.ts
```diff
@@ -5,6 +5,11 @@ export type ParamExtras = {
   invalid: (message?: string) => never,
 }
 
+export type ParamOptions = {
+  isOptional?: boolean,
+  isGreedy?: boolean,
+}
```
pleek91: Doesn't this exist already?

src/types/utilities.ts
```diff
@@ -26,3 +26,16 @@ export type AllPropertiesAreOptional<T> = Record<string, unknown> extends T
...
+  [K in keyof T]: [K, T[K]]
+}[keyof T]
+
+export type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+
+export type LastOf<T> =
+  UnionToIntersection<T extends any ? () => T : never> extends () => infer R ? R : never
+
+export type UnionToTuple<T, L = LastOf<T>, N = [T] extends [never] ? true : false> =
```
pleek91: What were these added for? I couldn't find them used in the PR (on my phone so probably missed them)

src/services/withParams.ts
stackoverfloweth: technically `QuerySource` is a different thing. And actually, it was incorrect to use `QuerySource` in this union, since I don't technically handle `URLSearchParams` class which is part of `QuerySource`

I can see this confusion, and pushed a rename that I think actually makes this a bit more congruent with other url parts

Screenshot 2026-02-09 at 14 15 46@2x

src/types/paramTypes.ts
stackoverfloweth: yeah this looks like a bad rebase, will remove

src/types/utilities.ts
stackoverfloweth: bad rebase, good catch

src/services/withParams.spec.ts
stackoverfloweth: 🤷 reverted

PR #699, opened 2026-02-13
Default params always optional

stackoverfloweth:
I've tripped over this multiple times and I'm not sure why we didn't just build it this way from the beginning.

Currently, a param with a default value is still considered **required** when stringifying

```ts
const url = createUrl({
  path: withParams('/[foo]', { foo: withDefault(Number, 42) }),
})

url.stringify()
//  ^ TS ERROR ❌
```

This PR updates that logic so that a parm with a default is always considered to be optional.

```ts
const url = createUrl({
  path: withParams('/[foo]', { foo: withDefault(Number, 42) }),
})

url.stringify()
//  ^ ok ✅
```

also works nicely for new query param syntax

```ts
const url = createUrl({
  query: {
    foo: withDefault(String, 'abc'),
  },
})

const response = url.stringify()
//  ^ ok ✅
```

### DIFF
--- a/src/services/createUrl.spec.ts
+++ b/src/services/createUrl.spec.ts
@@ -182,7 +182,7 @@ describe('url assembly', () => {
     test.each([
       ['/simple/[?simple]'],
       [withParams('/simple/[?simple]', { simple: String })],
-      [withParams('/simple/[?simple]', { simple: withDefault(String, 'abc') })],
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
I could have sworn this is how it worked all along. Did we lose this when we did the url refactor maybe?

PR #701, opened 2026-02-15
update docs for some recent changes

stackoverfloweth:
updates docs for recent changes
- route matching logic
- adds small item about createUrl
- mentions new syntax options for defining query, uses in several places instead of strings

### DIFF
--- a/docs/advanced-concepts/route-matching.md
+++ b/docs/advanced-concepts/route-matching.md
@@ -12,7 +12,7 @@ If a route is defined as [external](/core-concepts/external-routes), it will onl
 
 ## Depth
 
-A route’s depth is determined by the number of parent routes it has. The deeper the route, the higher the priority it has when matching. Kitbag Router pre‑sorts routes by depth in descending order and then returns the first route that matches all rules.
+A route’s depth is determined by the number of parent routes it has. The deeper the route, the higher the priority it has when matching.
 
 ```ts
 const external = createExternalRoute({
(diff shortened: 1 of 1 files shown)


docs/advanced-concepts/route-matching.md
```diff
@@ -6,15 +6,35 @@ There are several rules Kitbag Router uses to determine which of your routes cor
 
 Routes without a [name](/core-concepts/routes#name) property cannot be matched.
 
+## External Routes
+
+If a route is defined as [external](/core-concepts/external-routes), it will only be matched when the url is also external.
+
+## Depth
+
+A route’s depth is determined by the number of parent routes it has. The deeper the route, the higher the priority it has when matching. Kitbag Router pre‑sorts routes by depth in descending order and then returns the first route that matches all rules.
```
pleek91: The presort piece is an implementation detail. Prob not something we should include in the docs imo

docs/advanced-concepts/route-matching.md
```diff
@@ -6,15 +6,35 @@ There are several rules Kitbag Router uses to determine which of your routes cor
...
+:x: `https://example.com/about-us`  
+:x: `/about-us`  
+
 ## Path Matches
 
 Routes `path` must match the structure of the URL pathname.
 
 ```ts
-const route {
+const route createRoute({
```
pleek91: I think these are all missing a =

PR #703, opened 2026-02-16
add hook for setting document title

stackoverfloweth:
This PR enables devs to set `document.title` with an easy and powerful callback. Users can setup a global callback, but probably much more likely on the route level

```ts
const userView = createRoute(...)

userView.setTitle((to, { title, from }) => {
  const displayName = await userStore.get(to.userId)

  return `User: ${displayName}`
})
```

Each setTitle callback receives the `to` (resolved route) and the `from` (could be undefined) as well as `title` which is a string | undefined, allows for "nesting" titles

```ts
const usersParent = createRoute(....)
usersParent.setTitle(() => 'User')

const userView = createRoute(...)

userView.setTitle((to, { title, from }) => {
  const displayName = await userStore.get(to.userId)

  return `${title}: ${displayName}`
  // same result ^
})
```

This PR pivots from a [previous attempt at a similar feature](https://github.com/kitbagjs/router/pull/693) by moving from a static property on route options to something that behaves more like a "hooks". It's a bit more awkward to assign (add 1+ callbacks after creating the route) but more powerful in the context it can supply. (mostly the params for the resolved route)

### DIFF
--- a/src/services/hooks.spec.ts
+++ b/src/services/hooks.spec.ts
@@ -322,7 +322,7 @@ describe('setTitle hook', () => {
       name: 'parentA',
       component,
     })
-    parentRoute.setTitle((to, { title }) => [to.name, title].filter(Boolean).join('<'))
+    parentRoute.setTitle((_to, { title }) => ['parentA', title].filter(Boolean).join('<'))
 
     const toRoute = createRoute({
       parent: parentRoute,


src/services/createRouter.ts
```diff
@@ -186,6 +187,9 @@ export function createRouter<
         throw new Error(`Switch is not exhaustive for after hook response status: ${JSON.stringify(exhaustive)}`)
     }
 
+    const title = await hooks.runTitleHooks({ to, from })
+    setDocumentTitle(title)
```
pleek91, line 191: We shouldn't block on this I don't think. Otherwise a navigation might be triggered before we start listening again. Though FWIW I think we're already doing this for after hooks and we probably shouldn't be.

src/services/createRouterHooks.ts
```diff
@@ -184,6 +185,30 @@ export function createRouterHooks(): RouterHooks {
...
+
+      for (const callback of allHooks) {
+        title = await runWithContext(() => callback(to, {
+          from: from as ResolvedRoute,
+          title,
+        }))
+      }
+
+      return title
+    } catch (error) {
```
pleek91, line 207: I wonder if we can structure these similarly to props where if you want the parents title you await it with the hook itself. That way we don't waterfall and also we could prevent the ancestors title hooks from even running if they're not needed potentially

src/types/hooks.ts
```diff
@@ -39,6 +40,10 @@ export type InternalRouteHooks<
    * Registers a route hook to be called after the route is updated.
    */
   onAfterRouteUpdate: AddAfterUpdateHook<[TRoute] | RouteContextToRoute<TContext>, RouteContextToRejection<TContext>, TRoute, Route>,
+  /**
+   * Registers a route hook to set the document title.
+   */
+  setTitle: AddTitleHook<[TRoute] | RouteContextToRoute<TContext>, Route>,
```
pleek91, line 46: I don't think we should consider this a hook necessarily. There are the internal concept of hooks and then there is the external concept of hooks. To users hooks are just the "onThing" type hooks listed here. 

We build other features on top of the internal hooks system. But "setTitle" isn't a hook as far as users are concerned.

src/utilities/setDocumentTitle.ts
```diff
@@ -0,0 +1,7 @@
+import { isBrowser } from '@/utilities/isBrowser'
+
+export function setDocumentTitle(title: string | undefined): void {
+  if (isBrowser() && title !== undefined) {
+    document.title = title
+  }
```
pleek91, line 6: Did you look into actually updating the DOM at all? I forgot to research that.

src/types/hooks.ts
stackoverfloweth: yeah that makes sense, to be clear is your suggestion just to change the docstring?

src/utilities/setDocumentTitle.ts
stackoverfloweth: I asked Perplexity, it said `document.title` was all that's needed 🤷

src/types/hooks.ts
pleek91: I think this should be similar to redirect and have its own types separate from hooks. Possibly its own factory as well though I don't feel strongly about that unless this ends up having an abstraction on top of the internal title hook.

src/utilities/setDocumentTitle.ts
pleek91: This work work for SSR though. I'm doing some research and POC stuff to figure out what we need to do here

src/services/createRouter.ts
pleek91: Also I think we should move setting the title to before the after hooks. Thoughts?

src/services/createRouterHooks.ts
pleek91: Having thought about this more I think we provide `parent` as context and let the user `await parent.title()` to run the parents title. But we'll only run the title callback for the match. 

This is similar to how props work but a little different (props always run).

pleek91: That'll also expose `parent.name` and such on the setTitle callback which seems useful.

PR #704, opened 2026-02-16
Fix route match and router route types to fix narrowing and type guard issues

pleek91:
# Description
This PR fixes two bugs which both were causing issues with narrowing `router.route` either by narrowing the route itself or when using `useRoute` or `isRoute`. 

## `Match` widening to `string`
Because `CreateRouteOptions['name']` is type `string` when a route had no name (either by being omitted or by passing an empty string) then `Route['matches'][number]['name']` would get widened to `string`. Would would cause issues for any routes with a no name parent.

To fix this I've implemented the `toName` both in the types and in the match created within `createRoute`. 

Originally I avoided this because generally the "match" type directly reflects what the user passed in. But we're already adding default values for `meta`, `state`, and `id` is added. So I think having a default value for `name` also makes sense here. 

I did attempt to fix this without modifying the match types and just fixing it within the `RouteWithMatch` type but that proved difficult (I could get one test case passing but not another). 

## `router.route` including no name routes
`router.route` is a union of all of the route possibilities that could be matched by the router. However this was including no name routes. Which could produce some weird types when doing route narrowing. 

To fix this I updated the existing `RouterRouteUnion` which previously was just a typescript hack to make a distributive type to also filter out routes with no name which makes the type more accurate and also gives it a meaningful purpose.

### DIFF
--- a/src/guards/routes.ts
+++ b/src/guards/routes.ts
@@ -7,9 +7,6 @@ export type IsRouteOptions = {
   exact?: boolean,
 }
 
-/**
- * Filters route union to routes whose match chain contains TRouteName (self or ancestor).
- */
 export type RouteWithMatch<
   TRoute extends RouterRoute,
   TRouteName extends TRoute['name']
@@ -63,7 +60,7 @@ type IsRouteFunction<TRouter extends Router> = {
 }
 
 export function createIsRoute<TRouter extends Router>(routerKey: InjectionKey<TRouter>): IsRouteFunction<TRouter> {
-  return ((route: unknown, routeName?: string, { exact }: IsRouteOptions = {}): route is any => {
+  return (route: unknown, routeName?: string, { exact }: IsRouteOptions = {}): route is any => {
     if (!isRouterRoute(routerKey, route)) {
       return false
     }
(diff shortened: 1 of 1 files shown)


src/compositions/useRoute.ts
pleek91, line 1: These changes didn't seem to effect anything but I do think this is probably good to have.

src/services/createRoute.ts
```diff
@@ -42,7 +42,7 @@ export function createRoute(options: CreateRouteOptions, props?: CreateRouteProp
   const state = options.state ?? {}
   const context = options.context ?? []
   const { store, ...hooks } = createRouteHooks()
-  const rawRoute = markRaw({ id, meta, state, ...options, props })
+  const rawRoute = markRaw({ ...options, id, meta, state, props, name })
```
pleek91, line 45: I added name and moved options to the beginning. I don't believe moving options to the beginning has any real effect, but this makes sure that the normalized consts will always override anything passed in as options. For example if an `id` value was passed by a user now it will be ignored. Types would prevent this but we don't want users hacking these sorts of things.

stackoverfloweth (approved):

PR #706, opened 2026-02-17
Add support for ssr context

pleek91:
# Description
Adds support for vue's ssr context to the router. This will allow us specifically to add the page title to the context so that users can insert the title into their DOM when doing ssr. 

This will allow doing something like this 
```ts
const router = createRouter([...], {
  initialUrl: '/',
})

const ctx = {}

const app = createSSRApp({
  template: '<RouterView />',
})

app.use(router)

const appHtml = await renderToString(app, ctx)
const appTitle = ctx.title ?? 'Default Title'

 const fullHtml = `
  <!DOCTYPE html>
  <html>
    <head>
      <title>${appTitle}</title>
    </head>
    <body>
      <div id="app">${appHtml}</div>
      <!-- Link to your client bundle for hydration -->
      <!-- <script type="module" src="/entry-client.js"></script> -->
    </body>
  </html>
`;
```

### DIFF
--- a/src/components/routerView.spec.ts
+++ b/src/components/routerView.spec.ts
@@ -5,12 +5,16 @@ import { createRouter } from '@/services/createRouter'
 import { renderToString } from 'vue/server-renderer'
 
 describe('SSR', () => {
-  it('should render the route', async () => {
+  it('should render the route and set ssr context', async () => {
+    const title = 'hello world'
+
     const route = createRoute({
       name: 'foo',
       path: '/',
       component: { template: 'hello world' },
     })
+
+    route.setTitle(async () => Promise.resolve(title))
 
     const router = createRouter([route], {
       initialUrl: '/',
@@ -28,6 +32,6 @@ describe('SSR', () => {
 
     expect(html).toMatchInlineSnapshot('"hello world"')
 
-    expect(ctx).toEqual({ foo: 'bar', title: 'hello world' })
+    expect(ctx).toEqual({ foo: 'bar', title })
   })
 })
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -42,6 +42,7 @@ import { setupRouterDevtools } from '@/devtools/createRouterDevtools'
 import { getMatchForUrl } from './getMatchesForUrl'
 import { pathHasTrailingSlash, removeTrailingSlashesFromPath } from '@/utilities/trailingSlashes'
 import { setDocumentTitle } from '@/utilities/setDocumentTitle'
+import { getRouteTitle } from '@/utilities/getRouteTitle'
 
 type RouterUpdateOptions = {
   replace?: boolean,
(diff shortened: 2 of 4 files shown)


src/services/createRouter.ts
```diff
@@ -233,6 +236,8 @@ export function createRouter<
         }
       })
 
+    ssrContextStore.addSsrContext('title', 'hello world')
```
pleek91: This is just an example of adding title to the ssr context. We'll need to combine the concepts in this PR with the new route callback being introduced in https://github.com/kitbagjs/router/pull/703

src/components/routerView.spec.ts
```diff
@@ -16,14 +16,18 @@ describe('SSR', () => {
...
     })
 
     app.use(router)
 
-    const html = await renderToString(app)
+    const html = await renderToString(app, ctx)
 
     expect(html).toMatchInlineSnapshot('"hello world"')
+
+    expect(ctx).toEqual({ foo: 'bar', title: 'hello world' })
```
pleek91: we'll want to update this test (or create a new test in this file) to test that `route.setTitle` is returned on the context after calling `renderToString`

src/services/createSsrContextStore.ts
stackoverfloweth, line 1: My first thought was wondering if this extra layer necessary? I thought we'd be able to just store our context in `useSSRContext` right from vue?

Then after looking more closely it looks like this enables us to basically set the ctx from RouterView (in component context) but then update that ctx from router. What if we just made `title: string` a part of the Route? Then in createRouter.set, we assign route.title, set document.title to route.title, and then from within RouterView (already has useRoute)

```ts
if (import.meta.env.SSR) {
  const context = useSSRContext() ?? {}

  context.title = route.title
}
```

This feels a bit less complex and having `title` on route _could_ be nice for users? Maybe they have navigation breadcrumbs or something?

src/services/createRouter.ts
```diff
@@ -166,7 +170,15 @@ export function createRouter<
...
     if (!isExternal(url)) {
-      setPropsAndUpdateRoute(navigationId, to, from)
+      setRouteProps(navigationId, to, from)
+
+      updateRoute(to)
+
+      getRouteTitle(to).then((title) => {
+        setDocumentTitle(title)
+
+        ssrContextStore.addSsrContext('title', title)
```
stackoverfloweth, line 180: Forgive my SSR ignorance.

I see that we're adding `title` to the context, but I guess I also expected us to be consuming this value somewhere else? Like the a `<title>`. Your test only really asserts that the ctx gets updated to include the `title`, is that in of itself valuable?

src/components/routerView.ts
```diff
@@ -28,12 +29,20 @@ export function createRouterView<TRouter extends Router>(routerKey: InjectionKey
...
     const route = useRoute()
     const router = useRouter()
     const rejection = useRejection()
     const depth = useRouterDepth({ increment: true })
+    const { setSsrContext } = useRouterSsrContext()
+
+    if (import.meta.env.SSR) {
+      const context = useSSRContext() ?? {}
+
+      setSsrContext(context)
```
stackoverfloweth, line 44: each RouterView overwrites the context in the store, yeah? Does that mean we potentially lose title whenever RouterView gets mounted? Again, forgive my SSR ignorance

PR #707, opened 2026-02-19
fix: RouterRoute update query

stackoverfloweth:
`router.route` uses proxies on query, that allows users to update (or replace) the query for the current route.

```ts
const router = useRouter()

router.route.query.get('tab')
router.route.query.set('tab', '3')
```

A user in discord discovered that if the route has a query, updating the `router.route.query` will end up duplicating whatever query is in the route definition.

```ts
createRoute({
  path: '/',
  query: {
    tab: withDefault(Number, 1),
  },
})
```

image

The issue was that the update function sends the new query (including "tab") into resolve as `RouterResolveOptions`, which gets combined with whatever query gets regenerated when you resolve a route to a URL. This PR updates the logic inside `router.route` to do a simpler update of the current url.

Note, this also affects the `useQueryValue` composable, which wraps `router.route.query`

The `update` function itself on router route is part of the public API, so I'm hesitant to change that.

### DIFF
--- a/src/services/createRouterRoute.ts
+++ b/src/services/createRouterRoute.ts
@@ -1,9 +1,11 @@
 import { computed, InjectionKey, reactive, toRefs } from 'vue'
+import { parseUrl, stringifyUrl, updateUrl } from '@/services/urlParser'
 import { ResolvedRoute } from '@/types/resolved'
 import { RouterRoute } from '@/types/routerRoute'
 import { RouterPush, RouterPushOptions } from '@/types/routerPush'
 import { QuerySource } from '@/types/querySource'
 import { Router } from '@/types/router'
+import { isPropertyKey } from '@/utilities/guards'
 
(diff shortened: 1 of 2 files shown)


src/services/createRouterRoute.ts
```diff
@@ -12,43 +14,50 @@ export function isRouterRoute(routerKey: InjectionKey<Router>, value: unknown):
...
+    const paramUpdates = individualParamNameOrAllParams
+    const options = valueOrOptions
     const params = {
       ...route.params,
-      [nameOrParams]: valueOrOptions,
+      ...paramUpdates,
     }
 
-    return push(route.name, params, maybeOptions)
+    return push(route.name, params, options)
```
pleek91, line 39: Is this a fix or just a cleanup/refactor of the update method?

stackoverfloweth: no change in functionality. I found the method a bit confusing so I thought I'd try to explain it better

PR #713, opened 2026-02-23
Ability for routes to set the page title automatically

pleek91:
# Description
Somewhat crud POC of some of the concept we've discussed around https://github.com/kitbagjs/router/pull/703

Purposefully unpolished but quite effective.

### DIFF
--- a/src/types/titles.browser.spec.ts
+++ b/src/types/titles.browser.spec.ts
@@ -1,18 +1,8 @@
-import { beforeEach, expect, test, vi } from 'vitest'
+import { expect, test, vi } from 'vitest'
 import { createRoute } from '@/services/createRoute'
 import { component } from '@/utilities/testHelpers'
 import { createRouter } from '@/main'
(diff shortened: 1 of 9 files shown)


src/types/titles.browser.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const setDocumentTitle = vi.hoisted(() => vi.fn())
+
+vi.mock('@/utilities/setDocumentTitle', () => ({
+  setDocumentTitle,
+}))
+
+beforeEach(() => {
+  vi.clearAllMocks()
+})
```
stackoverfloweth: Curious why you went this route? If we're in a browser spec we can check the document, if we're going to mock the document piece we don't need to be in a browser spec?

stackoverfloweth (approved):
this looks great! I have 1 question about the test and a couple suggestions in the form of [a follow-up PR](https://github.com/kitbagjs/router/pull/718). I tested this locally with router-preview, works great!

PR #718, opened 2026-03-05
code review suggestions

stackoverfloweth:
This PR adds 2 suggestions

1.) `getTitle` is now always a promise, which enables us to pretty easily trigger the setting of document title without awaiting. 

```ts
to.getTitle(to).then(setDocumentTitle)
```

2.) `setDocumentTitle` accepts `undefined`, and will revert to whatever the initial document title was

### DIFF
--- a/src/utilities/setDocumentTitle.ts
+++ b/src/utilities/setDocumentTitle.ts
@@ -1,7 +1,10 @@
 import { isBrowser } from '@/utilities/isBrowser'
 
+let defaultTitle: string
+
 export function setDocumentTitle(title: string | undefined): void {
-  if (isBrowser() && title !== undefined) {
-    document.title = title
+  if (isBrowser()) {
+    defaultTitle ??= document.title
+    document.title = title ?? defaultTitle
   }
 }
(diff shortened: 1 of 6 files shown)


src/utilities/setDocumentTitle.ts
```diff
@@ -1,7 +1,10 @@
 import { isBrowser } from '@/utilities/isBrowser'
 
+let defaultTitle: string
+
 export function setDocumentTitle(title: string | undefined): void {
-  if (isBrowser() && title !== undefined) {
-    document.title = title
+  if (isBrowser()) {
+    defaultTitle ??= document.title
+    document.title = title ?? defaultTitle
```
pleek91, line 8: I don't understand why this is needed. Shouldn't we just not set the title if title is nullish?

stackoverfloweth: My thinking is if they have a title set in the actual `index.html`, that should be the default. Without this change, once a route defines the title, it'll just stick with that forever until another route overrides it.

A good practice would be to take whatever title you set in `index.html` and also define `setTitle` in your "root" level route(s). This is a nicer fallback IMO

pleek91: Right but if we just don't set the document.title unless we have a title we don't need to do anything fancy right?

PR #720, opened 2026-03-05
Add literal utility

stackoverfloweth:
this PR adds a new `literal` utility, similar to `unionOf`, `arrayOf`, etc. This is nice for the sake of pulling the logic out, but it also solves a problem with our new query input syntax options.

**the problem:**

when users define params via `withParams`, we assume any literals supplied are still params (still true after this PR)

```ts
const route = defineRoute({
  query: withParams('?static=value&dynamic=[dynamic]', { dynamic: 42 })
})

route.params
//         ^? { dynamic: 42 }
```

however, with the new syntax options for query, it's less clear when the user is defining a static value vs a literal value

```ts
const route = defineRoute({
  query: {
    static: 'value',
    dynamic: 42,
  },
})

route.params
//         ^? {}
```

this PR resolves this by assuming any literals supplied with the new syntax are intended to be static, and will **not** be considered params. If a user explicitly wants a literal param, they can use `literal()`

```ts
import { literal } from '@kitbag/router'

const route = defineRoute({
  query: {
    static: 'value',
    dynamic: literal(42),
  },
})

route.params
//         ^? { dynamic: 42 }
```

### DIFF
--- a/docs/core-concepts/params.md
+++ b/docs/core-concepts/params.md
@@ -170,7 +170,7 @@ const events = createRoute({
 })
 ```
 
-Params can also be defined as a record or tuple of key-value pairs where the value is either a literal string or a [Param](/api/types/Param) argument.
+Params can also be defined as a record or tuple of key-value pairs.
 
 ```ts
 const events = createRoute({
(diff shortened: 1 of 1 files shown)


docs/core-concepts/params.md
```diff
@@ -200,22 +200,6 @@ const events = createRoute({
...
-## Literal Types
-
-Params can be defined as a literal value. This can be useful when an optional param can only have a single value when it is present.
-
-```ts
-import { withParams } from '@kitbag/router'
-
-const events = createRoute({
-  query: withParams('enabled=[?enabled]', {
-    enabled: true,
```
pleek91, line 212: Did this still work and we're just discouraging it?

stackoverfloweth: correct. I thought about adding something to docs about how to use literals in the new query syntax but then after reading it I just felt like it's not really something users will need to know/care about probably

PR #722, opened 2026-03-07
pull hooks out from public type surface

stackoverfloweth:
While investigating TS errors, I found the following 

> The inferred type of 'routes' cannot be named without a reference to '../../node_modules/@kitbag/router/dist/models/hooks'. This is likely not portable. A type annotation is necessary.

Screenshot 2026-03-07 at 15 38 37@2x

There are several of these types that we should consider pivoting from patterns like `@internal` to patterns like `WithRejections` with type guards so that our public types stay as pure as possible.

This PR primarily looks at the `Hooks` class, but I also exports a couple things in main that I think probably are good to export? If this is a pattern we want to continue, I may follow this up with similar PRs.

### DIFF
--- a/src/services/createRouterRoute.ts
+++ b/src/services/createRouterRoute.ts
@@ -6,7 +6,7 @@ import { RouterPush, RouterPushOptions } from '@/types/routerPush'
 import { QuerySource } from '@/types/querySource'
 import { Router } from '@/types/router'
 import { isPropertyKey } from '@/utilities/guards'
-import { isWithHooks } from '@/types/hooks'
+import { getHooks } from '@/types/hooks'
 
 const isRouterRouteSymbol = Symbol('isRouterRouteSymbol')
 
@@ -137,7 +137,7 @@ export function createRouterRoute<TRoute extends ResolvedRoute>(routerKey: Injec
     name,
     href,
     update,
-    hooks: isWithHooks(route) ? route.hooks : [],
+    hooks: getHooks(route),
     [isRouterRouteSymbol]: true,
     [routerKey]: true,
   })
--- a/src/types/createRouteOptions.ts
+++ b/src/types/createRouteOptions.ts
@@ -1,6 +1,7 @@
 import { Component } from 'vue'
 import { CombineMeta, combineMeta } from '@/services/combineMeta'
 import { CombineState, combineState } from '@/services/combineState'
+import { combineHooks, WithHooks } from '@/types/hooks'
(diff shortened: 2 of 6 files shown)


src/types/resolved.ts
```diff
@@ -46,12 +45,6 @@ export type ResolvedRoute<TRoute extends Route = Route> = Readonly<{
    * Hash value of the route.
    */
   hash: string,
-  /**
-   * The stores for routes including ancestors.
-   * Order of routes will be from greatest ancestor to narrowest matched.
-   * @internal
-   */
-  hooks: Hooks[],
```
pleek91, line 54: FWIW ResolvedRoute as a type is basically internal as a whole right? We don't actually expect or want users interacting with resolved routes. They're just a container to pass around. 

We can probably keep hooks on ResolvedRoute and just remove them from Route. 

I don't feel strongly here, I'm just not sure how much it make sense to remove anything from ResolvedRoute. It's not a public interface. Every property is internal.

src/types/createRouteOptions.ts
```diff
@@ -181,12 +182,12 @@ export type ToRoute<
...
     >
 
-export function combineRoutes(parent: Route, child: Route): Route {
+export function combineRoutes(parent: Route, child: Route): Route & WithHooks {
   return {
     ...child,
     meta: combineMeta(parent.meta, child.meta),
     state: combineState(parent.state, child.state),
-    hooks: [...parent.hooks, ...child.hooks],
+    hooks: [...isWithHooks(parent) ? parent.hooks : [], ...isWithHooks(child) ? child.hooks : []],
```
pleek91: combineHooks here?

src/services/createRouterRoute.ts
```diff
@@ -129,14 +130,14 @@ export function createRouterRoute<TRoute extends ResolvedRoute>(routerKey: Injec
...
     matches,
-    hooks,
     state,
     query,
     hash,
     params,
     name,
     href,
     update,
+    hooks: isWithHooks(route) ? route.hooks : [],
```
pleek91: Maybe a utility for this to keep this file clean?

src/types/resolved.ts
pleek91: Thinking about this more I think you're right here. Because we create RouterRoute from ResolvedRoute. So even though we never intend for users to directly interact with ResolvedRoute they will interact with RouterRoute. So stripping hooks here makes sense.

PR #726, opened 2026-03-11
Fix duplicate query values

pleek91:
# Description
Fix for duplicating params defined on the route itself when using route.query.set or useQueryValue


src/utilities/urlSearchParams.ts
pleek91, line 1: @stackoverfloweth 
TBH I'm not totally sure about this implementation. Seems like the bug might be elsewhere. But this is what I can cursor came up with to make the tests pass. 

But what worries me is it skips over keys if they exist in a previous param group. 

```ts
combineUrlSearchParams('?foo=bar', '?foo=baz')
```
return `?foo=bar` instead of `?foo=bar&foo=baz`. So this works, but seems like maybe the real bug is somewhere else and this is just masking the real issue.

PR #727, opened 2026-03-11
dedupe query params in getMatchesForUrl

stackoverfloweth:
`getMatchesForUrl` is a key piece of most navigation events, including setting initial url, push, replace, router.route.query.set, etc. There was some logic in `getMatchesForUrl` that was attempting to retain any extra unexpected query params but it was not explicitly de-deplicating the query params that are sourced from the params, which means any pass through `getMatchesForUrl` was appending an additional query param for the values that are sourced by params.

This PR adds a new utility `filterQueryParams` that takes source and exclude query params and returns only the key=value pairs from source that are not in exclude arg. 

This PR also pivots the order of `updateUrl` on url so that the query params that are populated by the params come before the extras. This _felt_ right to me but totally open to reverting since order doesn't really matter?

### DIFF
--- a/src/services/queryParamFilter.ts
+++ b/src/services/queryParamFilter.ts
@@ -4,10 +4,8 @@ export function filterQueryParams(source: QuerySource, exclude: QuerySource): UR
   const sourceParams = new URLSearchParams(source)
   const excludeParams = new URLSearchParams(exclude)
 
-  for (const [key, value] of sourceParams.entries()) {
-    if (excludeParams.has(key) && excludeParams.get(key) === value) {
-      sourceParams.delete(key)
-    }
+  for (const [key, value] of excludeParams.entries()) {
+    sourceParams.delete(key, value)
   }
 
   return sourceParams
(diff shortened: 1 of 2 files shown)


src/services/queryParamFilter.ts
```diff
@@ -0,0 +1,14 @@
+import { QuerySource } from '@/types/querySource'
+
+export function filterQueryParams(source: QuerySource, exclude: QuerySource): URLSearchParams {
+  const sourceParams = new URLSearchParams(source)
+  const excludeParams = new URLSearchParams(exclude)
+
+  for (const [key, value] of sourceParams.entries()) {
+    if (excludeParams.has(key) && excludeParams.get(key) === value) {
+      sourceParams.delete(key)
+    }
```
pleek91: I don't think this is quite right. This will delete all keys from the source if the first value is found in excluded. 

We need to consider all values here. 

It might be better to string replace here rather than using URLSearchParams.

stackoverfloweth: good catch, I think there's a way to actually make this even simpler

pleek91 (approved):

PR #766, opened 2026-05-13
drop defaultTitle from setDocumentTitle logic

### LINKED ISSUE #765: Router sets title of the page when setTitle is not used
It seems Kitbag incorrectly sets the title of the page when `setTitle` is not used, when individual components use `document.title = "foo"`.

We wanted to use https://router.kitbag.dev/core-concepts/routes.html#title, but we need to update the title when users change the language of our application (while staying on the same page).

We use our i18n library to get the string, and use `document.title` to set the title page. We cannot use `setTitle` due to our i18n library not being reactive.

We tried to use `document.title = "foo"` inside our pages, but that works only once, after that it goes back to the page title from the main component, and `document.title` from components has no effect afterwards.

## Repro

How to reproduce: Create a new vite app with Vue and Typescript, add kitbag `^0.24.1` to dependencies.

index.html:
```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>title from html</title>
  </head>
  <body>
    <div id="app"></div>
    <script type="module" src="/src/main.ts"></script>
  </body>
</html>
```

src/main.ts:
```typescript
import { createApp, defineAsyncComponent } from 'vue'
import App from './App.vue'
import { createRouter, createRoute } from '@kitbag/router'

const router = createRouter([
  createRoute({
    name: 'foo',
    path: '/',
    component: defineAsyncComponent(() => import('./Foo.vue')),
  }),
  createRoute({
    name: 'bar',
    path: '/bar',
    component: defineAsyncComponent(() => import('./Bar.vue')),
  }),
])

const app = createApp(App)
app.use(router)
app.mount('#app')
```

src/App.vue:
```Vue
<script setup lang="ts">
document.title = 'app'
</script>

<template>
<router-view/>
</template>
```

src/Foo.vue:
```Vue
<script setup lang="ts">
import { useRouter } from '@kitbag/router'
const router = useRouter()
document.title = 'foo'
</script>

<template>
foo
<button @click="() => router.push('bar')">go to bar</button>
</template>
```

src/Bar.vue:
```Vue
<script setup lang="ts">
import { useRouter } from '@kitbag/router'
const router = useRouter()
document.title = 'bar'
</script>

<template>
bar
<button @click="() => router.push('foo')">go to foo</button>
</template>
```

I recorded a screencast and extracted frames to show the issue (you can see it directly when running the application).

I used firefox here, but it reproduces on chrome as well.

1. Go the root page: the title from the html shows as the JS is not yet loaded.

Image

2. `App.vue` has loaded and sets the page title correctly.

Image

3. The router takes over and loads `Foo.vue`, title gets set correctly.

Image

4. Click on the button to ask the router to go to `/bar`.

Image

5. The title goes back to being the one from `App.vue` (do not blink or you will miss that), `Bar.vue` is being loaded.

Image

6. And right after that `Bar.vue` has rendered and sets the title correctly. So far so good.

Image

7. Click on the button to ask the router to go to `/`. `Foo.vue` is rendered, but the page title goes back to the one from `App.vue`. This is not expected.

Image

8. Click on the button to ask the router to go to `/bar`. `Bar.vue` is shown, but the page title does not update at all and stays to the one from `App.vue`.

Image

After that, the page title stays the same.

After investigating, we think that the issue is located in https://github.com/kitbagjs/router/blob/043b3178dfce899fa58b0371a7cb61265194cd67/src/utilities/setDocumentTitle.ts#L15.

The `setDocumentTitle()` function gets called even though `setTitle` is not used when declaring routes.

We have a patch in the works that solves the issue, we wanted to see if you are interested to accept it.

stackoverfloweth:
closes #765 

When we added the document title support to routes, we predicted routes that would control title and routes that wouldn't. The idea was to capture the "default title" (probably whatever is in the index.html) when the router is initialized and then go back to that title whenever we're navigating to a route that doesn't control the title. What this failed to consider are use cases like the one sent by @deadbaed, where the title is set elsewhere (in this case a component) the router's current logic doesn't consider a title that will change, or one that gets set later in the lifecycle. 

The solution is to simply drop the attempt to revert to a default title and leave the problem in user's hands. Probably either setting a default title in a "root" `route.setTitle` if you want the functionality we previously attempted to have by default

### DIFF
--- a/src/utilities/setDocumentTitle.ts
+++ b/src/utilities/setDocumentTitle.ts
@@ -2,16 +2,16 @@ import { ResolvedRoute } from '@/types/resolved'
 import { isRoute } from '@/types/route'
 import { isBrowser } from '@/utilities/isBrowser'
 
-let defaultTitle: string | undefined
-
 export function setDocumentTitle(to: ResolvedRoute | null): void {
   if (!isRoute(to) || !isBrowser()) {
     return
   }
 
-  defaultTitle ??= document.title
-
   to.title.then((value) => {
-    document.title = value ?? defaultTitle ?? ''
+    if (value === undefined) {
+      return
+    }
+
+    document.title = value
   })
 }


pleek91 (approved):

pleek91:
@stackoverfloweth we should probably update docs to inform users of the gotcha here around pages without titles and suggest setting a title on the root. Or we should allow setting a title on the router itself as the default title.

PR #772, opened 2026-05-17
RFC: Use method chaining and `addView` to create route types

pleek91:


stackoverfloweth (commented):
more or less exactly what I expected.. Hard to grok how much complexity it adds given this sandbox approach

pleek91:
Replaced by https://github.com/kitbagjs/router/pull/786

PR #791, opened 2026-07-28
Simplify the Route type by deriving name, meta, state, and context from matches

pleek91:
# Description
`Route` had grown to 7 generics and I was about to add another. Before doing that I wanted to see if any of the existing ones could be removed.

Turns out four of them can. `matches` is already a complete per-depth snapshot of the options each route was created with, and `combineMeta`/`combineState` are just intersections. So `TName`, `TMeta`, `TState`, and `TContext` were all recoverable from `TMatches` and never needed their own slots.

```ts
// before
Route<TName, TUrl, TMeta, TState, TMatches, TContext, TViews>

// after
Route<TUrl, TMatches, TViews>
```

`TUrl` stays because it's intersected into the route and carries `stringify`/`parse`/`tryParse`. `TViews` stays because `addView` refines it after construction, so it isn't a function of the options snapshot.

Technically breaking for anyone writing `Route<...>` with explicit type arguments, though it's mostly an internal construction detail.

No runtime changes.

### DIFF
--- a/src/types/addView.ts
+++ b/src/types/addView.ts
@@ -129,12 +129,8 @@ type WithViewProps<
   TRoute extends Route,
   TNewProps
 > = Route<
-  TRoute['name'],
   Pick<TRoute, keyof Url>,
-  TRoute['meta'],
-  TRoute['state'],
   TRoute['matches'],
-  TRoute['context'],
   ReplaceLastViewProps<TRoute['views'], TNewProps>
 >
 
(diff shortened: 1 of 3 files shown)


stackoverfloweth (approved):
❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️

PR #792, opened 2026-07-29
Add per-view prefetch config to addView

pleek91:
# Description

Prefetching could only be configured per route, which is too coarse when a route has multiple views. A dashboard with a light sidebar and a heavy report had to prefetch both or neither.

Each view can now set its own prefetch config, and it overrides the route's config for that view only. View config sits between route and link in the precedence chain, so `link > view > route > router`.

Since this meant a fourth positional argument, `addView` now takes an options object instead:

```ts
// before
.addView('sidebar', Sidebar, (route) => ({ id: route.params.id }))

// after
.addView(Sidebar, {
  name: 'sidebar',
  prefetch: 'eager',
})
```

Docs are updated for the new signature and the per-view config.

### DIFF
--- a/src/services/addView.ts
+++ b/src/services/addView.ts
@@ -1,44 +1,43 @@
 import { Component } from 'vue'
 import { DEFAULT_VIEW_NAME } from '@/services/createRouteViews'
 import { PropsGetter } from '@/types/createRouteOptions'
+import { PrefetchConfig } from '@/types/prefetch'
 import { Route } from '@/types/route'
 import { RouteViews } from '@/types/routeViews'
 
-type DefaultViewArgs = [component: Component, props?: PropsGetter]
-type NamedViewArgs = [name: string, component: Component, props?: PropsGetter]
+/**
+ * The loose runtime shape of the `addView` options. Purposely wide: the getter and name types each
+ * `RouteAddView` describes are refined per route and component.
+ */
+type ViewOptions = {
+  name?: string,
+  props?: PropsGetter,
+  prefetch?: PrefetchConfig,
+}
 
 /**
- * The runtime arguments accepted by `addView`: a default view (component first) or a named view (name first).
+ * The runtime arguments accepted by `addView`.
  */
-type AddViewParameters = DefaultViewArgs | NamedViewArgs
+type AddViewParameters = [component: Component, options?: ViewOptions]
 
 type NamedView = {
   name: string,
   component: Component,
   props: PropsGetter | undefined,
+  prefetch: PrefetchConfig | undefined,
 }
 
 /**
- * Whether the `addView` arguments are for a named view (a name string as the first argument) rather than
- * the default view (a component as the first argument).
+ * Normalizes `addView` arguments into a `{ name, component, props, prefetch }` view, defaulting the
+ * name to 'default'.
  */
-function isNamedView(args: AddViewParameters): args is NamedViewArgs {
-  return typeof args[0] === 'string'
-}
-
-/**
- * Normalizes `addView` arguments into a `{ name, component, props }` view, defaulting the name to 'default'.
- */
-function toNamedView(args: AddViewParameters): NamedView {
-  if (isNamedView(args)) {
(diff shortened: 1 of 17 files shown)


src/services/addView.ts
```diff
@@ -1,44 +1,43 @@
...
   name: string,
   component: Component,
   props: PropsGetter | undefined,
+  prefetch: PrefetchConfig | undefined,
 }
 
 /**
- * Whether the `addView` arguments are for a named view (a name string as the first argument) rather than
- * the default view (a component as the first argument).
+ * Normalizes `addView` arguments into a `{ name, component, props, prefetch }` view, defaulting the
```
pleek91: Lets just remove this comment. Its redundant to the type definition

src/services/addView.ts
```diff
@@ -1,44 +1,43 @@
...
+ * The runtime arguments accepted by `addView`.
  */
-type AddViewParameters = DefaultViewArgs | NamedViewArgs
+type AddViewParameters = [component: Component, options?: ViewOptions]
 
 type NamedView = {
   name: string,
   component: Component,
   props: PropsGetter | undefined,
+  prefetch: PrefetchConfig | undefined,
```
pleek91: WHy is this called a `NamedView` shouldn't it just be a  `View`? 🤔

src/services/addView.ts
```diff
@@ -71,18 +70,32 @@ function toPropsRecord(props: ViewProps): Record<string, PropsGetter> {
...
- * Returns a new {@link RouteViews} with a view (component + optional props getter) merged in. Components
- * are always stored as a record; props keep the bare getter for a lone default view and promote to a
- * record once a named view is added (pulling any existing bare default under 'default').
+ * Returns a new {@link RouteViews} with a view (component + optional props getter + optional per-view
+ * settings) merged in. Components and prefetch configs are always stored as records keyed by view name;
+ * props keep the bare getter for a lone default view and promote to a record once a named view is added
+ * (pulling any existing bare default under 'default').
  */
-function addToViews(views: RouteViews, name: string, component: Component, props: PropsGetter | undefined): RouteViews {
+function addToViews(views: RouteViews, { name, component, props, prefetch }: NamedView): RouteViews {
```
pleek91: As a follow up, lets investigate the shape of the views. I don't like how we have component, props, and prefect objects. We should jsut have a single object of { [viewName]: { component, props, prefetch } that we can use efficiently.

src/services/addView.ts
pleek91: Removed in cf1162df — the only non-obvious part was the name defaulting, and that's visible in the body.

src/services/addView.ts
pleek91: Agreed, it's a leftover. `NamedView` distinguished the two old argument shapes (`addView(Component)` vs `addView('sidebar', Component)`) normalized to always have a name. With the name in options there's only one shape, so renamed to `View`/`toView` in cf1162df.

src/services/addView.ts
pleek91: Opened as #798 (draft). Source is done — the three records collapse into one keyed by view name, and normalizing the props argument's two shapes at construction removes `ViewProps`, `isBareProps`, `isPropsRecord`, `toPropsRecord`, `addProps` and `addPrefetch`. It also turned up that parent props were being derived twice (once for the createRoute props argument, once for addView) with the same logic written two different ways — now one shared helper.

Still draft: 28 spec-d assertions target the old shape and need rewriting, and it wants a rebase after #793–#797 since those touch the same parent props types. Two open questions in the description, including whether to drop `id` from the entry so `route.views` is just a tuple of records.

stackoverfloweth (approved):
I like this. I could easily see us moving `component` into object too some day. Almost feels like the options object is another route at this point.

pleek91:
> I like this. I could easily see us moving `component` into object too some day. Almost feels like the options object is another route at this point. 

Maybe if ts inference gets better. But two args lets us infer the props better.

PR #795, opened 2026-07-29
Fix a child receiving undefined where it awaits parent props

pleek91:
# Description

Stacked on #802.

A child awaiting its parent's props received `undefined` whenever those props hadn't been computed yet — either because they aren't prefetched, or because they're prefetched at a different strategy than the child's. The types promise the parent's props, so destructuring silently produced undefined fields.

```ts
const parent = createRoute({
  name: 'parent',
  path: '/parent',
}).addView(ParentLayout, {
  props: async () => ({ foo: 123 }),
  prefetch: { props: 'intent' },
})

const child = createRoute({
  name: 'child',
  parent,
  path: '/child',
}).addView(ChildPage, {
  props: async (route, { parent }) => {
    const value = await parent.props
    // before: undefined, because the parent prefetches at a different strategy
    // after:  waits, then { foo: 123 }

    return { value }
  },
  prefetch: { props: 'eager' },
})
```

The child now waits for the parent's props to be computed by the parent's own lifecycle. Nothing runs earlier than configured — running the parent's getter early would override a parent explicitly told not to prefetch — and navigation writes every view's props before it awaits anything, so a waiting child always resolves by then.

Props already prefetched at an earlier strategy are read from that batch rather than waited for, so a child prefetching later than its parent no longer stalls until navigation. A view name with no getter reads as `undefined` instead of waiting for props that can never arrive.

## Breaking change

`parent.props` is always a promise now and must be awaited. The type used to mirror the parent's getter, so a synchronous getter typed it as a plain value, which can't be true once the props might still be pending.

```ts
// before
const { foo } = parent.props

// after
const { foo } = await parent.props
```

### DIFF
--- a/src/compositions/usePrefetching.ts
+++ b/src/compositions/usePrefetching.ts
@@ -32,13 +32,15 @@ export function createUsePrefetching<TRouter extends Router>(routerKey: Injectio
     const { isElementVisible } = useVisibilityObserver(element)
 
     const commit: UsePrefetching['commit'] = () => {
-      const props = Array.from(prefetchedProps.values()).reduce<Record<string, MaybePromise<PropsResult>>>((accumulator, value) => {
+      setPrefetchProps(getAccumulatedProps())
+    }
+
+    function getAccumulatedProps(): Record<string, MaybePromise<PropsResult>> {
+      return Array.from(prefetchedProps.values()).reduce<Record<string, MaybePromise<PropsResult>>>((accumulator, value) => {
         Object.assign(accumulator, value)
 
         return accumulator
       }, {})
-
-      setPrefetchProps(props)
     }
 
     watch(() => toValue(config), ({ route, ...configs }) => {
@@ -78,7 +80,7 @@ export function createUsePrefetching<TRouter extends Router>(routerKey: Injectio
       prefetchComponentsForRoute(strategy, route, configs)
 
       if (!prefetchedProps.has(strategy)) {
-        prefetchedProps.set(strategy, getPrefetchProps(strategy, route, configs))
+        prefetchedProps.set(strategy, getPrefetchProps(strategy, route, configs, getAccumulatedProps()))
       }
     }
 
--- a/src/services/createPropStore.ts
+++ b/src/services/createPropStore.ts
@@ -1,4 +1,4 @@
-import { reactive } from 'vue'
+import { effectScope, reactive, watch, WatchHandle } from 'vue'
 import { isWithComponentProps, isWithComponentPropsRecord, PropsGetter } from '@/types/createRouteOptions'
 import type { PrefetchConfigs, PrefetchStrategy } from '@/types/prefetch'
 import { getPrefetchOption } from '@/utilities/prefetch'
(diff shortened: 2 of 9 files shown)


src/services/createRoute.spec.ts
```diff
@@ -242,8 +242,12 @@ describe('props', () => {
       name: 'child',
       parent,
       path: '/child',
-    }, (__, { parent }) => {
-      seen.push({ self: 'child', parentName: parent.name, parentProps: parent.props })
+    }, async (__, { parent }) => {
+      // parent.props is a promise at runtime, but its type collapses to undefined because the parent's
+      // own getter consumes the callback context, which makes its inferred props type as wide as the
+      // constraint in ToRoute
+      // eslint-disable-next-line @typescript-eslint/await-thenable
```
pleek91: this is fixed in the next PR #797

src/services/createPropStore.ts
```diff
@@ -179,41 +188,100 @@ export function createPropStore(): PropStore {
...
+  }
+
+  function warnWaitingForParentProps(name: string, route: ResolvedRoute): void {
+    const routeName = route.name || 'unknown'
+
+    console.warn(`
+      Waiting on parent props "${name}" while prefetching props for route "${routeName}".
+      The parent's props are not being prefetched at this point, so these props cannot resolve until the
+      parent's props are computed — either by the parent's own prefetch strategy or by navigating.
+      Prefetch the parent's props with the same strategy to avoid stalling here.
```
stackoverfloweth, line 270: is this actually true? I read this as, we can't actually get the parents props but it seems like we fire this warning but then continue to `waitForProps`. I like the idea of letting the developer know we're circumventing their prefetching strategy on the parent - but it looks like we do go get the parent props regardless?

probably just misunderstood the warning message, since the docs change seems to enforce what I believe to be true

>  if the parent's props are prefetched at a different strategy than yours, or not prefetched at all, awaiting them waits until they are.

src/compositions/usePrefetching.ts
```diff
@@ -32,13 +32,15 @@ export function createUsePrefetching<TRouter extends Router>(routerKey: Injectio
     const { isElementVisible } = useVisibilityObserver(element)
 
     const commit: UsePrefetching['commit'] = () => {
-      const props = Array.from(prefetchedProps.values()).reduce<Record<string, MaybePromise<PropsResult>>>((accumulator, value) => {
+      setPrefetchProps(getAccumulatedProps())
+    }
+
+    function getAccumulatedProps(): Record<string, MaybePromise<PropsResult>> {
+      return Array.from(prefetchedProps.values()).reduce<Record<string, MaybePromise<PropsResult>>>((accumulator, value) => {
```
stackoverfloweth, line 39: nit: this doesn't seem to be needed

```suggestion
      return Array.from(prefetchedProps.values()).reduce((accumulator, value) => {
```

stackoverfloweth (approved):
took a minute to fully grasp but it seems like a great solution for a tricky problem. I also love the waiters and use of our own detached effect scope

PR #797, opened 2026-07-30
Fix a parent's props type being lost in either syntax when its getter declares arguments

pleek91:
# Description

Bug 4 of 4. Stacked on #795.

A child's `parent.props` typed as `undefined` whenever the parent's props getter declared any argument — which is most of them. The route itself kept its props type; only the child's view of it was lost. Both ways of writing a getter are affected, `addView` and the `createRoute` props argument.

```ts
const parent = createRoute({
  name: 'parent',
  path: '/parent/[id]',
}).addView(ParentLayout, {
})

const child = createRoute({
  name: 'child',
  parent,
  path: '/child',
}).addView(ChildPage, {
  props: async (route, { parent }) => {
    const { id } = await parent.props
    // before: parent.props is typed undefined, so this does not compile
    // after:  Promise<{ id: string }>

    return { id }
  },
})
```

The same happens when the parent declares its getter through the props argument:

```ts
const parent = createRoute({
  name: 'parent',
  path: '/parent/[id]',
}, (route) => ({ id: route.params.id }))
```

Both types that reconstruct a parent's props for a child asked whether the getter `extends PropsGetter`. That signature's `route` argument is the wide `ResolvedRoute<Route>`, and a getter written `(route) => ...` has its argument narrowed to one specific route. Parameters are contravariant, so the check fails and falls through to `undefined`. Getters taking no arguments passed only because a zero argument function is assignable to any signature — which is why this went unnoticed, since the tests used `() => ({ foo: 123 })`.

These checks meant "is this a function", so they now ask that. `AddViewProps` already did it this way, so the same `AnyFunction` is shared between them.

Also removes the `eslint-disable` that #795 needed for exactly this reason.

### DIFF
--- a/src/types/utilities.ts
+++ b/src/types/utilities.ts
@@ -12,6 +12,8 @@ export type LastInArray<T, TFallback = never> = T extends [...any[], infer Last]
 
 export type MaybePromise<T> = T | Promise<T>
 
+export type AnyFunction = (...args: any[]) => any
+
 type OnlyRequiredProperties<T> = {
   [K in keyof T as Extract<T[K], undefined> extends never ? K : never]: T[K]
 }
--- a/src/services/addView.spec-d.ts
+++ b/src/services/addView.spec-d.ts
@@ -257,6 +257,45 @@ describe('addView', () => {
         return {}
       })
     })
+
+    test('bare parent props are passed to a child when the parent getter declares its arguments', () => {
+      const parent = createRoute({ name: 'parent', path: '/parent/[id]' })
+        .addView(component, {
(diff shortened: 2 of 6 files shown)


src/types/utilities.ts
```diff
@@ -12,6 +12,8 @@ export type LastInArray<T, TFallback = never> = T extends [...any[], infer Last]
 
 export type MaybePromise<T> = T | Promise<T>
 
+export type AnyFunction = (...args: any[]) => any
```
stackoverfloweth, line 15: how did we not have this?!?

stackoverfloweth (approved):

PR #800, opened 2026-07-30
Remove the deprecated component, components and props options

pleek91:
# Description

Removes the deprecated `component` and `components` options and the props argument. Views are added with `addView`.

## Breaking change

```ts
// before
createRoute({
  name: 'user',
  path: '/user/[id]',
  component: UserPage,
}, (route) => ({ id: route.params.id }))

// after
createRoute({
  name: 'user',
  path: '/user/[id]',
})
.addView(UserPage, {
})
```

Named views move from `components` to a name per view.

```ts
// before
createRoute({
  name: 'user',
  path: '/user/[id]',
  components: { default: UserPage, sidebar: UserSidebar },
}, {
  default: (route) => ({ id: route.params.id }),
  sidebar: (route) => ({ id: route.params.id }),
})

// after
createRoute({
  name: 'user',
  path: '/user/[id]',
})
.addView(UserPage, {
})
.addView(UserSidebar, {
  name: 'sidebar',
})
```

Passing props without a component is no longer supported. A view needs a component to render.

Passing `component` or `components` is now an error rather than being ignored, so the compiler points at anything left to migrate.


stackoverfloweth (approved):

pleek91:
@stackoverfloweth not going to merge this until after we do the next minor release so we have one version where both are supported for users to use as a migration version.

PR #802, opened 2026-07-31
Record how a props getter settled instead of just its value

pleek91:
# Description

Stacked on #794.

Props were stored as whatever the getter returned, which left two real values indistinguishable from the states they stood in for.

A getter returning an `Error` looked the same as one that threw, so the view rendered nothing:

```ts
createRoute({
  name: 'route',
  path: '/route',
}).addView(ShowsProps, {
  props: () => someValueThatMightBeAnError(),
})
// before: renders nothing whenever the value happens to be an Error
// after:  renders, because a returned Error is a value like any other
```

A getter returning `undefined` looked the same as one that had not run yet. On its own that is harmless, but it becomes a hang once children can wait for their parent's props (#795).

Each entry now records how the getter settled alongside what it produced, so `undefined` and `Error` are ordinary values a getter may return. Nothing about the public API changes.

### DIFF
--- a/src/utilities/props.ts
+++ b/src/utilities/props.ts
@@ -1,20 +1,28 @@
+import { MaybePromise } from '@/types/utilities'
 import { isPromise } from './promises'
 
 /**
- * Takes a props callback and returns a value
- * For sync props, this will return the props value or an error.
- * For async props, this will return a promise that resolves to the props value or an error.
+ * How a props getter settled. Tagged so that a getter returning `undefined` or an `Error` stays a value.
  */
-export function getPropsValue(callback: () => unknown): unknown {
+export type PropsResult = { kind: 'value', value: unknown } | { kind: 'error', error: unknown }
+
+/**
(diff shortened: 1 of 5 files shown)


src/utilities/props.ts
```diff
@@ -1,20 +1,28 @@
...
 import { isPromise } from './promises'
 
 /**
- * Takes a props callback and returns a value
- * For sync props, this will return the props value or an error.
- * For async props, this will return a promise that resolves to the props value or an error.
+ * How a props getter settled. Tagged so that a getter returning `undefined` or an `Error` stays a value.
  */
-export function getPropsValue(callback: () => unknown): unknown {
+export type PropsResult = { kind: 'value', value: unknown } | { kind: 'error', error: unknown }
```
stackoverfloweth, line 7: ❤️

stackoverfloweth (approved):
nice, simple and elegant

PR #806, opened 2026-07-31
Add a store for values that may not be computed yet

pleek91:
# Description

First step of #805. Nothing uses this yet — the store is added on its own so it can be reviewed without the churn of rewiring props at the same time.

A value becomes a promise the moment it is first asked for, rather than only existing once its getter has run. Whoever needs it can await it without knowing when, or by whom, it will be computed:

```ts
store.subscribe('user')          // nothing has computed this yet — waits
store.set('user', getUser)       // resolves the promise the subscriber already holds
```

Alongside the promise, each key records where it has got to — nothing has asked for it, something is waiting on it, its getter is in flight, or how that getter settled. Rendering reads that synchronously instead of waiting a tick, and a getter returning `undefined` or an `Error` stays an ordinary value rather than being mistaken for "not ready" or "failed".

A store is also the unit of lifetime. Disposing one rejects everything still outstanding, so anything waiting resumes rather than being suspended for the life of the page — which is what prefetching needs, since most prefetched work is for a navigation that never happens.

### DIFF
--- a/src/services/createDataStore.spec.ts
+++ b/src/services/createDataStore.spec.ts
@@ -1,0 +1,198 @@
+import { describe, expect, test, vi } from 'vitest'
+import { effect } from 'vue'
+import { createDataStore } from './createDataStore'
+
+describe('set and subscribe', () => {
+  test('a value subscribed to before anything computes it resolves when it is computed', async () => {
+    const store = createDataStore()
+    const settled: unknown[] = []
+
+    store.subscribe('key').then((value) => settled.push(value))
+
+    expect(settled).toStrictEqual([])
+
+    store.set('key', () => 'computed')
+
+    await Promise.resolve()
+
+    expect(settled).toStrictEqual(['computed'])
+  })
+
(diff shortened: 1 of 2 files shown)


stackoverfloweth (approved):
My only question is the value of having both a pending state and a running state. I see that something can "watch" for a value by being subscribed and that's meaningful for the store. I also see that when it's running it's actually trying to fulfill a promise that will update those who are watching. Is that basically it?

pleek91:
> My only question is the value of having both a pending state and a running state. I see that something can "watch" for a value by being subscribed and that's meaningful for the store. I also see that when it's running it's actually trying to fulfill a promise that will update those who are watching. Is that basically it?

A `pending` state means something requested it but nothing has actually run. So a promise was created and is just waiting for something to actually resolve/reject it. 

Vs `running` means something is actively attempting to resolve/reject that promise. Subtle difference but important for tracking the state of the promise (which we cannot inspect to see if its being resolve or not).

PR #807, opened 2026-08-01
Compute props through the data store

pleek91:
# Description

Second step of #805, stacked on #806. No behaviour changes — every existing test passes unchanged.

Props move onto the store added in #806, one store per navigation. A link prefetches into its own store and stages it when followed; the navigation it triggers adopts that store, and the one it replaces is disposed outright rather than picked over for entries the new route still needs.

A getter waiting on its parent now awaits a value rather than watching for one to appear:

```ts
await Promise.any([
  own.subscribe(parentKey),                  // the parent's props, if this link prefetches them
  navigation.current().subscribe(parentKey), // or the ones navigation computes
])
```

That removes the waiter machinery entirely — the detached effect scope, the watch handles, and the bookkeeping to discard them. Prefetch strategies also stop being separate buckets, so a value prefetched on render is visible to a getter that runs on hover without being passed along explicitly.

### DIFF
--- a/src/components/routerLink.browser.spec.ts
+++ b/src/components/routerLink.browser.spec.ts
@@ -11,7 +11,7 @@ import { visibilityObserverKey } from '@/compositions/useVisibilityObserver'
 import { VisibilityObserver } from '@/services/createVisibilityObserver'
 import { UrlString } from '@/types/urlString'
 import { RouterPushOptions } from '@/types/routerPush'
-import { ParentPropsAbandonedError, RouterLink } from '@/main'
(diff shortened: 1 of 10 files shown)


stackoverfloweth (approved):
left some small suggestions in [a commit](https://github.com/kitbagjs/router/pull/807/changes/f8ff1157c5b01dec1799c7399eca52d818526e2d), love how much this simplifies the prefetching. The props store changes are pretty sweeping and hard to really grok fully but I like what I see

PR #810, opened 2026-08-03
Give every match a loaders record

pleek91:
# Description

Third step of #805, stacked on #809. No behaviour changes — nothing reads these records yet.

Loaders are a route's data, declared the way views are declared but with no component attached and no hold on rendering. Storing them per match, alongside views, means a route's data can be derived from its matches the same way its views already are — including an ancestor's loaders, which a child should be able to use as its own.

`route.data` is that derivation: the loaders of every match combined, flattened to the loader's data directly when the only loader in the tree is unnamed. This PR adds the types and the empty record; the method that fills it and the store that runs it follow.


pleek91:
Folded into #811 — the loaders record isn't useful on its own without the method that fills it.

PR #811, opened 2026-08-03
Add loaders to routes with a chainable addLoader

pleek91:
# Description

Third step of #805, stacked on #809.

Loaders are a route's data, declared the way views are declared but with no component attached and no hold on rendering:

```ts
createRoute({ name: 'user', path: '/user/[id]' })
  .addView(UserView, { props: (route) => ({ id: route.params.id }) })
  .addLoader(async (route) => api.getUser(route.params.id))
  .addLoader(async (route) => api.getPosts(route.params.id), { name: 'posts' })
```

They are stored per match, alongside views, so a route's data can be derived from its matches the same way its views already are — including an ancestor's loaders, which a child should be able to use as its own.

Both chainable methods now rebuild through `withRouteMethods`. Previously `addView` re-attached only itself, so chaining a view after a loader would have dropped `addLoader` from the route — the two have to be re-attached together, and neither file needs to know about the other for that.

A loader name an ancestor already uses throws `LoaderNameConflict`. A route's data combines the loaders of every match, so the same name in two places would be ambiguous rather than an override — unlike adding a loader the route itself already has, which replaces it, as views do.

Nothing runs a loader yet. What this proves is declaration: names, prefetch configs, immutability, chaining in both orders, and ancestor conflicts.

### DIFF
--- a/src/errors/loaderNameConflict.ts
+++ b/src/errors/loaderNameConflict.ts
@@ -1,0 +1,11 @@
+/**
+ * An error thrown when a loader is added with the same name as a loader on one of the route's ancestors.
+ * A route's data combines the loaders of every match, so the same name twice would be ambiguous rather
+ * than an override.
+ * @group Errors
+ */
+export class LoaderNameConflict extends Error {
+  public constructor(name?: string) {
+    super(`Loader "${name}" conflicts with a loader of the same name on an ancestor route.`)
+  }
+}
--- a/src/services/addLoader.spec.ts
+++ b/src/services/addLoader.spec.ts
@@ -1,0 +1,145 @@
+import { describe, expect, test } from 'vitest'
+import { createRoute } from '@/services/createRoute'
+import { LoaderNameConflict } from '@/errors/loaderNameConflict'
+import { RouteLoaders } from '@/types/routeLoaders'
+import { RouteViews } from '@/types/routeViews'
+import { component } from '@/utilities/testHelpers'
+
+const load = (): { foo: string } => ({ foo: 'bar' })
+const other = (): { baz: number } => ({ baz: 1 })
+
+function lastMatch(route: { matches: { loaders: RouteLoaders, views: RouteViews }[] }): { loaders: RouteLoaders, views: RouteViews } {
+  return route.matches[route.matches.length - 1]
+}
+
+function pick(route: { matches: { loaders: RouteLoaders, views: RouteViews }[] }, key: 'load' | 'prefetch'): Record<string, unknown> {
+  return Object.fromEntries(
+    Object.entries(lastMatch(route).loaders)
+      .filter(([, loader]) => loader[key] !== undefined)
+      .map(([name, loader]) => [name, loader[key]]),
+  )
+}
+
+describe('loaders', () => {
+  test('a loader added without a name is stored under "default"', () => {
+    const route = createRoute({ name: 'route' }).addLoader(load)
+
+    expect(pick(route, 'load')).toStrictEqual({ default: load })
+  })
+
+  test('a named loader is stored under its name', () => {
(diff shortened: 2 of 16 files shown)


src/types/addView.ts
```diff
@@ -162,6 +140,7 @@ export type RouteWithMethods<
   TMatches extends CreatedRouteOptions[] = CreatedRouteOptions[]
 > = Route<TUrl, TMatches>
   & RouteAddView<TUrl, TMatches>
+  & RouteAddLoader<TUrl, TMatches>
```
stackoverfloweth: it's surprising to see this inside `addView.ts`. I wonder if we should have a types equivalent of your nice `withRouteMethods` abstraction

src/services/addLoader.ts
```diff
@@ -0,0 +1,67 @@
+import { markRaw } from 'vue'
+import { LoaderNameConflict } from '@/errors/loaderNameConflict'
+import { PrefetchConfig } from '@/types/prefetch'
+import { CreatedRouteOptions, Route } from '@/types/route'
+import { AnyFunction } from '@/types/utilities'
+
+export const DEFAULT_LOADER_NAME = 'default'
```
stackoverfloweth, line 7: I'm interested in this paradigm. I see how this nicely mirrors our treatment of views but it got me thinking about alternative approaches that might be worth a discussion. 

I talked to Claude about this, and feel like what we have is actually pretty ideal. I'll include my thoughts anyways if you're curious.  I think it's interesting how similar we store views and loaders, could probably see some future code sharing.

> if loaders aren't given names we could use a tuple. So a single loader without name could be 
> 
> ```
> const [myData] = route.data
> ```
> 
> multiple loaders could be added without name and accessed by index. 
> 
> if this feels right, could we do the same for views? Do we use `default` because that's what vue-router does?

src/errors/loaderNameConflict.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
stackoverfloweth, line 1: I'm interested in this too. I think we talked about how nice it would be if you could just override a parent data key with a Childs, and still access the parent if you need to with `useRoute('parent').data.sameKey`. It seems like you found that to be too ambiguous in practice?

src/services/addLoader.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+  return route.matches[route.matches.length - 1]
+}
+
+function pick(route: { matches: { loaders: RouteLoaders, views: RouteViews }[] }, key: 'load' | 'prefetch'): Record<string, unknown> {
+  return Object.fromEntries(
+    Object.entries(lastMatch(route).loaders)
+      .filter(([, loader]) => loader[key] !== undefined)
+      .map(([name, loader]) => [name, loader[key]]),
+  )
+}
```
stackoverfloweth, line 21: I personally don't like this kind of thing and prefer verbose tests over layered utilities, but I won't stop you

stackoverfloweth (approved):
❤️

src/services/addLoader.ts
pleek91: Variable length tuples are a bad devx IMO. And there's no guarantee that there is a default for either views or loaders. So I think referencing them by name in an object is preferred here.

src/errors/loaderNameConflict.ts
pleek91: I think it's a good idea still but needs some more thought/work. I think what I have here doesn't prevent us from supporting that in the future. But for now it's easy enough to just require unique names.

src/services/addLoader.spec.ts
pleek91: Yeah this is Claude. What do you mean layered utilities?

src/types/addView.ts
pleek91: Valid, moved

PR #817, opened 2026-08-16
Add AGENTS.md and SKILL.md for LLM agent discoverability

stackoverfloweth:
## Summary

- Adds `AGENTS.md` at package root with install instructions, 5 canonical usage patterns, type-inference gotchas, an "APIs that changed — don't write the old form" table, and key differences from vue-router
- Adds `skills/kitbag-router/SKILL.md` with frontmatter trigger conditions and a complete API reference covering routes, params, navigation, components, composables, hooks, rejections, and plugins
- Updates `package.json` `"files"` to include both so they ship in the npm tarball

These files help LLM coding agents generate correct @kitbag/router code instead of hallucinating vue-router patterns from stale training data. Compatible with [TanStack Intent](https://tanstack.com/intent/latest) discovery and manual agent config.

## Test plan

- [ ] Verify `npm pack --dry-run` includes `AGENTS.md` and `skills/kitbag-router/SKILL.md`
- [ ] Review AGENTS.md content for accuracy against current API
- [ ] Review SKILL.md frontmatter description for appropriate trigger conditions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -78,12 +78,22 @@ router.replace('home')
 ```ts
 import { useRoute } from '@kitbag/router'
 
-// Unnarrrowed — all possible params are optional
+// Unnarrowed — all possible params are optional
 const route = useRoute()
 
 // Narrowed to a specific route — params are typed
 const userRoute = useRoute('user')
 userRoute.params.userId // number
+```
+
+Params are reactive and support `v-model` for two-way binding (especially useful for query params):
+
+```vue
+<input v-model="route.params.search" />
+<select v-model="route.params.sort">
+  <option value="asc">Ascending</option>
+  <option value="desc">Descending</option>
+</select>
 ```
 
 ### 5. Use hooks for guards, redirects, and lifecycle
@@ -97,10 +107,17 @@ const protectedRoute = createRoute({
 })
   .addView(Dashboard)
 
-// Route-level hook — use throw to stop execution
-protectedRoute.onBeforeRouteEnter((_to, { reject }) => {
+// Route-level before-hook — use `throw reject(...)` to stop execution
+protectedRoute.onBeforeRouteEnter((to, { reject, replace }) => {
   if (!isAuthenticated()) {
     throw reject('NotFound')
+  }
+})
+
+// Route-level after-hook — second arg includes `push` for navigation
+protectedRoute.onAfterRouteEnter((to, { push }) => {
+  if (to.params.search === 'secret') {
+    push('hiddenRoute')
   }
 })
 
@@ -114,9 +131,73 @@ onBeforeRouteLeave((to, { abort }) => {
 })
 ```
 
+## Additional Features
+
+### Query params, `unionOf`, and `withDefault`
+
+Query params are first-class. Use shorthand strings for untyped params, or `withParams` with `unionOf`/`withDefault` for constrained values:
+
+```ts
+import { createRoute, withParams, unionOf, withDefault } from '@kitbag/router'
+
+// Shorthand — defaults to String type
+const search = createRoute({
+  name: 'search',
+  path: '/search',
+  query: 'q=[?q]',
+})
+
+// Typed with constrained values and a default
+const keys = createRoute({
+  name: 'keys',
+  path: '/keys',
+  query: withParams('sort=[?sort]', {
+    sort: withDefault(unionOf(['asc', 'desc']), 'asc'),
+  }),
+})
+```
+
+### Context — scoping routes and rejections
+
+Routes can declare `context` — an array of other routes or rejections that are only accessible when that route is active:
+
+```ts
+const secretPage = createRoute({ name: 'secret', path: '/secret' })
+  .addView(SecretView)
+
+const main = createRoute({
+  name: 'main',
+  path: '/main',
+  context: [secretPage],
+}).addView(MainView)
+
+// Rejections can also be scoped via context
+const notAuthorized = createRejection({ type: 'NotAuthorized', component: LoginView })
+
+const protectedRoute = createRoute({
+  name: 'protected',
+  path: '/protected',
+  context: [notAuthorized],
+})
+```
+
+### RouterView scoped slot for transitions
+
+Use the `#default` slot with `{ component }` to wrap route views in `<transition>`:
+
+```vue
+<router-view>
+  <template #default="{ component }">
+    <transition name="fade" mode="out-in">
+      <component :is="component" />
+    </transition>
+  </template>
+</router-view>
+```
+
 ## Type Inference Gotchas
 
-1. **You must register the router** via declaration merging or use `createRouterAssets` — without this, composables like `useRoute` and `useRouter` have no type context.
+1. **You must register the router** via declaration merging — without this, composables like `useRoute` and `useRouter` have no type context.
 2. **Routes array must use `as const`** — without it TypeScript widens route names to `string` and params lose their types.
 3. **`useRoute('name')` narrows types** — only use this inside components you know are mounted under that route. Using the wrong name gives a runtime error.
 4. **Param names must be unique** across the full route tree (including parent routes). Duplicate param names throw `DuplicateParamsError`.
@@ -136,7 +217,7 @@ LLMs trained on older data may generate outdated patterns. The correct forms are
 | `router.beforeEach(...)` | Use `onBeforeRouteEnter` on the route or pass hooks to `createRouter` |
 | `{ path: '/user/:id' }` | `{ path: withParams('/user/[id]', { id: Number }) }` (square brackets, explicit type) |
 | `meta: { requiresAuth: true }` inline | `createRoute({ meta: { requiresAuth: true } })` (same place, but access differs) |
-| `import { useRejection } from 'vue-router'` | `import { useRejection } from '@kitbag/router'` (rejections are a Kitbag concept) |
+| `reject('NotFound')` | `throw reject('NotFound')` (`throw` is needed to stop hook execution) |
 
 ## Key Differences from vue-router
 
@@ -148,3 +229,5 @@ LLMs trained on older data may generate outdated patterns. The correct forms are
 - **Props are type-safe** — assigned via `.addView(Component, { props: (route) => ({ ... }) })`.
 - **Query params are first-class** — defined with the same `withParams` syntax on the `query` property.
 - **Plugins** bundle routes, rejections, and hooks into reusable units via `createRouterPlugin`.
+- **Context** scopes routes and rejections to only be accessible when a parent route is active.
+- **Params support `v-model`** — two-way binding works directly on `route.params`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,150 @@
+# @kitbag/router - Agent Guide
+
+Type-safe router for Vue 3. Not a fork of vue-router — different API, different mental model.
+
+Docs: https://kitbag.dev/router
+
+## Install
+
+```bash
+npm install @kitbag/router
+```
+
+## 5 Canonical Patterns
+
+### 1. Define routes with `createRoute` and `addView`
+
+```ts
+import { createRoute, withParams } from '@kitbag/router'
+import Home from './views/Home.vue'
+import User from './views/User.vue'
+
+const home = createRoute({ name: 'home', path: '/' })
+  .addView(Home)
+
+const user = createRoute({
+  name: 'user',
+  path: withParams('/user/[userId]', { userId: Number }),
+})
+  .addView(User, {
+    props: (route) => ({ id: route.params.userId }),
+  })
+
+const routes = [home, user] as const
+```
+
+Always use `as const` on your routes array for correct type inference.
+
+### 2. Create and install the router
+
+```ts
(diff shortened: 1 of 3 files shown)


pleek91:
@stackoverfloweth why both? I assumed just one would be sufficient to instruct agents.

stackoverfloweth:
> @stackoverfloweth why both? I assumed just one would be sufficient to instruct agents.

@pleek91 They serve different purposes, it's typical to see projects have both

**AGENTS.md** is the quick-reference/overview — an agent loads it unconditionally when it sees @kitbag/router in the project's dependencies. It's short and focused on "don't get this wrong."

**SKILL.md** is the on-demand deep reference — agents only load it when the task at hand matches the frontmatter description. It's much more detailed but only triggered when relevant.

craig-pplx (commented):
Reviewed the documented forms against the source on `main`. The guidance is broadly accurate, but a handful of code forms don't exist or are inverted — which matters more than usual here since agents copy these verbatim. Inline comments cover 6 outright mistakes and 4 smaller items.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+const protectedRoute = createRoute({
+  name: 'dashboard',
+  path: '/dashboard',
+})
+  .addView(Dashboard)
+
+// Route-level before-hook — use `throw reject(...)` to stop execution
+protectedRoute.onBeforeRouteEnter((to, { reject, replace }) => {
+  if (!isAuthenticated()) {
+    throw reject('NotFound')
```
craig-pplx: `throw reject(...)` is wrong — `reject` already throws internally, so the `throw` does nothing.

```ts
// src/services/createRouterCallbackContext.ts:35
const reject: RouterCallbackContext['reject'] = (type) => {
  throw new ContextRejectionError(type)
}
```

It's typed `(type: TSource) => void` (`src/types/routerReject.ts`), so `throw reject('NotFound')` is literally throwing a `void` expression — it only "works" because `reject` throws first. Our own docs and tests use the bare call ([rejections.md#L63](docs/advanced-concepts/rejections.md), `src/tests/routeLoaders.spec.ts:202` → `.addLoader((_route, { reject }) => reject('NotFound'))`).

Commit 1 had this right; commits 2–3 changed every example to the `throw` form. Suggest reverting all of them: here, the table row on line 220, SKILL.md lines 312–315 and 331, and the Common Mistakes bullet at SKILL.md:349. Same applies to `abort()` and `push()` in hook context — they throw too, so no `throw` needed.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+| `createRoute({ component: Foo })` | `createRoute({ ... }).addView(Foo)` |
+| `createRoute({ props: { ... } })` | `.addView(Foo, { props: (route) => ({ ... }) })` |
+| `router.addRoute(...)` | Routes are static — pass all routes to `createRouter(routes)` |
+| `useRoute().params.value.id` | `useRoute().params.id` (params is already reactive, not a ref) |
+| `<router-link to="/path">` | `<router-link :to="(resolve) => resolve('name', params)">` (callback style) |
+| `<router-link :to="{ name: 'foo' }">` | `<router-link :to="(resolve) => resolve('foo')">` |
+| `router.beforeEach(...)` | Use `onBeforeRouteEnter` on the route or pass hooks to `createRouter` |
+| `{ path: '/user/:id' }` | `{ path: withParams('/user/[id]', { id: Number }) }` (square brackets, explicit type) |
+| `meta: { requiresAuth: true }` inline | `createRoute({ meta: { requiresAuth: true } })` (same place, but access differs) |
+| `reject('NotFound')` | `throw reject('NotFound')` (`throw` is needed to stop hook execution) |
```
craig-pplx: This row inverts the correct form — `reject('NotFound')` is right and `throw reject('NotFound')` is the form to avoid (see the comment on the hook example above). As written, this table teaches the anti-pattern to exactly the audience the file exists to correct.

If a row is useful here at all, it could be the opposite direction:

| `return next(false)` / `next('/login')` | `reject('NotFound')` (throws internally, no `return`/`throw` needed) |

skills/kitbag-router/SKILL.md
```diff
@@ -0,0 +1,349 @@
...
+  routes: [loginRoute],
+  rejections: [AuthRequired],
+  onBeforeRouteEnter: [(_to, { reject }) => {
+    if (!isAuthenticated()) throw reject('AuthRequired')
+  }],
+})
+
+const router = createRouter(routes, {
+  plugins: [authPlugin],
+})
```
craig-pplx: `plugins` isn't a `RouterOptions` key — it's the third positional argument:

```ts
// src/services/createRouter.ts:77
(routes: TRoutes, options?: TOptions, plugins?: TPlugin[]): Router<TRoutes, TOptions, TPlugin>
```

So this should be:

```ts
const router = createRouter(routes, { rejections: [AuthRequired] }, [authPlugin])
```

As written it's a type error (excess property on `TOptions`), and the plugin's routes/rejections/hooks would never be registered.

skills/kitbag-router/SKILL.md
```diff
@@ -0,0 +1,349 @@
...
+```ts
+import { createRouterPlugin } from '@kitbag/router'
+
+const authPlugin = createRouterPlugin({
+  routes: [loginRoute],
+  rejections: [AuthRequired],
+  onBeforeRouteEnter: [(_to, { reject }) => {
+    if (!isAuthenticated()) throw reject('AuthRequired')
+  }],
+})
```
craig-pplx, line 341: `createRouterPlugin` doesn't accept hooks in its options. `CreateRouterPluginOptions` is only `{ routes?, rejections? }` (`src/types/routerPlugin.ts:25`); hooks are registered as methods on the returned plugin, matching [plugins.md](docs/advanced-concepts/plugins.md):

```ts
const authPlugin = createRouterPlugin({
  routes: [loginRoute],
  rejections: [AuthRequired],
})

authPlugin.onBeforeRouteEnter((to, { reject }) => {
  if (!isAuthenticated()) {
    reject('AuthRequired')
  }
})
```

skills/kitbag-router/SKILL.md
```diff
@@ -0,0 +1,349 @@
...
+link.isMatch // boolean ref
+link.isExactMatch // boolean ref
+link.push() // navigate
+```
+
+### useQueryValue
+
+```ts
+import { useQueryValue } from '@kitbag/router'
+const tab = useQueryValue('tab') // Ref<string | undefined>
```
craig-pplx: Wrong return type. `useQueryValue` returns an object, not a ref (`src/compositions/useQueryValue.ts:9-13`):

```ts
type UseQueryValue<T> = {
  value: Ref<T>,
  values: Ref<NonNullable<T>[]>,
  remove: () => void,
}
```

and the value is `string | null`, not `string | undefined`. Matching [docs/composables/useQueryValue.md](docs/composables/useQueryValue.md):

```ts
const { value: tab } = useQueryValue('tab')
//              ^? Ref<string | null>

const { values: tabs } = useQueryValue('tab', Number)
//              ^? Ref<number[]>
```

Worth mentioning the optional second `param` argument and `remove()` here too, since that's the whole reason to use this over `route.query`.

skills/kitbag-router/SKILL.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+## Param Syntax
+
+Params use **square brackets**, not colons:
+
+- `[id]` — required param (matches one or more characters including `/`)
+- `[?id]` — optional param
+- `[id*]` — greedy (explicitly captures multiple segments)
+- `[?id*]` — optional greedy
```
craig-pplx: The first two bullets describe greedy behavior for a plain param, which inverts the distinction. Per [params.md#greedy-params](docs/core-concepts/params.md): "By default, a path param matches only up to the next `/` (a single path segment)" — `[date]` matches `/photos/2026` but not `/photos/2026/01`; only `*` crosses segments.

```
- `[id]` — required param, matches a single path segment (stops at `/`)
- `[?id]` — optional param, type is `string | undefined`
- `[id*]` — greedy, matches across segments including `/`
- `[?id*]` — optional greedy
```

This one is likely to produce broken routes: an agent that believes `[id]` spans slashes won't reach for `[id*]` when it's actually needed.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+    throw reject('NotFound')
+  }
+})
+
+// Route-level after-hook — second arg includes `push` for navigation
+protectedRoute.onAfterRouteEnter((to, { push }) => {
+  if (to.params.search === 'secret') {
+    push('hiddenRoute')
+  }
+})
```
craig-pplx, line 130: This example doesn't type-check. `protectedRoute` is `/dashboard` with no params, so `to.params.search` doesn't exist, and `'hiddenRoute'` isn't a route defined anywhere in the file. Since these files exist to be copied verbatim, the examples should compile against their own surrounding context — either give `protectedRoute` the param it reads, or demo `push` against `'home'`/`'user'` which are already defined above.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+
+LLMs trained on older data may generate outdated patterns. The correct forms are:
+
+| Wrong (old/never existed) | Correct |
+|---|---|
+| `createRoute({ component: Foo })` | `createRoute({ ... }).addView(Foo)` |
+| `createRoute({ props: { ... } })` | `.addView(Foo, { props: (route) => ({ ... }) })` |
+| `router.addRoute(...)` | Routes are static — pass all routes to `createRouter(routes)` |
+| `useRoute().params.value.id` | `useRoute().params.id` (params is already reactive, not a ref) |
+| `<router-link to="/path">` | `<router-link :to="(resolve) => resolve('name', params)">` (callback style) |
```
craig-pplx: `<router-link to="/path">` is supported, not wrong. `RouterLinkProps.to` is `UrlString | ResolvedRoute | ToCallback<TRouter>` (`src/types/routerLink.ts:13`), and the component handles the string case directly (`isUrlString(to) ? router.find(to) : to`). It's also the only option for external or otherwise arbitrary URLs.

Suggest reframing as a preference rather than a prohibition — something like "prefer the callback form for internal routes so names and params are type-checked; plain URL strings and `ResolvedRoute` values are also accepted." Same for the Common Mistakes bullet at SKILL.md:344, which currently tells agents a valid API is off-limits.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+3. **`useRoute('name')` narrows types** — only use this inside components you know are mounted under that route. Using the wrong name gives a runtime error.
+4. **Param names must be unique** across the full route tree (including parent routes). Duplicate param names throw `DuplicateParamsError`.
+
+## APIs That Changed — Don't Write the Old Form
+
+LLMs trained on older data may generate outdated patterns. The correct forms are:
+
+| Wrong (old/never existed) | Correct |
+|---|---|
+| `createRoute({ component: Foo })` | `createRoute({ ... }).addView(Foo)` |
```
craig-pplx: "never existed" isn't accurate for the first two rows — `component` and `components` are still on `CreateRouteOptions`, just `@deprecated` in favor of `addView` (`src/types/createRouteOptions.ts:93-104`), and `createRejection` uses `createRoute({ name, component })` internally.

Retitling the column to "Deprecated / old form" would avoid an agent concluding that existing code using `component:` is broken and needs rewriting.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
...
+    <transition name="fade" mode="out-in">
+      <component :is="component" />
+    </transition>
+  </template>
+</router-view>
+```
+
+## Type Inference Gotchas
+
+1. **You must register the router** via declaration merging — without this, composables like `useRoute` and `useRouter` have no type context.
```
craig-pplx: Commit 3 dropped the `createRouterAssets` mention, which loses the one case where declaration merging isn't enough: with `isGlobalRouter: false` the library's exported composables and components don't bind to your router at all, and `createRouterAssets` is required (`src/types/router.ts:56-62` — "When false, createRouterAssets must be used for component and hooks"). Worth keeping a clause for it, since an agent hitting a multi-router setup has no other path from these docs.

AGENTS.md
```diff
@@ -0,0 +1,233 @@
+# @kitbag/router - Agent Guide
+
+Type-safe router for Vue 3. Not a fork of vue-router — different API, different mental model.
+
+Docs: https://kitbag.dev/router
```
craig-pplx, line 6: Scope question on the file's location rather than its content: by convention a root `AGENTS.md` is read by an agent working *in this repo*, but this one is entirely consumer usage docs — nothing about `npm run test` / lint / the docs site / how `src/` is laid out. An agent asked to fix a bug in kitbag-router itself will load this and get an API tour instead of contributor instructions.

Two options: add a short "Working in this repo" section (test/lint/build commands, where docs live) alongside the usage guidance, or state in the opening line that the file documents library usage for consumers of the npm package. Either keeps the tarball benefit while making the repo-local read unambiguous.

AGENTS.md
stackoverfloweth: I thought it still made sense to `throw reject` so that TS knows nothing past that will get executed

craig-pplx: Doesn't reject have return type never which does that automatically?

PR #818, opened 2026-08-16
Allow state shadowing between parent and child routes

stackoverfloweth:
## Summary

- Parent and child routes can now define state with the same key name (previously threw `DuplicateParamsError`)
- Child's state definition overrides the parent's using `Omit<TParent, keyof TChild> & TChild` instead of intersection
- State values are stored per-match in history (array of per-match records) rather than flat, so each route level retains its own values
- `useRoute('parentName')` scopes state to that match level — the runtime value matches the TypeScript type
- Leaf matches still return `router.route` directly (preserving reference equality)

## Test plan

- [x] Existing tests pass (882 passed, 1 expected fail)
- [x] New unit tests for `combineState` with duplicate keys
- [x] New unit tests for `setMatchStates`, `getMatchStates`, `resolveMatchStates`
- [x] New integration tests for shadowed state in router push
- [x] New test for `createRoute` with shadowed state key
- [x] Type checking passes with no errors

### DIFF
--- a/src/types/route.ts
+++ b/src/types/route.ts
@@ -63,15 +63,21 @@ type RouteMetaOf<TMatches extends CreatedRouteOptions[]> = CreatedRouteOptions[]
     : {}
 
 /**
- * The state of every matched route, combined. Mirrors `combineState`, which is an intersection.
- * A match without state contributes `{}` rather than `ToState<undefined>`, which widens to
- * `Record<string, Param>` and would swallow the rest of the intersection.
+ * The state of every matched route, combined. Later matches (children) override earlier ones
+ * (parents) when they share a key, mirroring `combineState` which uses spread semantics.
  */
 type RouteStateOf<TMatches extends CreatedRouteOptions[]> = CreatedRouteOptions[] extends TMatches
   ? Record<string, Param>
   : TMatches extends [infer THead, ...infer TRest extends CreatedRouteOptions[]]
-    ? (THead extends { state: infer TState extends Record<string, Param> } ? ToState<TState> : {}) & RouteStateOf<TRest>
+    ? Omit<THead extends { state: infer TState extends Record<string, Param> } ? ToState<TState> : {}, RouteStateKeysOf<TRest>> & RouteStateOf<TRest>
     : {}
+
+/**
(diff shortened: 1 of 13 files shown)


src/types/routerRoute.ts
```diff
@@ -39,6 +39,7 @@ export type RouterRoute<TRoute extends ResolvedRoute = ResolvedRoute> = {
 
   params: TRoute['params'],
   state: TRoute['state'],
+  matchStates: TRoute['matchStates'],
```
pleek91, line 42: why is this new property needed?

src/types/route.ts
```diff
@@ -63,16 +63,22 @@ type RouteMetaOf<TMatches extends CreatedRouteOptions[]> = CreatedRouteOptions[]
...
- * A match without state contributes `{}` rather than `ToState<undefined>`, which widens to
- * `Record<string, Param>` and would swallow the rest of the intersection.
+ * The state of every matched route, combined. Later matches (children) override earlier ones
+ * (parents) when they share a key, mirroring `combineState` which uses spread semantics.
  */
 type RouteStateOf<TMatches extends CreatedRouteOptions[]> = CreatedRouteOptions[] extends TMatches
   ? Record<string, Param>
   : TMatches extends [infer THead, ...infer TRest extends CreatedRouteOptions[]]
-    ? (THead extends { state: infer TState extends Record<string, Param> } ? ToState<TState> : {}) & RouteStateOf<TRest>
+    ? Omit<THead extends { state: infer TState extends Record<string, Param> } ? ToState<TState> : {}, RouteStateKeysOf<TRest>> & RouteStateOf<TRest>
```
pleek91, line 72: I'm a little unsure about this. I like the idea but I wonder if it creates confusion at the child level. Like when parent and child both have name now the child cannot access the parent's name state. But you can access the parents description state if the child doesn't have that same key. 

I'm not sure either way but did you consider if the child should ONLY have access to its own state? Like make state not inherited. 

Idk if that's a good idea, but something I want to think about

pleek91 (commented):
I think this sets a precedent for loaders and whatever we do here we'll want to make sure its compatible and feels right for both features. 

I think `meta` also falls into this same pattern?

PR #826, opened 2026-08-29
Render pending async props when server rendering

pleek91:
# Description

A view with an async props getter rendered nothing when server rendered. The wrapper its pending props go through was chosen by reading the component instance's `suspense` property, which is client only and always false during SSR, so the server always got the wrapper that fills a ref from a promise handler — and the server renderer never awaits that.

It now picks the awaiting wrapper when vue's ssr context is present. The server renderer awaits async setup whether or not a `<Suspense>` boundary exists, so this needs nothing from the app. Client behaviour is unchanged.

```ts
// renderToString: "" before, "<div>hello craig</div>" after
createRoute({ name: 'user', path: '/' })
  .addView(UserPage, { props: async () => ({ user: await getUser() }) })
```

`import.meta.env.SSR` is not usable for this — it would be replaced when this package is built rather than when an app using it is.

### DIFF
--- a/src/services/component.browser.spec.ts
+++ b/src/services/component.browser.spec.ts
@@ -3,7 +3,6 @@ import { expect, test } from 'vitest'
 import echo from '@/components/echo'
 import { createRoute } from '@/services/createRoute'
 import { createRouter } from '@/services/createRouter'
-import { h, defineComponent, getCurrentInstance } from 'vue'
 
 test('renders component with sync props', async () => {
   const route = createRoute({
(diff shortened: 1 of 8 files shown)


src/services/component.browser.spec.ts
```diff
@@ -64,11 +65,10 @@ test('renders component with async props', async () => {
   expect(wrapper.html()).toBe('echo')
 })
 
-test('component instance has suspense property when suspense is used', async () => {
+test('useHasSuspenseBoundary reports whether suspense is used', async () => {
```
craig-pplx: This test should move to a dedicated test file for the hook its testing.

pleek91:
Moved the `useHasSuspenseBoundary` test to `src/compositions/useHasSuspenseBoundary.browser.spec.ts` — it was only in `component.browser.spec.ts` because that's where the code used to live.

stackoverfloweth (approved):

PR #827, opened 2026-08-29
Require a rejection to declare an http status

pleek91:
# Description

A server rendering the app needs to know what to respond with, and the rejection in effect is the thing that knows.

`status` is **required** rather than defaulted, because no default suits every rejection — 404 for something missing, 401 or 403 for something gated, 503 for something temporary. A client only app never reads it, but the decision has to live somewhere and the rejection is the only place with enough context. The built in `NotFound` declares 404.

```ts
const authNeeded = createRejection({
  type: 'AuthNeeded',
  status: 401,
})
```

Breaking: every `createRejection` call needs a status. Nothing reads it yet — `router.render` does, later in the stack.

### DIFF
--- a/src/services/createRejection.ts
+++ b/src/services/createRejection.ts
@@ -1,17 +1,14 @@
 import { createRejectionHooks } from '@/services/createRejectionHooks'
 import { genericRejection } from '@/components/rejection'
 import { RejectionHooks } from '@/types/hooks'
-import { IS_REJECTION_SYMBOL, Rejection, RejectionInternal } from '@/types/rejection'
-import { Component, markRaw } from 'vue'
+import { IS_REJECTION_SYMBOL, Rejection, RejectionInternal, RejectionOptions } from '@/types/rejection'
+import { markRaw } from 'vue'
 import { createRoute } from '@/services/createRoute'
 import { RouteSetTitle } from '@/types/routeTitle'
 
-export function createRejection<TType extends string>(options: {
-  type: TType,
-  component?: Component,
-}): Rejection<TType> & RejectionHooks<TType> & RouteSetTitle
+export function createRejection<TType extends string>(options: RejectionOptions<TType>): Rejection<TType> & RejectionHooks<TType> & RouteSetTitle
 
-export function createRejection({ type, component }: { type: string, component?: Component }): Rejection {
+export function createRejection({ type, component, status }: RejectionOptions): Rejection {
   const { store, ...hooks } = createRejectionHooks()
 
   const route = createRoute({
@@ -29,6 +26,7 @@ export function createRejection({ type, component }: { type: string, component?:
 
   const rejection = {
     type,
+    status,
     setTitle,
     ...hooks,
     ...internal,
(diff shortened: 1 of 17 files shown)


src/services/createRejection.ts
```diff
@@ -9,9 +9,13 @@ import { RouteSetTitle } from '@/types/routeTitle'
...
   type: TType,
   component?: Component,
+  /**
+   * Suggested HTTP status when this rejection is in effect. Read by `router.start()`.
+   */
+  status?: number,
 }): Rejection<TType> & RejectionHooks<TType> & RouteSetTitle
 
-export function createRejection({ type, component }: { type: string, component?: Component }): Rejection {
+export function createRejection({ type, component, status }: { type: string, component?: Component, status?: number }): Rejection {
```
craig-pplx: We should write a dedicated type for `RejectionOptions`

src/services/getRoutesForRouter.ts
```diff
@@ -24,7 +24,7 @@ export function getRoutesForRouter(routes: Routes | Routes[], plugins: RouterPlu
   ]
 
   const allRejections = [
-    ...BUILT_IN_REJECTION_TYPES.map((type) => createRejection({ type })),
+    ...BUILT_IN_REJECTION_TYPES.map((type) => createRejection({ type, status: BUILT_IN_REJECTION_STATUS[type] })),
```
craig-pplx: Rather than having two conts for the rejections and their type we should have one const that defines both

pleek91:
Both addressed: `RejectionOptions` is now a named exported type that `createRejection` takes, and the two constants collapsed into one so they can't drift.

```ts
export const BUILT_IN_REJECTIONS = { NotFound: 404 } as const
export type BuiltInRejectionType = keyof typeof BUILT_IN_REJECTIONS
```

src/services/createRouter.ts
```diff
@@ -155,10 +156,17 @@ export function createRouter<
...
+      // than a successful navigation, even though the hooks had nothing to object to.
       case 'SUCCESS':
         history.update(url, options)
-        clearRejection()
+
+        if (matched) {
+          clearRejection()
+        } else {
+          setRejection('NotFound', to, from)
+        }
```
stackoverfloweth: this change feels risky but also feels like it should always have worked this way 😅

stackoverfloweth (approved):

src/services/createRouter.ts
stackoverfloweth: tests pass, including additional router-preview playwright tests

pleek91:
Your two comments here were on a change that shouldn't have been in this PR at all — closing #834 removed the PR but left its commit in the branches above it, so the `setRejection` inside `SUCCESS` was showing up in this diff.

That's fixed: the commit is dropped and replaced by #838, which routes an unmatched url through the existing `REJECT` branch instead. This PR is now just the rejection status.

Thanks for running the router-preview Playwright tests against it.

pleek91:
Following up on the `200` default discussion from #829 — the default is gone and `status` is now required.

The short version of why: every candidate default is wrong somewhere. `200` gets the rejection page indexed and cached as valid, `500` pages on-call for routine auth gates, and `404` deindexes a temporary maintenance page that should have been a `503`. Requiring it moves the decision to the only place with enough context to make it.

The cost is real — most apps don't server render and won't read the status — but it's one number per rejection definition, and the alternative was a silent wrong answer.

PR #828, opened 2026-08-29
Add a tracker for outstanding asynchronous work

pleek91:
# Description

Waiting for the router to go quiet means waiting for work registered *while* earlier work settles: a props getter or loader that pushes starts a navigation whose own values must also settle. So `idle` loops rather than awaiting once.

Nothing uses it yet — `router.render` does, next in the stack.

### DIFF
--- a/src/services/createActivityTracker.ts
+++ b/src/services/createActivityTracker.ts
@@ -1,0 +1,37 @@
+export type ActivityTracker = {
+  add: (...work: Promise<unknown>[]) => void,
+  wrap: <TArgs extends unknown[], TResult>(fn: (...args: TArgs) => Promise<TResult>) => (...args: TArgs) => Promise<TResult>,
+  idle: () => Promise<void>,
+}
+
+export function createActivityTracker(): ActivityTracker {
+  const outstanding = new Set<Promise<unknown>>()
+
+  const add: ActivityTracker['add'] = (...work) => {
+    work.forEach((promise) => {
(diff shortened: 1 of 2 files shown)


src/services/createActivityTracker.ts
```diff
@@ -0,0 +1,39 @@
+export type ActivityTracker = {
+  add: (...work: Promise<unknown>[]) => void,
+  wrap: <TArgs extends unknown[], TResult>(fn: (...args: TArgs) => Promise<TResult>) => (...args: TArgs) => Promise<TResult>,
+  idle: () => Promise<void>,
+}
+
+export function createActivityTracker(): ActivityTracker {
```
stackoverfloweth, line 7: should we have a way to clear this manually? like when a rote changes before the previous was finished?

pleek91:
> should we have a way to clear this manually? like when a route changes before the previous was finished?

Turns out it already self-clears. I tested a navigation superseded while its loader never resolves at all, and `render()` still completes: `dispose(NavigationAbandonedError)` rejects the pending promises, they settle, and the `finally` in `add` removes them.

The real gap is a different one — a loader that hangs with *no* subsequent navigation waits forever, since nothing disposes it. A manual clear wouldn't help there either, because the caller has no way to know it should call it. That wants a timeout plus `AbortSignal` plumbing into loaders, which is its own piece of work.

stackoverfloweth (approved):

PR #829, opened 2026-08-29
Add router.render for waiting on a navigation's data

pleek91:
# Description

Awaiting `push` only waits for the route to commit, and props and loaders deliberately don't hold up a navigation. That leaves a server with no point at which the page is renderable, and no way to learn that a props getter or loader redirected or rejected. Without waiting it renders markup for a route it has already left; with something awaiting the data, the abandoning navigation rejects that promise and it renders nothing at all.

`router.render` resolves once there is nothing left to render, following any navigation the data caused, and reports what a server should respond with. `start` no longer returns anything.

```ts
await router.start()

const response = await router.render()

if (response.status >= 300) {
  return respond(response)
}

const html = await renderToString(app)
```

The response is derived rather than recorded during navigation because the ordinary 404 sets no rejection at all — it commits the `NotFound` route — so reading the rejection alone would report 200 for most of them.

### DIFF
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -2,6 +2,10 @@ import { createPath } from '@/services/history'
 import { App, ref } from 'vue'
 import { createCurrentRoute } from '@/services/createCurrentRoute'
 import { createIsExternal } from '@/services/createIsExternal'
+import { createActivityTracker } from '@/services/createActivityTracker'
+import { getResponse } from '@/services/getResponse'
+import { RenderInBrowserError } from '@/errors/renderInBrowserError'
+import { isBrowser } from '@/utilities/isBrowser'
 import { parseUrl, updateUrl } from '@/services/urlParser'
 import { createRouteValueStore, RouteValueResponse } from '@/services/createRouteValueStore'
 import { DataKind } from '@/services/createNavigationStores'
@@ -11,7 +15,7 @@ import { getInitialUrl } from '@/services/getInitialUrl'
 import { setStateValues } from '@/services/state'
 import { Routes } from '@/types/route'
 import { NOT_FOUND_REJECTION_TYPE } from '@/types/rejection'
-import { Router, RouterOptions } from '@/types/router'
+import { Router, RouterOptions, RenderOutcome } from '@/types/router'
 import { RouterPush, RouterPushOptions } from '@/types/routerPush'
 import { RouterReplace, RouterReplaceOptions } from '@/types/routerReplace'
 import { RoutesName } from '@/types/routesMap'
@@ -91,6 +95,8 @@ export function createRouter<
   const isGlobalRouter = options?.isGlobalRouter ?? true
   const routerKey = isGlobalRouter ? routerInjectionKey : Symbol()
   const shouldRemoveTrailingSlashes = options?.removeTrailingSlashes ?? true
+  const redirectStatus = options?.redirectStatus ?? 302
+  const activity = createActivityTracker()
   const { routes, getRouteByName, getRejectionByType } = getRoutesForRouter(routesOrArrayOfRoutes, plugins, options)
   const notFoundRejection = getRejectionByType('NotFound')
   const valueStore = createRouteValueStore()
@@ -120,7 +126,7 @@ export function createRouter<
     return getMatchForUrl(filteredRoutes, url, { ...resolveOptions, ...parseOptions })
   }
 
-  async function set(url: string, options: RouterUpdateOptions = {}): Promise<void> {
+  const set = activity.wrap(async (url: string, options: RouterUpdateOptions = {}): Promise<void> => {
     if (pathHasTrailingSlash(url) && shouldRemoveTrailingSlashes) {
       const cleanedUrl = removeTrailingSlashesFromPath(url)
 
@@ -199,13 +205,15 @@ export function createRouter<
     } finally {
       history.startListening()
     }
-  }
+  })
 
   function setRouteValuesAndUpdateRoute(to: ResolvedRoute, from: ResolvedRoute | null): void {
     const { props, loaders } = valueStore.setRouteValues(to)
 
-    handleRouteValueResponse(props, 'props', to, from)
-    handleRouteValueResponse(loaders, 'loader', to, from)
+    activity.add(
+      handleRouteValueResponse(props, 'props', to, from),
+      handleRouteValueResponse(loaders, 'loader', to, from),
+    )
 
     updateRoute(to)
   }
@@ -214,8 +222,8 @@ export function createRouter<
    * Props and loaders are handled the same way, and neither is awaited here: a push or a rejection from
    * either is acted on whenever it arrives, without holding up the navigation that started it.
    */
-  function handleRouteValueResponse(response: Promise<RouteValueResponse>, source: DataKind, to: ResolvedRoute, from: ResolvedRoute | null): void {
-    response
+  function handleRouteValueResponse(response: Promise<RouteValueResponse>, source: DataKind, to: ResolvedRoute, from: ResolvedRoute | null): Promise<void> {
+    return response
       .then((response) => {
         switch (response.status) {
           case 'SUCCESS':
@@ -380,6 +388,29 @@ export function createRouter<
     started.value = true
   }
 
+  /**
+   * Does not resolve until the router has finished everything a view needs to render completely, and
+   * reports the status a server should respond with.
+   *
+   * Only available on the server for ssr. Throws {@link RenderInBrowserError} when called in the client.
+   */
+  async function render(): Promise<RenderOutcome> {
+    if (isBrowser()) {
+      throw new RenderInBrowserError()
+    }
+
+    await start()
+    await activity.idle()
+
+    return getResponse({
+      initialUrl,
+      route: currentRoute,
+      rejection: currentRejection.value,
+      removeTrailingSlashes: shouldRemoveTrailingSlashes,
+      redirectStatus,
+    })
+  }
+
   function stop(): void {
     history.stopListening()
   }
@@ -435,6 +466,7 @@ export function createRouter<
     prefetch: options?.prefetch,
     start,
     started,
+    render,
     stop,
     key: routerKey,
     hasDevtools: false,
--- a/src/services/getResponse.ts
+++ b/src/services/getResponse.ts
@@ -1,0 +1,49 @@
+import { Rejection } from '@/types/rejection'
+import { ResolvedRoute } from '@/types/resolved'
+import { RedirectStatus, RenderOutcome } from '@/types/router'
+import { isSameUrl } from '@/services/urlParser'
+import { pathHasTrailingSlash } from '@/utilities/trailingSlashes'
+
+export type GetResponseContext = {
+  /**
+   * The url the router started on.
+   */
+  initialUrl: string,
+  /**
+   * The route the router settled on.
+   */
+  route: ResolvedRoute,
+  /**
+   * The rejection in effect, or null.
+   */
+  rejection: Rejection | null,
+  /**
+   * Whether the router removes trailing slashes.
+   */
+  removeTrailingSlashes: boolean,
+  /**
+   * The status to report when the url was normalized rather than redirected by the app.
+   */
+  redirectStatus: RedirectStatus,
+}
+
+/**
+ * What a server should respond with, given where the router settled.
+ */
+export function getResponse({ initialUrl, route, rejection, removeTrailingSlashes, redirectStatus }: GetResponseContext): RenderOutcome {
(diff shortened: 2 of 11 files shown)


src/services/createRouter.ts
```diff
@@ -355,6 +361,36 @@ export function createRouter<
...
+      return { status: 301, location: currentRoute.href, rejection: type }
+    }
+
+    if (rejection) {
+      return { status: rejection.status ?? 200, rejection: type }
+    }
+
+    if (currentRoute.id === notFoundRoute.id) {
+      return { status: 404, rejection: type }
+    }
```
craig-pplx: why do we need this? Won't `rejection` be set for this case?

src/services/createRouter.ts
```diff
@@ -376,6 +412,16 @@ export function createRouter<
...
 
+  /**
+   * Resolves once there is nothing left to render. `idle` follows cascades, so a loader that pushes is
+   * waited on through to the navigation it caused rather than resolving on the one it abandoned.
+   */
+  async function render(): Promise<RouterResponse> {
+    await activity.idle()
+
+    return getResponse()
+  }
```
craig-pplx: Since getResponse depends on initialUrl this method is really only useful and valid in an ssr context. Calling this in the client would always 302 whenever the route is not the initialUrl. So we should throw an error if this is called in the client.

src/services/createRouter.ts
```diff
@@ -355,6 +361,36 @@ export function createRouter<
...
   const { promise: initialize, resolve: initialized } = Promise.withResolvers<void>()
 
+  /**
+   * What a server should respond with, given where the router currently is. Derived rather than recorded
+   * during navigation: the ordinary 404 sets no rejection at all, it just commits the NotFound route, so
+   * the rejection alone would report 200 for most of them.
+   *
+   * Deliberately avoids `find`, which resolves a route and so runs its title callback as a side effect.
+   */
+  function getResponse(): RouterResponse {
```
craig-pplx: We should move this to its own file, add unit tests, and then call the utility directly in render.

src/types/router.ts
```diff
@@ -162,6 +180,14 @@ export type Router<
...
    */
   started: Ref<boolean>,
+  /**
+   * Resolves once the router has nothing left to render: every prop and loader has settled, following
+   * any navigation one of them caused, so this waits for the next *full* render rather than for one
+   * navigation. Resolves with what a server should respond with.
+   *
+   * Awaiting `push` only waits for the route to commit; this also waits for its data.
+   */
+  render: () => Promise<RouterResponse>,
```
craig-pplx: Since `render` should only be called in an ssr context. Can we just call vue's renderToString within our method and return the html, status, etc as part of RouterResponse?

pleek91:
Addressed the extraction and the client guard:

- `getResponse` moved to `src/services/getResponse.ts` with 10 unit tests covering each branch directly, and `render` calls it.
- `render` now throws `RenderInBrowserError` in the browser. The guard is `isBrowser()` rather than `useIsServerRendering()` because the latter needs `inject`, so it only works inside setup and `render` isn't a component. No browser test called `render`, so nothing broke.

Left the two open questions for discussion rather than guessing:

**The 404 branch** — a rejection is *not* set for an unmatched url. `set` does `find(url) ?? notFoundRoute` and then the SUCCESS branch calls `clearRejection()`, so you commit the NotFound route with `rejection: null`. Measured: `rejection = null`, `status = 404`. Remove the branch and the ordinary 404 reports 200. It reads better in its new home, where the check is a line of `getResponse` with the reason next to it.

**`renderToString` inside `render`** — replied separately, it has a measured bundle cost.

pleek91:
Closing out the two open threads per discussion:

**`renderToString` inside `render`** — not doing it. Measured cost is +15.5 KB ESM / +12.3 KB UMD inlined into every client app, since `external: ['vue']` matches exactly and doesn't cover the `vue/server-renderer` subpath. More importantly it would put `ssrContext` out of reach (`renderToString(app, ctx)` is how teleports and `useSSRContext()` data are collected, which the hydration payload will need) and foreclose streaming. `render` stays as the two-step.

**The 404 branch** — leaving as is for now. Worth recording why it exists: a url that matches nothing commits the NotFound route *without* setting a rejection, because `set` does `find(url) ?? notFoundRoute` and the SUCCESS branch then calls `clearRejection()`. Measured: `rejection: null`, `status: 404`. Drop the branch and the ordinary 404 reports 200. Making `set` set the rejection instead would let it go, but that changes what `useRejection()` reports, so it's its own change.

Also pushed a test that was missing: `wrap(set)` turned out to be load-bearing but uncovered — the existing cascade test passed with it removed, because nothing in it delayed `setRouteValues`. `waits for a cascading navigation that registers its data behind a slow hook` puts a slow `onBeforeRouteEnter` on the redirect target and fails without the wrapping.

src/types/router.ts
```diff
@@ -63,6 +63,24 @@ export type RouterOptions = {
   isGlobalRouter?: boolean,
 }
 
+/**
+ * What a server should respond with for what the router has rendered.
+ */
+export type RouterResponse = {
```
stackoverfloweth: I vote we pick a name here that's more tightly coupled to the SSR/render itself

stackoverfloweth:
```ts
export type RenderOutcome =
  | { status: 200 | 404; rejection: RejectionType | null }
  | { status: 301 | 302; location: string; rejection: RejectionType | null }
```

src/services/createRouter.ts
```diff
@@ -384,6 +393,25 @@ export function createRouter<
...
   }
 
+  /**
+   * Resolves once there is nothing left to render. `idle` follows cascades, so a loader that pushes is
+   * waited on through to the navigation it caused rather than resolving on the one it abandoned.
+   */
+  async function render(): Promise<RouterResponse> {
+    if (isBrowser()) {
+      throw new RenderInBrowserError()
+    }
```
stackoverfloweth, line 400: I agree that `render` should be reserved for server-side only, but I think there's value in surfacing the idle half of this. What do you think about separating and having `router.idle` that returns a simple `Promise<void>` for both clients?

src/services/createRouter.ts
```diff
@@ -384,6 +393,25 @@ export function createRouter<
...
+
```
stackoverfloweth, line 401: should we add a guard here to verify that `router.start` was called? maybe return `RouterNotInstalledError`?

src/types/router.ts
```diff
@@ -63,6 +63,24 @@ export type RouterOptions = {
...
+   */
+  status: number,
+  /**
+   * Value for the `Location` header. Present only when `status` is a redirect.
+   */
+  location?: string,
+  /**
+   * The type of rejection in effect, or null when there is none.
+   */
+  rejection: string | null,
```
stackoverfloweth: can we not have a better type here than `string`?

src/services/getResponse.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export function getResponse({ initialUrl, route, rejection, removeTrailingSlashes }: GetResponseContext): RouterResponse {
+  const type = rejection?.type ?? null
+
+  if (removeTrailingSlashes && pathHasTrailingSlash(initialUrl)) {
+    return { status: 301, location: route.href, rejection: type }
+  }
+
+  if (rejection) {
+    return { status: rejection.status ?? 200, rejection: type }
```
stackoverfloweth: I wonder if `200` is a sensible default for rejections

src/services/createRouter.ts
```diff
@@ -199,13 +204,17 @@ export function createRouter<
...
     const { props, loaders } = valueStore.setRouteValues(to)
 
-    handleRouteValueResponse(props, 'props', to, from)
-    handleRouteValueResponse(loaders, 'loader', to, from)
+    // The derived chains are tracked rather than the responses themselves: a push or rejection the
+    // response causes is new work that only exists once the response has been handled.
+    activity.add(
+      handleRouteValueResponse(props, 'props', to, from),
+      handleRouteValueResponse(loaders, 'loader', to, from),
+    )
```
stackoverfloweth, line 216: how does this surface errors?

src/services/getResponse.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Derived from the router's state rather than recorded during navigation, and deliberately without
+ * calling `find`, which resolves a route and so runs its title callback as a side effect.
+ */
+export function getResponse({ initialUrl, route, rejection, removeTrailingSlashes }: GetResponseContext): RouterResponse {
+  const type = rejection?.type ?? null
+
+  if (removeTrailingSlashes && pathHasTrailingSlash(initialUrl)) {
+    return { status: 301, location: route.href, rejection: type }
+  }
```
stackoverfloweth, line 38: Browsers and CDNs cache 301s indefinitely, so changing app-level redirect later leaves poisoned caches you can't recall. I wonder if we should use 302 here and/or give the developer a tool to choose between 301 and 302

pleek91:
Went through all seven. Five changes, two answers.

**Renamed to `RenderOutcome`, and `location` now only exists on a redirect.** Adjusted your sketch in one place — `status: 200 | 404` can't hold a custom rejection status, and `createRejection({ status: 401 })` is already tested. So:

```ts
export type RenderOutcome<TRejectionType extends string = string> =
  | { status: number, location?: undefined, rejection: TRejectionType | null }
  | { status: RedirectStatus, location: string, rejection: TRejectionType | null }
```

`if (outcome.location !== undefined)` narrows `status` to `301 | 302` and `location` to `string`, with no extra discriminant field. There's a `spec-d` test pinning it.

**`rejection` is typed properly now** — the same union `router.reject()` already accepts, so `'Unauthorized' | 'NotFound' | null` rather than `string | null`. Also covered by a `spec-d` test.

**Split out `router.idle(): Promise<void>`.** Agreed, and it fills a real gap: `render` throws in the browser, so the wait was unavailable on the client entirely. `idle` is available everywhere.

**`render` now awaits `start`** rather than guarding. Measured before the change: `render()` on an unstarted router returned `{ status: 200, rejection: null }` — a confident success for a router that had never navigated, so a server forgetting `start` shipped an empty page as a 200. `start` is idempotent, so awaiting it makes the mistake impossible instead of merely diagnosable. `RouterNotInstalledError` didn't fit since it's about installation rather than starting.

**Trailing-slash status is configurable, defaulting to 302.** Your reasoning won: `removeTrailingSlashes` is on by default, so a 301 default would bake unrecallable redirects into every visitor's browser for people who never thought about trailing slashes. `redirectStatus: 301` opts into permanent.

**"How does this surface errors?"** — through `onError`. A props getter that throws calls it exactly once; the tracker's `.catch(() => undefined)` only silences its own bookkeeping chain, leaving the original promise untouched.

That did surface something worth deciding separately though: **a thrown props getter reports status 200.** Same family as your question about `200` being the default for rejections, and I don't think either should be settled quietly.

**On `200` as the rejection default** — I've left it, but the argument isn't airtight. For: the page rendered, it just rendered a different component; 500 would break working pages and 404 would be a guess. Against: an `Unauthorized` screen served as 200 is bad for crawlers and monitoring. The third option is making `status` **required** on `createRejection` so nobody defaults by accident — small breaking surface, and it forces the decision at the only place with enough context. Worth a call from you both rather than me picking.

pleek91:
Reversing one of these after discussion: **`router.idle` is not exposed.** The tracker stays, but `render` is its only consumer.

The reasoning that decided it isn't cost — I measured, and tracking is unmeasurable: **68.7µs vs 69.2µs** per navigation over 4000 navigations, with tracking nominally *faster*, i.e. below the noise floor. It adds three `Set` add/delete pairs and six promise allocations against a navigation that already does regex matching, url parsing, hook running and prop computation.

It's the API commitment. `idle` can be added later without breaking anyone; removing it later would be breaking. So shipping it on a reviewer's intuition rather than a user's report is the asymmetric bet.

Your instinct that the wait is useful beyond SSR is right though, and worth recording for when someone does ask. The real case is **app level testing** — `await router.push('x'); await router.idle(); expect(...)`. Without it an app author has to `flushPromises()` and hope, which doesn't drain timers at all. That trap caught me three separate times writing the tests in this PR, with full visibility into the internals: a props getter resolving in a microtask had already settled during `start`'s own awaits, so tests passed for the wrong reason and only a real timer exposed the difference.

If you think that outweighs the commitment, say so and I'll put it back — it's four lines and the tracking already supports it.

Related: the tracking is deliberately *not* gated on SSR. The only signal available at router creation is `isBrowser()`, which is the same thing we rejected for the props wrapper because it isn't "is server rendering" — false in node tests, true for `renderToString` in a browser. Adding a mode built on a known-wrong signal to save noise wasn't worth it.

stackoverfloweth (approved):

PR #834, opened 2026-09-02
Reject rather than silently commit a route for a url that matched nothing

pleek91:
# Description

A url matching no route committed the `NotFound` route and *cleared* the rejection, so the most common rejection in any app was the one that never set one. `onRejection` never fired for a 404, `useRejection()` reported `null`, and the two ways of reaching the same NotFound screen — an unmatched url versus a hook calling `reject('NotFound')` — left completely different state behind.

Rendered output is unchanged: `RouterView` reaches the same component through the rejection instead of through the route.

`route` is still seeded with the NotFound route, so `useRoute` keeps working outside `RouterView`. That the seeded route isn't one of the caller's declared routes is a separate problem, noted below.

### DIFF
--- a/src/services/createRouter.spec.ts
+++ b/src/services/createRouter.spec.ts
@@ -1017,3 +1017,48 @@ describe('options.removeTrailingSlashes', () => {
     expect(router.route.href).toBe('/')
   })
 })
+
+describe('a url that matches no route', () => {
+  test('sets the NotFound rejection', async () => {
+    const onRejection = vi.fn()
+    const route = createRoute({ name: 'route', component, path: '/foo' })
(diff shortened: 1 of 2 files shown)


pleek91:
Closing — wrong shape. It patches the `SUCCESS` branch (which is self-contradictory: the hooks approved the navigation and then we reject it) and leaves `matched ?? notFoundRoute` intact, so it adds a rejection *beside* the fabricated route rather than removing it. The type hole it was meant to address survives untouched.

Superseded by a design discussion covering nullable `router.route`, `useRoute` throwing, and per-occurrence rejection data.

PR #838, opened 2026-09-08
Reject with NotFound when a url matches no route

pleek91:
# Description

A url matching nothing cleared the rejection and rendered the NotFound screen through the *route* rather than through the rejection. So the most common rejection in any app was the one that never set one: `onRejection` never fired for a 404 and `useRejection()` reported `null`, while a hook calling `reject('NotFound')` set both.

A successful hook response for an unmatched url now becomes a rejection, so it takes the same path. Hooks still run first, so one that redirects or rejects an unknown url wins over this.

Rendered output is unchanged — `RouterView` reaches the same component through the rejection instead of through the route.

## Hooks receive a null `to`

Rather than substituting the NotFound route as the destination, the hook runner accepts `to: null`. Enter and update hooks are keyed off the destination's matches, so for an unmatched url there are none to run. Leave hooks are keyed off the route being left, so they still run and now see `to` as `null`. `onBeforeRouteLeave` and `onAfterRouteLeave` are the only hook types whose `to` widens to `ResolvedRoute | null`.

Plugins get `PluginBeforeRouteLeaveHook` and `PluginAfterRouteLeaveHook`. The existing `PluginBeforeRouteHook` covered enter, update and leave, and widening it would have given plugin enter hooks a `null` they never receive.

An unmatched url has no params to update, so calling `update` from a leave hook throws the new `UpdateWithoutRouteError`.

`router.route` is unaffected and remains non-null, so `useRoute` still needs no guard.

Note the route committed for a 404 is still the NotFound route rather than one of the caller's declared routes, so `router.route.name` can be a name the caller never declared. That's a separate problem and not addressed here.

### DIFF
--- a/src/errors/updateWithoutRouteError.ts
+++ b/src/errors/updateWithoutRouteError.ts
@@ -1,0 +1,9 @@
+/**
+ * An error thrown when a hook calls update for a url that matched no route.
+ * @group Errors
+ */
+export class UpdateWithoutRouteError extends Error {
+  public constructor() {
+    super('update called without a route to update')
+  }
+}
--- a/src/main.ts
+++ b/src/main.ts
@@ -87,6 +87,7 @@ export { MetaPropertyConflict } from './errors/metaPropertyConflict'
 export { NavigationAbandonedError } from './errors/navigationAbandonedError'
 export { RouterNotInstalledError } from './errors/routerNotInstalledError'
 export { UseRouteInvalidError } from './errors/useRouteInvalidError'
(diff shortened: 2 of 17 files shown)


stackoverfloweth (approved):

pleek91:
Superseded by #844. This PR was marked merged automatically when `ssr-async-props` was pushed with this branch's commit in its history, making the head an ancestor of the base. The change itself is unmodified and now targets `main` in #844 so it can be reviewed and merged ahead of the SSR work.

PR #844, opened 2026-09-11
Reject with NotFound when a url matches no route

pleek91:
# Description

Replaces #838, which GitHub marked as merged when its head commit became an ancestor of its base branch. Stacked on #845, and ahead of the SSR work.

A url matching nothing cleared the rejection and rendered the NotFound screen through the *route* rather than through the rejection. So the most common rejection in any app was the one that never set one: `onRejection` never fired for a 404 and `useRejection()` reported `null`, while a hook calling `reject('NotFound')` set both.

A successful hook response for an unmatched url now becomes a rejection, so it takes the same path. Hooks still run first, so one that redirects or rejects an unknown url wins over this.

## Hooks receive a null `to`

Rather than substituting the NotFound route as the destination, the hook runner accepts `to: null`. Enter and update hooks are keyed off the destination's matches, so for an unmatched url there are none to run. Leave hooks are keyed off the route being left, so they still run and now see `to` as `null`. `onBeforeRouteLeave` and `onAfterRouteLeave` are the only hook types whose `to` widens to `ResolvedRoute | null`.

Plugins get `PluginBeforeRouteLeaveHook` and `PluginAfterRouteLeaveHook`. The existing `PluginBeforeRouteHook` covered enter, update and leave, and widening it would have given plugin enter hooks a `null` they never receive.

## `router.route` is no longer overwritten by a 404

The route committed for an unmatched url was the NotFound route, so `router.route.name` could be a name the caller never declared. Nothing rendered from it — `RouterView` reaches the rejection component through `rejection.route` — so it is no longer committed at all, and `router.route` holds the last route that actually matched. It remains non-null, so `useRoute` still needs no guard.

## Rejections set the document title

`setRejection` is folded into `reject`, which now takes the routes the rejection happened between. Every rejection path shares one implementation, so a rejection raised from a loader or an error hook updates the document title, which it previously never did.

The context argument is internal: `RouterRejectInternal` intersects the public `RouterReject` with a signature that accepts it, so `router.reject('NotFound', { to, from })` is still a compile error for callers.

## The history listener is always restored

`set` stopped listening to history at the top and only restarted at the very end, so any early return left the router deaf to back/forward. `ABORT` already had this bug on `main`. The body is now wrapped in `try`/`finally`.

### DIFF
--- a/src/services/createRouter.ts
+++ b/src/services/createRouter.ts
@@ -10,6 +10,7 @@ import { createRouterHooks, getRouterHooksKey } from '@/services/createRouterHoo
 import { getInitialUrl } from '@/services/getInitialUrl'
 import { setStateValues } from '@/services/state'
 import { Routes } from '@/types/route'
+import { NOT_FOUND_REJECTION_TYPE } from '@/types/rejection'
 import { Router, RouterOptions } from '@/types/router'
 import { RouterPush, RouterPushOptions } from '@/types/routerPush'
 import { RouterReplace, RouterReplaceOptions } from '@/types/routerReplace'
@@ -22,7 +23,7 @@ import { RouterResolve, RouterResolveOptions } from '@/types/routerResolve'
 import { RouteNotFoundError } from '@/errors/routeNotFoundError'
 import { createResolvedRoute } from '@/services/createResolvedRoute'
 import { ResolvedRoute } from '@/types/resolved'
-import { RouterReject } from '@/types/routerReject'
+import { RejectContext, RouterRejectInternal } from '@/types/routerReject'
 import { EmptyRouterPlugin, RouterPlugin } from '@/types/routerPlugin'
 import { getRoutesForRouter } from './getRoutesForRouter'
 import { getGlobalHooksForRouter } from './getGlobalHooksForRouter'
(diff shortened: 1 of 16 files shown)


src/services/createRouter.ts
```diff
@@ -317,7 +321,7 @@ export function createRouter<
     return push(source, options)
   }
 
-  function setRejection(type: string, to: ResolvedRoute | null = null, from: ResolvedRoute | null = null): void {
+  const reject: RouterRejectInternal<TOptions['rejections'] | TPlugin['rejections']> = (type: string, { to = null, from = null }: RejectContext = {}) => {
```
pleek91, line 329: before we had `reject` and `setRejection`. And depending on how you used them sometimes the hooks would run and sometimes the documentTitle would update. Unified into a single function with an internal second argument for providing the to/from if the rejection happens during navigation.

src/services/createRouter.ts
```diff
@@ -132,64 +133,67 @@ export function createRouter<
 
     history.stopListening()
 
-    const to = find(url, options) ?? notFoundRoute
+    try {
```
pleek91, line 136: recommend viewing this with whitespace off

stackoverfloweth (approved):

PR #852, opened 2026-09-13
Support arktype params

pleek91:
# Description

Arktype implements standard schema, so validation already flows through the shared dispatcher — supporting it is a matter of teaching the dispatcher where arktype keeps its type information.

Arktype types describe sets rather than naming schemas: `string.email` is an intersection, a literal is a unit, `boolean` is a union of `true | false`. So instead of reading a type name, the adapter asks which set the schema is a subtype of through the public `extends` method, checking domains before fanning out unions so booleans and enum-like unions of literals coerce as their domain. A nice side effect is that literals coerce by their domain directly — `type('1')` parses `'1'` through the number coercion — rather than needing the try-everything approach the other vendors use.

Unions fan out through `branches`, which arktype exposes on every type.

## Callable schemas

Arktype types are functions, so the param getter check caught them before the standard schema check and invoked them as `(value, extras)` getters. The standard schema check now runs first, both in the value dispatch and in `ExtractParamType`, where the callable signature made `TParam extends ParamGetter` true and inferred the wrong output type.

Covered by the same round-trip table as zod and valibot — literals, formats, domains, `Date`, objects, arrays, tuples, unions in both directions, a discriminated union, `Record`, `Map`, `Set`, and `bigint` — plus type-level tests that `withParams` accepts arktype schemas and infers their output type.


stackoverfloweth (approved):
nice! we should update the docs at the bottom of [this page](https://router.kitbag.dev/core-concepts/params.html#zod-param-types)

pleek91:
Added an ArkType section to the params page, matching the zod and valibot ones. Also fixed the valibot paragraph while there — it said "some zod schemas" and named `v.intersection`, which is actually `v.intersect`. Intersections aren't listed as unsupported for arktype because it normalizes `.and()` into plain object types, so they work.

PR #855, opened 2026-09-14
Hydrate the client from what the server rendered

pleek91:
# Description

A server can render a page with `router.render`, but the client then computed everything again: every props getter and loader re-ran on hydration, and because the router starts asynchronously, vue hydrated against an empty `RouterView` and reported a mismatch on the first paint. This makes hydration adopt what the server rendered, with no work on the user's part.

## The payload

`render` now resolves with two more things: `title`, the resolved title of the route or rejection that rendered, and `payload`, a `<script type="application/json">` tag carrying the url, the rejection in effect if there was one, and what every props getter and loader settled on. The server embeds both in the html it sends. `<` and the unicode line separators are escaped as json unicode escapes, so nothing in a value can break out of the script tag and `JSON.parse` returns exactly what went in.

## Adoption

`start` looks for a payload — `install` calls `start` during `app.use`, before anything mounts. A payload is always adopted when present: silently falling back to a fresh navigation would leave a developer believing they have hydration when they do not, so a payload whose route no longer matches rejects with NotFound, visibly.

Adoption is the ordinary navigation: `hydrate` decodes the payload, prefills the value store, and runs `set` with a `hydrating` option that changes exactly two things — before hooks are not consulted, and the title the markup carries is kept. Before hooks are not consulted — they cannot change an outcome that has already been rendered — and the document title is left as the markup set it. After hooks never run for a rendering router — the `ssr` option (#857) marks every navigation as part of the server render from the moment the router is created, so there is no window where a navigation slips past the flag. The response returns what the navigation committed and nothing after that. Client side effects like analytics fire once, on the client, live or hydrated. A rejection payload never navigates at all: `hydrate` rejects directly through the same `reject` as a live rejection, so rejection hooks fire identically.

The outcome is committed synchronously within the `start` call, before its first await, so the first paint matches the server's markup. Tests pin the route, params, and `started` immediately after calling `start`, which is what catches anyone later adding an await ahead of the commit.

## Values cross as json

`JSON.stringify` and `JSON.parse` carry every value. A value json cannot carry is left out of the payload rather than sent corrupted, and its getter runs on the client as usual — never silently: the server warns and returns each one on the `render` response as `failures`, and the client warns for every value it expected but did not receive, naming the getter so the two sides can be matched up. #859 on top adds `stringify` and `parse` options for declaring how such a value crosses.

Schema params resolve synchronously at construction thanks to #849 removing the zod import, so a url with zod params adopts without waiting — pinned by a test.


stackoverfloweth (approved):
this PR is where I start to feel like supporting SSR isn't worth it. I'm going to approve because I feel like that decision shouldn't be made right here.

PR #857, opened 2026-09-14
Rename RenderOutcome to ServerRenderResponse and require the ssr option to render

pleek91:
# Description

What `render` resolves with is the material for the http response: the status, the `Location` header, and why. `RenderOutcome` is renamed `ServerRenderResponse` to say so:

```ts
const response = await router.render()

if (response.location) {
  return { status: response.status, location: response.location }
}

return { status: response.status, html: await renderToString(app) }
```

The union member that previously held the response name — the non-redirect half — becomes `RenderComplete`, distinguishing a render that finished here from one that points somewhere else.

## The ssr option

`render` was gated by sniffing the environment, which blocked browser runtimes from producing ssr builds. The `ssr` router option declares the intent instead — it marks the router as rendering on a server from the moment it is created, and `render` throws `SsrOptionRequiredError` without it:

```ts
  initialUrl: url,
  ssr: true,
})
```

`RenderInBrowserError` is gone with the environment check.

`render` merged in #829 and has not shipped in a release, so nothing public breaks.


stackoverfloweth (approved):
this makes sense to me, should we set history mode based on `options.ssr` as well?

src/tests/render.browser.spec.ts
```diff
@@ -1,14 +1,14 @@
 import { expect, test } from 'vitest'
 import { createRoute } from '@/services/createRoute'
 import { createRouter } from '@/services/createRouter'
-import { RenderInBrowserError } from '@/errors/renderInBrowserError'
 import { component } from '@/utilities/testHelpers'
+import { SsrOptionRequiredError } from '@/errors/ssrOptionRequiredError'
 
-test('throws in the browser, where the response it reports would be meaningless', async () => {
+test('render requires the router to be created for server rendering', async () => {
```
stackoverfloweth: should this test also exist on non-browser side?

pleek91: Hmm I think it just needs to be moved. Its not browser specific anymore.

pleek91:
> this makes sense to me, should we set history mode based on `options.ssr` as well?

The default is `auto` which chooses `browser` in the browser and `memory` in a node env. So I don't think we need to do anything here.

PR #861, opened 2026-09-15
Fall back to the current route's title and set it when the navigation commits

pleek91:
Falls back to the current route's title when a rejection doesn't declare one, and updates the title as soon as the navigation commits.

The current rejection and current route are the only sources of truth: a rejection with a title wins, otherwise the title reflects whatever route the router is on. A rejection in an after hook inherits the committed route's title; a rejection in a before hook never commits, so the title stays put.

The document title now updates when the navigation commits, before the after hooks run, so a slow after hook no longer leaves the old title up while the new route is already rendered. A rejection from an after hook still updates the title itself through `reject`.

- Rejections expose their own `getTitle`, so `currentRejectionRoute` is gone
- The router resolves title priority in one place and `setDocumentTitle` just applies it in the browser
- Adds a `GetTitleCallback` type for the shared `() => Promise<string | undefined>` shape


src/services/getResponse.spec.ts
```diff
@@ -45,28 +45,28 @@ test('given removeTrailingSlashes is false, does not report a normalization', ()
 })
 
 test('given the NotFound rejection, returns 404', () => {
-  const rejection: Rejection = { type: 'NotFound', status: 404 }
+  const rejection = createRejection({ type: 'NotFound', status: 404 })
```
stackoverfloweth, line 48: ❤️

stackoverfloweth (approved):
this cleans up a lot 👏

PR #866, opened 2026-09-16
Expose prefill and settled on the route value store

pleek91:
# Description

Hydration (#855) needs two things from the value store: hand it values the server already resolved, and read back what settled so the server can send them. This gives the store those two generic primitives, and moves selection logic that was never its business out:

- `prefill` stages values that already resolved — the next navigation adopts them in place of running their getters, riding the same staged-store path a followed link's prefetch uses
- `getValues` reads back the values the store currently holds
- `getComputations` moves to its own module, so anything can derive what a route computes from the route alone
- Choosing *which* props to prefetch for a strategy moves to `usePrefetching`, which already owned the strategy timing — the store's prefetch just starts what it is handed

No behavior changes.


stackoverfloweth (approved):
I love this PR, I wish we had better names for `Computed` and having `prefetch` property right next to `routePrefetch` feels like a miss but I don't have any better ideas

PR #878, opened 2026-09-19
Animate navigations with the View Transitions API

pleek91:
Adds a `viewTransition` option at the router, route and navigation level. When a navigation transitions, the router computes the destination's props and loaders into the staged store and loads its async components first, with the previous page still live, then wraps the commit in `document.startViewTransition`. The browser captures a finished page rather than a placeholder and nothing is frozen while data is fetched, which is how TanStack Router, React Router and SvelteKit sequence it too.

Types from every level combine and reach css through `:active-view-transition-type()`. A types callback sees `to` and `from` and can return `false` to skip.

The transition in flight is exposed as reactive state, `router.viewTransition` and the `useViewTransition` composable, from the moment a navigation is decided to transition until its animation finishes. The page being left can name the element a shared-element morph starts from before it is captured, and the page arriving can reach the browser's `ViewTransition` once it starts. Links derive `isTransitioning` from it, on `useLink` and the `RouterLink` slot, so a list names only the card being navigated to. Hooks are unchanged.

The router owns the one in-flight transition, so two navigations in one frame never call `startViewTransition` re-entrantly and a superseded navigation never commits. The promises a skipped transition rejects are handled here rather than surfacing as unhandled errors.

Off by default, since the `<Transition>` pattern on `RouterView` does not mix with it. Nothing changes for navigations that do not transition, on the server, on the first navigation or in a browser without the api.

Based on #879, which gives the value store the staged store this computes into.


docs/advanced-concepts/view-transitions.md
```diff
@@ -0,0 +1,187 @@
...
+A navigation that transitions loads everything the destination renders with before the transition starts: the route's [props](/core-concepts/component-props), its loaders and any [async components](/advanced-concepts/prefetching#prefetching-components). The previous page stays on screen and interactive while that happens, and the browser then captures the destination with its content rather than a placeholder. Navigations that do not transition commit immediately, as they always have.
+
+If props or a loader push somewhere else or reject, the navigation commits without a transition and the response is handled the same way it is without one.
+
+## What Does Not Transition
+
+- The first navigation, since there is no page to animate from.
+- Navigations on the server.
+- Navigations in a browser without `document.startViewTransition`. These navigate exactly as they do with the option off.
+- Navigations to a url no route matches.
```
pleek91: or navigations to external routes/urls

docs/api/types/RouterOptions.md
```diff
@@ -127,3 +127,17 @@ optional ssr?: boolean;
 
 Marks the router as rendering on a server, so every navigation is part of the server render from
 the moment the router is created. Required to call `render`.
+
+### viewTransition?
+
+```ts
+optional viewTransition?: ViewTransitionConfig;
```
pleek91: I think this should link to the type? Might not because the type isn't exported?

PR #882, opened 2026-09-20
Track the navigation under way with an abort signal

pleek91:
# Description

A navigation carried an id and asked whether it was still the current one after each await, and stopping the router made every id stale. An `AbortController` says the same thing more directly: beginning a navigation aborts the one before it, stopping the router aborts the one under way and every one begun after it, and the checks become `signal.aborted`.

The id also told the first navigation apart so it could leave from no route. That is really "has anything been committed", which is what `started` answers, so `started` now flips when the first route or rejection commits instead of when the initial navigation settles. A side effect worth knowing: the first paint no longer waits on the initial route's after hooks, matching every later navigation.

This sets up the next PR, which hands the signal to hooks, props getters and loaders so they can stop their own work.


stackoverfloweth (approved):
❤️

PR #883, opened 2026-09-20
Hand every hook, props getter and loader an abort signal

pleek91:
# Description

The router knows when a callback's outcome stopped mattering, but the callback had no way to find out, so a slow fetch in a hook or a loader ran to completion for a navigation that had moved on. Each callback's context now carries a `signal` to pass along.

```ts
route.addLoader(async (route, { signal }) => {
  const response = await fetch(`/api/users/${route.params.id}`, { signal })

  return response.json()
})
```

Hooks get the navigation's signal, which aborts when a newer navigation begins or the router stops. Props getters and loaders get their store's signal, which aborts when the store is disposed: when the next navigation commits, or when a link that prefetched into it goes away.

Stacked on #882.


stackoverfloweth (approved):
❤️❤️❤️

Issue #885, opened 2026-09-20
A root `/` route doesn't work?


KazimirPodolski:
I'm migrating from vue-router and currently I cannot get it to work. One issue I have is a `/` "umbrella" route for most of my routes which loads heavy stuff (e.g. PowerSync) doesn't seem to work:

```typescript
const main = createRoute({
    name: 'main',
    path: '/'
}).addView(defineAsyncComponent(() => import('@/views/MainView.vue')));

const dashboard = createRoute({
    parent: main,
    name: 'dashboard',
    path: '/dashboard'
}).addView(defineAsyncComponent(() => import('@/views/dashboard/DashboardView.vue')));
main.redirectTo(dashboard);

export const router = createRouter([
    main,
    dashboard
] as const, {
    historyMode: 'hash'
});

```

Now when I open `http://localhost:5173/` it somehow decides to redirect to `http://localhost:5173/#http://localhost:5173/`... And `MainView` is rendered, but `DashboardView` in it's `<RouterView/>` is not.

Repro: https://stackblitz.com/edit/vitejs-vite-nyb1qawn.

pleek91:
@KazimirPodolski Thanks for the repro — this surfaced two bugs, both fixed in 0.26.1

- Hash mode ignored the hash on load (#887)
- A redirect into its own subtree looped (#886)

You'll also need one config change. Child paths combine with the parent verbatim, no slash magic. So `/` + `/dashboard` is `//dashboard`. So you'll want to drop the leading `/` on your dashboard route's path. 

Could you update your config and router version and try again, and let us know if that fixes it?

KazimirPodolski:
Updated the repro to the new version https://stackblitz.com/edit/vitejs-vite-t7b8rkxw. It now doesn't go crazy in hash mode, but I still cannot get the `/` to behave properly:

<img width="1843" height="472" alt="Image" src="https://github.com/user-attachments/assets/c942aa93-ad56-40ea-9f39-2126b28eaec8" />

Now with `path: ''` for the main route opening `http://localhost:5173` immediately redirects to `http://localhost:5173/#/` and it's still NotFound (should be `http://localhost:5173/#/dashboard` and working).

Opening `http://localhost:5173/#/dashboard` manually works.

pleek91:
[...]

PR #891, opened 2026-09-21
Throw for a named route whose path does not start with a slash

pleek91:
# Description

Url pathnames always start with `/`, so a named route whose combined path does not start with `/` can never match a url. The router still resolves an href for it, so `push('main')` or a `router-link` to it lands on NotFound with no hint about why. #885 is this exact case: a root route with `path: ''` and a child with `path: '/dashboard'`.

The router now throws `UnreachableRouteError` at creation for any named, relative route whose path (after combining with its parents and the `base` option) does not start with `/`. Unnamed routes and external routes are not checked, since unnamed routes are never matched and external navigation is handed to the browser. Path combining itself is unchanged.

The routes, route matching, and vue-router migration docs get a short note stating the rule.


docs/core-concepts/routes.md
```diff
@@ -41,6 +41,10 @@ const home = createRoute({
 })
 ```
 
+::: info
+A named route's full path, its own path combined with any parents, must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```
stackoverfloweth: this sounds awkward to me, like it picks up mid-sentence.


```suggestion
A named route's final path (its own path combined with any parents) must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```

docs/advanced-concepts/route-matching.md
```diff
@@ -30,6 +30,10 @@ const external = createExternalRoute({
 
 Routes `path` must match the structure of the URL pathname.
 
+::: info
+A named route's full path, its own path combined with any parents, must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```
stackoverfloweth: this sounds awkward to me, like it picks up mid-sentence.


```suggestion
A named route's final path (its own path combined with any parents) must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```

docs/migrating-vue-router.md
```diff
@@ -15,6 +15,10 @@
 
 Child routes in vue-router have very different behavior depending on if the path starts with `/` or not. In Kitbag Router, the behavior is always the same, so add slashes where you want them and leave them off where you don't.
 
+::: info
+A named route's full path, its own path combined with any parents, must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```
stackoverfloweth: this sounds awkward to me, like it picks up mid-sentence.


```suggestion
A named route's final path (its own path combined with any parents) must start with `/`. Otherwise it can never match a url and the router throws an `UnreachableRouteError`.
```

src/services/createRoute.spec.ts
```diff
@@ -226,6 +226,7 @@ describe('props', () => {
     const spy = vi.fn()
     const parent = createRoute({
       name: 'parent',
+      path: '/parent',
```
stackoverfloweth, line 229: why did this test need to change?

stackoverfloweth: same question throughout this file

pleek91: Because they are named routes. And with the new error any named route must have a valid path. These tests omitted the path for brevity but the new error would make the tests fail from throwing the validation error.

PR #892, opened 2026-09-21
Count the work a navigation waits on

pleek91:
# Description

Every router exposes at most a boolean for a navigation in flight, so apps bolt on a progress bar whose motion is a fake trickle. Kitbag can do better because every unit a navigation waits on is enumerable before it runs: its before hooks, its props getters and loaders, and the async components it renders. A ledger per navigation counts those units as they start and settle.

`useNavigation` reports whether a navigation is pending, which routes it moves between, and the counts, for a determinate bar:

```vue
<script setup lang="ts">
const { pending, settled, total } = useNavigation()
</script>

<template>
  <progress v-if="pending" :value="settled" :max="total" />
</template>
```

A navigation ends when its route has everything it renders with, not when it commits. A rejection completes the count, an abort wipes it, and a navigation begun in place of another, from the user or from a `push` in a loader, starts a fresh one. After hooks and prefetch work are not counted, and neither is a server render or a hydrating navigation. Nothing is awaited in the navigation path; the ledger only observes promises that already exist.

The counts are what a bar needs, so a shipped `RouterProgress` component follows in the next PR.


docs/advanced-concepts/navigation-progress.md
```diff
@@ -0,0 +1,64 @@
...
+
+A navigation is under way from the moment it is asked for until the route it leads to has everything it renders with. Kitbag Router counts the work a navigation waits on, so you can show feedback while slow hooks and data hold a page up.
+
+## What is counted
+
+Every unit of work a navigation waits on is known before it runs, so the progress is real rather than an animated guess.
+
+- **Before hooks**, including redirects and global hooks.
+- **Props getters and loaders** of every route in the match.
+- **Async components** defined with `defineAsyncComponent`.
```
pleek91: Async route components. If an async component is in the component tree but not a route component, we don't know it exists and cannot track it.

docs/advanced-concepts/navigation-progress.md
```diff
@@ -0,0 +1,64 @@
...
+
+Every unit counts the same. After hooks are not counted because they do not hold up the page, and prefetching is not counted because it is not a navigation.
```
pleek91: We probably shouldn't mention perfecting except in how if a route is prefetched that means the navigation will finish sooner.

docs/advanced-concepts/navigation-progress.md
```diff
@@ -0,0 +1,64 @@
...
+
+A navigation ends when its route has everything it renders with, when it is rejected, when it is aborted, or when another navigation begins in its place. A rejection completes the count, since a rejection page is a page too. An abort wipes it. A new navigation, whether from the user or from a `push` in a hook or loader, starts a fresh count of its own.
+
+Nothing is counted while server rendering or while hydrating the server's response, since neither has anything to show.
```
pleek91: This sentence seems unnecessary.

pleek91: Maybe a note that progress isn't tracked during SSR or hydration. Only spa navigations.

docs/advanced-concepts/navigation-progress.md
pleek91: Now "Async route components", with a sentence saying only components the routes render are known to the router.

docs/advanced-concepts/navigation-progress.md
pleek91: Dropped the exclusion. It now says a prefetched route has some units settled already, so its navigation finishes sooner.

docs/advanced-concepts/navigation-progress.md
pleek91: Replaced with an info note: progress is tracked for client side navigations only, not while server rendering or hydrating.

stackoverfloweth (approved):

PR #893, opened 2026-09-21
Ship a RouterProgress bar driven by the navigation counts

pleek91:
# Description

`useNavigation` from #892 gives an app real counts for a navigation in flight, but every app would write the same bar on top of them. `RouterProgress` is that bar, registered alongside `RouterView` and `RouterLink`:

```vue
<template>
  <router-progress />
  <router-view />
</template>
```

It waits 150ms before showing so fast navigations never flash, creeps toward the next unit while waiting so a single slow hook does not look frozen, snaps forward as units settle, and once the route is ready fills and fades, staying at least a moment so it does not blink. An abort removes it without filling. `delay` and `color` are the only props. Color and height can also be set with the `--router-progress-color` and `--router-progress-height` custom properties, and anything else through the `router-progress` classes.

The creeping, the delay, and the fade are all cosmetic and happen inside the component. `useNavigation` still reports the real counts, so a bar sitting at 40% while waiting on a slow hook does not mean the router thinks 40% of the work is done.

[router-progress.webm](https://github.com/user-attachments/assets/c100bb0f-e8ac-4daf-add5-c4a610e39bb4)


src/components/routerProgress.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        afterFinishing(() => {
+          finishing.value = true
+
+          afterFinishing(() => {
+            finishing.value = false
+            hide()
+          }, FADE_TRANSITION)
+        }, WIDTH_TRANSITION)
+      }, remaining)
```
stackoverfloweth: 🤢

src/components/routerProgress.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      const percent = Math.round(shown.value * 100)
+
+      return h('div', {
+        'class': 'router-progress',
+        'role': 'progressbar',
+        'aria-valuemin': 0,
+        'aria-valuemax': 100,
+        'aria-valuenow': percent,
```
stackoverfloweth: not sure what exact aria labels fit best for this sort of component, but it feels like it should at least be labeled as "page loading".


```suggestion
        'aria-valuenow': percent,
        'aria-label': 'Page loading',
```

src/components/routerProgress.ts
```diff
@@ -0,0 +1,209 @@
...
+      } else {
+        end()
+      }
+    }, { immediate: true })
+
+    watch(settled, () => {
+      if (visible.value && pending.value) {
+        shown.value = Math.max(shown.value, progress.value)
+      }
+    })
```
stackoverfloweth, line 37: I wondered what would happen if another navigation event happens before the first one finished. AI came up with this test

```ts
test('a new navigation starting while the bar is showing starts the bar over', async () => {
  const homeFirst = Promise.withResolvers<string>()
  const homeSecond = Promise.withResolvers<string>()
  const otherHold = Promise.withResolvers<string>()

  const home = createRoute({ name: 'home', path: '/', component })

  home.onBeforeRouteEnter(async () => {
    await homeFirst.promise
  })
  home.onBeforeRouteEnter(async () => {
    await homeSecond.promise
  })

  const other = createRoute({ name: 'other', path: '/other', component })

  other.onBeforeRouteEnter(async () => {
    await otherHold.promise
  })

  const router = createRouter([home, other], { initialUrl: '/' })

  const wrapper = mount({ template: '<RouterProgress />' }, {
    global: {
      plugins: [router],
    },
  })

  await advance(150)
  homeFirst.resolve('ok')
  await flushPromises()
  await nextTick()

  expect(width(wrapper)).toBe(50)

  void router.push('/other')
  await flushPromises()
  await nextTick()

  expect(width(wrapper)).toBe(5)

  otherHold.resolve('ok')
  homeSecond.resolve('ok')
})
```

which is fixed by including `to` in `useNavigation` and adding a watch here
```suggestion
    })
    
    watch(to, () => {
      if (pending.value) {
        begin()
      }
    })
```

src/components/routerProgress.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    function hide(): void {
+      clearInterval(trickleTimer)
+      trickleTimer = undefined
+      visible.value = false
+      shown.value = 0
+    }
+
+    function begin(): void {
+      clearFinish()
```
stackoverfloweth: part 2 of my suggestion below


```suggestion
      clearFinish()
      clearTimeout(showTimer)
      showTimer = undefined
```

stackoverfloweth (approved):
couple suggestions, 1 possible bug

pleek91:
> couple suggestions, 1 possible bug

@stackoverfloweth I refactored this so be simpler and more css focused.

